PR #1, opened 2015-07-06
Update README.md

martyngigg:
Use correct grammar


martyngigg:
Failed as expected as there is only the template directory. This was just a test pull request for the Jenkins integration

PR #6, opened 2015-10-29
Tom Perkins exercises - ex 2 (shapes)

tom-perkins:
WordCounter exercise: counts number of occurrences of unique words in an ascii file


exercises/tom_perkins/ex01_basics/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+# Common setup
+include ( CommonSetup )
+
+###############################################################################
+# Executable
+###############################################################################
+
+set ( SRC_FILES
+  src/main.cpp
+  # add any additional source files here
```
OwenArnold, line 21: Build is currently failing because `src/counter.cpp` is missing

exercises/tom_perkins/ex01_basics/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+)
+
+set ( HDR_FILES
+  # add any additional header files here
```
OwenArnold, line 26: Also need `src/counter.h` added

exercises/tom_perkins/ex01_basics/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+)
+
+add_executable ( WordCounter ${SRC_FILES} ${HDR_FILES} )
```
OwenArnold, line 30: I got a warning on my platform because you're using c++11 features without the flags being set correctly. You can fix this by adding the following.
`set_property(TARGET WordCounter PROPERTY CXX_STANDARD 11)` **However** the RHEL6 build will continue to fail because it's using an older version of GCC, which won't support features such as the range-based for loops you used.

exercises/tom_perkins/ex01_basics/src/counter.h
```diff
@@ -0,0 +1,31 @@
...
+#include <vector>
+#include <iostream>
+#include <algorithm>
+
+/**
+ * The Counter object keeps track of words that it has seen
+ * and can report on how many times it has seen each.
+ * Currently works for ASCII only (char not wchar_t).
+ */
+class Counter
```
OwenArnold: Use of separate well-named class for this is good. Not that it's wrong, but why did you use a class over say a free function?

exercises/tom_perkins/ex01_basics/src/counter.cpp
```diff
@@ -0,0 +1,87 @@
+#include "counter.h"
+
+/**
+ * Compares two (word, frequency) pairs to enable sorting.
```
OwenArnold: Good use of comments and free-function. But better to put in an anonymous namespace to keep linkage internal. Good use of const refs.

exercises/tom_perkins/ex01_basics/src/counter.cpp
```diff
@@ -0,0 +1,87 @@
...
+bool compareRecords(const std::pair<std::string, int>& x, const std::pair<std::string, int>& y)
+{
+	return x.second > y.second;
+}
+
+/**
+ * Writes a table of words and frequencies to supplied stream.
+ * Words are sorted in descending order of usage.
+ */
+void Counter::generateReport(std::ostream& stream)
```
OwenArnold: Good function naming. 
- Prefer to return something rather than modify the stream directly (I'll explain in detail in a bit)
- Shouldn't this function be const?
- What about error handling considerations?

OwenArnold:
Looks good so far. Moving forward. We use clang-format for code formatting. There's a VS plugin for that. Also we use doxygen for code comments.

exercises/tom_perkins/ex02_oo_basics/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+set ( HDR_FILES
+  # add any additional header files here
+  src/shape.h
+  src/rectangle.h 
+  src/square.h
+  src/circle.h 
+  src/triangle.h   
+  src/shapesorter.h 
+  src/ShapeTester.h
```
OwenArnold, line 39: File naming case is differnt. On Mantid we use Pascal case. Except for python files.

exercises/tom_perkins/ex02_oo_basics/src/ShapeTester.h
```diff
@@ -0,0 +1,18 @@
...
+//----------------------------------------------------------------------
+#include "shape.h"
+#include <memory>
+
+/**
+ * \brief Abstract base class for testing Shape objects
+ */
+class ShapeTester {
+public:
+  virtual bool match(const Shape &toCompare) const = 0;
```
OwenArnold, line 15: Excellent

exercises/tom_perkins/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,29 @@
...
+ * calculate its perimeter and area.
+ * Each individual shape has a name to refer to it.
+ */
+class Shape {
+public:
+  virtual double perimeter() const = 0;
+  virtual double area() const = 0;
+  const int sides;
+  const std::string type;
+  std::string name;
```
OwenArnold: Why is name non-const but the others are const?

exercises/tom_perkins/ex02_oo_basics/src/shapesorter.cpp
```diff
@@ -0,0 +1,119 @@
+#include "shapesorter.h"
+
+typedef std::vector<std::shared_ptr<Shape>>
```
OwenArnold, line 3: Nicely done. I would probably also typedef std::shape_ptr as something like Shape_sprt

Also, I believe that your Shapes are immutable (if done correctly). So you could store them as std::shared_ptr<const Shape>

exercises/tom_perkins/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/**
+ * Shape: Create a selection of shapes in a vector.
+ * Use these to test the ShapeSorter's methods.
+ */
+int main(int argc, char **argv) {
+  // Create shapes and put them in a vector
+  std::vector<std::shared_ptr<Shape>> shapes;
+  shapes.push_back(std::make_shared<Circle>(1.0, "circle1"));
```
OwenArnold, line 26: Good

exercises/tom_perkins/ex02_oo_basics/src/rectangle.h
```diff
@@ -0,0 +1,24 @@
...
+ */
+class Rectangle : public Shape {
+public:
+  Rectangle(double xIn, double yIn, std::string nameIn);
+  Rectangle(double xIn, double yIn, std::string nameIn, std::string typeIn);
+  virtual double perimeter() const;
+  virtual double area() const;
+
+private:
+  double x;
```
OwenArnold: should be `const double`

exercises/tom_perkins/ex02_oo_basics/src/rectangle.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * \brief Construct a new Rectangle with given sides x, y
+ * \param xIn Length of one side of rectangle
+ * \param yIn Length of other side of rectangle
+ * \param nameIn Name to give the shape
+ */
+Rectangle::Rectangle(double xIn, double yIn, std::string nameIn)
+    : Shape(4, "rectangle", nameIn) {
+  x = xIn;
```
OwenArnold: I think you should use an initialiser list here instead of assignment

exercises/tom_perkins/ex02_oo_basics/src/rectangle.cpp
```diff
@@ -0,0 +1,46 @@
...
+    : Shape(4, typeIn, nameIn) {
+  x = xIn;
+  y = yIn;
+}
+
+/**
+ * \brief Calculates the perimeter of this rectangle.
+ * \returns Perimeter of the rectangle.
+ */
+double Rectangle::perimeter() const { return x + x + y + y; }
```
OwenArnold, line 34: Good

PR #7, opened 2015-10-29
Exercise 1 started. Need to remove words with "--" separators.

LamarMoore:
Exercise 1 is complete and I am submitting for review by a senior developer.

### DIFF
--- a/exercises/lamar_moore/ex01_basics/src/main.cpp
+++ b/exercises/lamar_moore/ex01_basics/src/main.cpp
@@ -2,36 +2,98 @@
 #include <fstream>
 #include <string>
 #include <algorithm>
+#include <vector>
 #include <map>
+
 using namespace std;
 
-bool checkWord(string &wrd);
-bool isPunctuation(char c);
-
-/** Flips an std pair around so that the key becomes the value and the value becomes the key
-
-	@param p This is the pair which will be swapped around.
-*/
-template<typename A, typename B>
-std::pair<B,A> flipPair(const std::pair<A,B> &p)
+namespace
 {
-    return std::pair<B,A>(p.second, p.first);
-}
+	/** Removes the double dash character "--" from strings
 
-/** Creates a flipped multimap which swaps the keys with values for conventional maps
+	@param str String which might contain "--"
+	*/
+	vector<string> splitDoubleDash(string &str)
+	{
+		vector<string> stringList;
+		size_t pos;
 
-	Multimaps are used because the allow duplicate keys.
+		if((pos = str.find("--", 0)) != string::npos)
+		{
+			stringList.push_back(string(str.begin(), str.begin()+pos));
+			stringList.push_back(string(str.begin()+pos+2, str.end()));
+		}
+		else
+			stringList.push_back(str);
 
-	@param src Source map whose keys and values will be swapped
-*/
-template<typename A, typename B>
-std::multimap<B,A> flipMap(const std::map<A,B> &src)
-{
-    std::multimap<B,A> dst;
-    std::transform(src.begin(), src.end(), std::inserter(dst, dst.begin()), 
-                   flipPair<A,B>);
-    return dst;
-}
+		return stringList;
+	}
+
+	/** Predicate function which tests whether or not character is a punctuation character
+
+	@param c character tested against the list ".,?'"!():"
+	*/
+	bool isPunctuation(char c)
+	{
+		bool isPunc = false;
+
+		static const char chars[] = ".,?\'\"!():";
+
+		for(int i=0; i<strlen(chars); i++)
+		{
+			if(c == chars[i])
+			{
+				isPunc = true;
+				break;
+			}
+		}
+
+		return isPunc;
+	}
+
+	/** Checks that the string is at lease 4 chars long, does not start with a digit, removes punctiation and forces lower case.
+
+		@param str String to be examined and if necessary modified.
+	*/
+	bool checkWord(string &str)
+	{
+		bool isWord = false;
+
+		if(str.size() > 4 && !isdigit(str[0]))
+		{
+			isWord = true;
+			str.erase(std::remove_if(str.begin(), str.end(), isPunctuation), str.end());
+			std::transform(str.begin(), str.end(), str.begin(), ::tolower);
+		}
+
+		return isWord;
+	}
+
+	/** Flips an std pair around so that the key becomes the value and the value becomes the key
+
+		@param p This is the pair which will be swapped around.
+	*/
+	template<typename A, typename B>
+	std::pair<B,A> flipPair(const std::pair<A,B> &p)
+	{
+		return std::make_pair(p.second, p.first);
+	}
+
+	/** Creates a flipped multimap which swaps the keys with values for conventional maps
+
+		Multimaps are used because the allow duplicate keys.
+
+		@param src Source map whose keys and values will be swapped
+	*/
+	template<typename A, typename B>
+	std::multimap<B,A> flipMap(const std::map<A,B> &src)
+	{
+		std::multimap<B,A> dst;
+		std::transform(src.begin(), src.end(), std::inserter(dst, dst.begin()), 
+						flipPair<A,B>);
+		return dst;
+	}
+};
 
 int main(int argc, char **argv)
 {
@@ -52,15 +114,22 @@ int main(int argc, char **argv)
 	ifstream infile(fname);
 	ofstream outfile("HolmesWordUsage.txt");
 	map<string, int> wordFreq;
+	vector<string> separated;
 	string word;
 
 	//Read each string from file and process before adding to frequence map
 	while(!infile.eof())
 	{
 		infile >> word;
-		
-		if(checkWord(word))
-			wordFreq[word]++;
+
+		separated = splitDoubleDash(word);
+
+		for(int i=0; i<separated.size(); i++)
+		{
+			if(checkWord(separated[i]))
+				wordFreq[separated[i]]++;
+		}
+
 	}
 
 	multimap<int, string> sortedList = flipMap(wordFreq);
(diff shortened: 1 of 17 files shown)


exercises/lamar_moore/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+using namespace std;
+
+bool checkWord(string &wrd);
+bool isPunctuation(char c);
+
+/** Flips an std pair around so that the key becomes the value and the value becomes the key
+
+	@param p This is the pair which will be swapped around.
+*/
+template<typename A, typename B>
```
OwenArnold: Good.

exercises/lamar_moore/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+std::pair<B,A> flipPair(const std::pair<A,B> &p)
+{
+    return std::pair<B,A>(p.second, p.first);
```
OwenArnold: Could have used std::make_pair here, but otherwise fine.

exercises/lamar_moore/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+	multimap<int, string> sortedList = flipMap(wordFreq);
+
+	outfile<< "Word" << "\tUsage\n" << endl;// write header to output file
+
+	//Write entries in reverse order since keys are sorted in ascending order 1 to n.
+	for(map<int, string>::const_reverse_iterator iter = sortedList.rbegin(); iter != sortedList.rend(); iter++)
+		outfile<<iter->first<<"\t"<<iter->second<<endl;
+
+	system("notepad.exe HolmesWordUsage.txt");
```
OwenArnold: I don't think you would want to leave anything like this in the file.

exercises/lamar_moore/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+/** Predicate function which tests whether or not character is a punctuation character
+
+@param c character tested against the list ".,?'"!():"
+*/
+bool isPunctuation(char c)
+{
+	bool isPunc = false;
+
+	static const char chars[] = ".,?\'\"!():";
```
OwenArnold: Good. Could also have been done with a regex.

exercises/lamar_moore/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+	cout<<"Complete!"<<endl;
+
+	return 0;
+}
+
+/** Checks that the string is at lease 4 chars long, does not start with a digit, removes punctiation and forces lower case.
+
+	@param str String to be examined and if necessary modified.
+*/
+bool checkWord(string &str)
```
OwenArnold: Should link declarations and definitions together in the cpp file. Place in an anonymous namespace.

exercises/lamar_moore/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+	@param c character tested against the list ".,?'"!():"
+	*/
+	bool isPunctuation(char c)
+	{
+		bool isPunc = false;
+
+		static const char chars[] = ".,?\'\"!():";
+
+		for(int i=0; i<strlen(chars); i++)
```
OwenArnold: Failing RHEL6 build. strlen is not defined. As the  I think you need to include cstring to use this, or just convert your char \* to as std::string and use std::string::size()

exercises/lamar_moore/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,34 @@
+#ifndef SHAPE_H
+#define SHAPE_H
+
+namespace Shapes
+{
+	/** List of possible shape types.
+	*/
+	typedef enum
```
OwenArnold: Not a bad solution here. However, 
1. If we are to add more shapes then we'll have to add to this enum. If we're writing a library, this header and enum will be read-only to our clients. So they may not be able to update the definition when they subtype Shape.
2. In a large project (such as Mantid), adding to a header which is used throughout the project could lead to a seriously long recompilation. 
3. This isn't quite as bad as making Shape aware of it's derived types, but it's close. 

For those reasons, I would have chosen to have the type as a std::string.

exercises/lamar_moore/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,34 @@
...
+
+	/** Shape Shape.h "Shape.h"
+		
+		Base interface for all shape types. This includes methods for calculating the perimeter, area
+		and volume of shapes.
+	*/
+	class Shape
+	{
+	public:
+		virtual double calculateArea()=0;
```
OwenArnold: Good naming. However, all of these methods should be marked const.

exercises/lamar_moore/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,34 @@
...
+		virtual double calculatePerimeter()=0;
+		virtual ShapeType getType();
+		virtual int getNumSides();
+	protected:
+		int numSides;
```
OwenArnold: I have a preference for enforcing What they need to do, which you have done with your pure virtual functions, but now How. 

The protected `numSides` and `ShapeType` implies that derived classes should set these values. But this is not enforced. int for example does not need to be explicitly initialised, so you the derived classes do not have to set `type` or `numSlides` and the base class portion of the end type will behave incorrectly.

Solve this in either of the following ways.

Have a constructor on the base class which takes both int for numSides and the ShapeType for type as arguments. That way there is no default constructor on the base and derived classes have to call the base class constructor and consciously pass the required arguments. At the same time make both member fields const, so that they cannot be overwritten once set.

Alternately, (my preference) make `getType` and `getNumSides` pure virtual and remove the protected fields so that derived classes are forced to implement everything. I prefer this because then the Shape interface is purely and interface, and not a mix of a base class and and interface.

exercises/lamar_moore/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,34 @@
...
+		ShapeType type;
+	};
```
OwenArnold, line 22: As a minor point, when it comes to coding in mantid, we use _Hungarian notation_ for member fields. We use a prefix of `m_`. This is the only _Hungarian notation_ we do use.

exercises/lamar_moore/ex02_oo_basics/src/Rectangle.cpp
```diff
@@ -0,0 +1,37 @@
+#include "Rectangle.h"
+
+
+namespace Shapes
+{
+	/** Constructor for Rectangle object
+	@param length Length of the rectangle
+	@param width Width of the rectangle
+	*/
+	Rectangle::Rectangle(double length, double width)
```
OwenArnold: Rectangle::Rectangle(double length) : m_length(length)

exercises/lamar_moore/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include "ShapeSorter.h"
+
+
+using namespace std;
+
+namespace Shapes
+{
+	/** Functor which is used to compare shape types with a matching value in std::remove_if
+	*/
+	class MatchTypeFunctor
```
OwenArnold, line 15: Excellent solution!

Two minor points.
1. put these definitions into an anonymous namespace block.
2. passing a shared pointer by value implies change of ownership and can be costly in a tight-loop due to internal increment/decrement of the copies of the shared_ptr. Would be better to pass by `const shared_ptr<Shape>::element_type&` I'd be happy to explain this. But it's not mandatory for you to change this.

exercises/lamar_moore/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	/** Prints all shapes which have the matching type specified by the first argument
+
+	@param type Shape type for comparison with list of shapes
+	@param shapes List of shapes for which types will be selectively printed
+	*/
+	void ShapeSorter::printByType(ShapeType type, vector<shared_ptr<Shape>> &shapes)
+	{
+		vector<shared_ptr<Shape>> copy = shapes;
```
OwenArnold: I think you could save a lot of allocations deallocations if you used std::copy_if. But I didn't think about that until I looked at the stl.

exercises/lamar_moore/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "ShapeSorter.h"
+
+using namespace std;
+using namespace Shapes;
+
+namespace
+{
+	/** Obtains a random number between 0 and 3 which corresponds to ShapesTypes.
+	*/
+	ShapeType getRandomShape()
```
OwenArnold: Nice way of making any number of shapes.

**Something to consider**
When you're developing in Mantid. We would be using UnitTests CXXTest rather than using hand-rolled main functions for testing. In that case It's nice to see explicit creation rather than the use of random. It also helps a lot with debugging when you want to trace the failure of a particular test case.

exercises/lamar_moore/ex02_oo_basics/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+# Find custom cmake modules
+set ( CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../../../CMake")
+
+# Common setup
+include ( CommonSetup )
+
+###############################################################################
+# Executable
+###############################################################################
+
```
OwenArnold, line 18: Good. Exactly what the CMake list should look like.

OwenArnold:
No longer needed.

PR #10, opened 2016-05-10
ex01 completed

gvardany:
Implemented exercise 1.


exercises/gagik_vardanyan/ex01_basics/src/main.cpp
```diff
@@ -102,7 +104,7 @@ int main (int argc, char ** argv)
     infile.close();
 
     ///sort the map by value
-    pair_vec counts = MapToPairVector(wordCounts);
+    pair_vec counts = mapToPairVector(wordCounts);
```
OwenArnold, line 107: Good

exercises/gagik_vardanyan/ex01_basics/src/main.cpp
```diff
@@ -120,5 +122,7 @@ int main (int argc, char ** argv)
 
     outfile.close();
 
+    std::cout << "output.txt has been succesfully created!\n";
```
OwenArnold, line 125: Good

exercises/gagik_vardanyan/ex02_oo_basics/CMakeLists.txt
```diff
@@ -18,11 +18,27 @@ include ( CommonSetup )
```
OwenArnold, line 18: Excellent. These changes look like they should work. I have to admit that I haven't checked any compiler outputs yet.

exercises/gagik_vardanyan/ex02_oo_basics/src/include/Shape.h
```diff
@@ -1,82 +1,59 @@
...
-    Shape(ShapeBase * sh) { if(sh) cp = sh;}
+    ///default constructor
+    Shape() : type(""), nSides(0) {}
 
-    ///destructor
-    ~Shape() {delete cp;}
+public:
+    ///virtual clone function
+    ///@return a pointer to the new object of the same type
+    virtual Shape* clone() const = 0;
```
OwenArnold, line 21: Yes, this is the right thing to do.

exercises/gagik_vardanyan/ex02_oo_basics/src/include/ShapeSorter.h
```diff
@@ -12,14 +12,14 @@
 
 class ShapeSorter {
 private:
-    std::vector<Shape> & shapes; /**< vector of shapes */
+    std::vector<Shape*> & shapes; /**< vector of shape pointers */
```
OwenArnold: Good. You do indeed need to use a pointer here. If you had c++11 available, I would suggest using a scoped manage smart pointer of some kind rather than storing the raw pointer directly, but this will do fine.

exercises/gagik_vardanyan/ex02_oo_basics/src/include/Square.h
```diff
@@ -4,18 +4,17 @@
...
-class Square: public ShapeBase {
+class Square: public Shape {
 private:
     double side; /**< side of the square */
 
-protected:
+public:
     ///clone function
     /// @return new square object
     Square* clone() const {return new Square(*this);}
```
OwenArnold, line 16: Good. In c++ you can also take advantage of "Covariant return types", so you could have declared the `clone` to be:

``` cpp
 Square* clone() const {return new Square(*this);}
```

That would have been slightly better, but it's a minor tweak to make.
The clone implementation should be in the cpp, but generally you've done a good job moving things into cpp files.

exercises/gagik_vardanyan/ex02_oo_basics/src/main.cpp
```diff
@@ -14,22 +14,17 @@
 ///the main program
 int main(int argc, char ** argv)
 {
-    std::vector<Shape> shapes;
+    std::vector<Shape*> shapes;
 
     Rectangle * rect = new Rectangle(10,50);
```
OwenArnold: Good. However this will leak resources, because you don't ever `delete` the contents of the `std::vector`. You could have solved this in a couple of different ways. 1. You could have introduced a clean-up that enumerated over the vector, and deleted each element. 2. You could have just declared the `Shape` objects on the stack and then just taken their address when filling the vector.

``` cpp
Triangle tri(10, 5)
shapes.push_back(&tri);
```

exercises/gagik_vardanyan/ex02_oo_basics/src/main.cpp
```diff
@@ -14,22 +14,17 @@
...
-    Shape sh4(cq);
-
-    shapes.push_back(sh1);
-    shapes.push_back(sh2);
-    shapes.push_back(sh3);
-    shapes.push_back(sh4);
+    shapes.push_back(rect);
+    shapes.push_back(tri);
+    shapes.push_back(sq);
+    shapes.push_back(cq);
```
OwenArnold, line 28: You have created named temporaries (`sh1`, `sh2`, ...) when you only wanted them to put them directly into the shapes vector. So I would suggest doing something like this to save lines of code:

``` cpp
shapes.push_back(new Square(7));
```

That supposes that you will have to write a delete function, because you have no handle to otherwise delete the individual shapes.

exercises/gagik_vardanyan/ex02_oo_basics/src/main.cpp
```diff
@@ -14,22 +14,17 @@
 ///the main program
 int main(int argc, char ** argv)
 {
-    std::vector<Shape> shapes;
+    std::vector<Shape*> shapes;
 
     Rectangle * rect = new Rectangle(10,50);
     Triangle * tri = new Triangle(10,5);
     Square * sq = new Square(7);
```
OwenArnold: You could have made these shapes constant. Your sorting does not (and should not) modify the Shapes in the vector, so I would suggest enforcing that by storing a `const Shape*` in the `vector`

OwenArnold:
@gvardany. Thanks for making the changes. The changes are generally very good. I've added a few comments, but you shouldn't feel that you have to make these changes.

In terms of style, I am a strong advocate of having pure virtual base classes when you are trying to represent a concept, in this case `Shape` and then having base classes (or other mechanisms) to deal with code sharing. I would therefore have suggested that `Shape` should be pure virtual, and the fields and common methods should have lived in something like a class called `ShapeCommon`. However, what you've done looks good enough to me, so i don't think it's worth changing.

gvardany:
@OwenArnold Thanks for the detailed review. I implemented most of the changes, except making Shape pure virtual, though I see your point, that would be more correct.

OwenArnold:
@gvardany. That all looks in order to me 👍 . I therefore recommend that these exercises be marked as complete for @gvardany. @ianbush. Is my recommendation OK with you?

Gagik, I look forward to working with you on some real issues in the near future.

@ianbush. Is my recommendation OK with you?

ianbush:
Yes, I agree, the new changes look and good!

PR #25, opened 2017-07-05
Michael turner exercises - ex01 & ex02

michaeljturner:
Have written the WordCounter program for ex01_basics.  Runs correctly and outputs data in a file, in the same directory as the input file.


exercises/michael_turner/ex01_basics/src/counter.cpp
```diff
@@ -0,0 +1,104 @@
...
+ *
+ */
+
+void countWords(std::string& line, wordMap& counters){
+	//Lower any capital letters, to ensure output is case insensitive
+	std::transform(line.begin(), line.end(), line.begin(), ::tolower);
+
+	//Iterate through string to find words
+	string_size i = 0;
+	while (  i != line.size()){
```
gemmaguest: While this is logically fine, it's generally better to use less-than rather than not-equal when iterating through indices like this, i.e. `while (i < line.size())`. It aids clarity about what the intent is and avoids any potential mishaps if somehow `i` ends up > `line.size`.

exercises/michael_turner/ex01_basics/src/counter.cpp
```diff
@@ -0,0 +1,104 @@
...
+		//Ignore leading blanks and punctuation (including hyphens)
+		while (i != line.size() && (isPunc(line[i]) || isHyphen(line[i])) )
+			++i;
+
```
gemmaguest: Would it be worth adding an early `break` here if `i >= line.size()`?

exercises/michael_turner/ex01_basics/src/counter.cpp
```diff
@@ -0,0 +1,104 @@
...
+		//Find end of next word, allow hyphenations, but no double hyphens
+		string_size k = i, j = i;
+		//k is always a character ahead of j
+		++k;
```
gemmaguest: Since `k` is not used in many places you could just use `j+1`, and this might actually aid clarity (see below about indexing out of range). I'm ambivalent on this, though.

exercises/michael_turner/ex01_basics/src/counter.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+		//If current character is hyphen, following character cannot be
+		while ( j != line.size() && !isPunc(line[j]) && !(isHyphen(line[j]) && isHyphen(line[k]) )){
```
gemmaguest: Is it possible that `line[k]` could be out of range here? This is where I think using `j+1` instead might add clarity.

Also, this is another place I'd use `j < line.size`

exercises/michael_turner/ex01_basics/src/counter.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+void countWords(std::string& line, wordMap& counters){
+	//Lower any capital letters, to ensure output is case insensitive
+	std::transform(line.begin(), line.end(), line.begin(), ::tolower);
+
+	//Iterate through string to find words
+	string_size i = 0;
+	while (  i != line.size()){
+		//Ignore leading blanks and punctuation (including hyphens)
+		while (i != line.size() && (isPunc(line[i]) || isHyphen(line[i])) )
```
gemmaguest: Again, I'd use `i < line.size()` here

exercises/michael_turner/ex01_basics/src/counter.cpp
```diff
@@ -0,0 +1,104 @@
+#include "counter.h"
+
+#include <algorithm>
+#include <cctype>
+#include <fstream>
+#include <iomanip>
+
+/** Counts the number of occurences of unique words (over 4 letters)
+ *
+ *		@param line	The input text containing words to be counted
```
gemmaguest: As I mentioned, you have tabs instead of spaces. This will be fixed by clang-format in the main repository, but it's good to set up your editor to use spaces instead of tabs or you'll constantly get annoying clang-format errors.

exercises/michael_turner/ex01_basics/src/counter.cpp
```diff
@@ -0,0 +1,104 @@
...
+ * @param outFile		The output ofstream object to write the rank to
+ *
+ */
+
+void rankWords(wordMap& counters, std::ofstream& outFile){
+	wordRank rank = flipMap(counters);
+
+	//Output rank to file
+	outFile << std::left << std::setw(OUT_WIDTH) << "Word" << "Usage\n" << std::endl;
+	for (wordRank::const_reverse_iterator it = rank.rbegin(); it != rank.rend(); ++it)
```
gemmaguest: This is a good place to use `auto` to aid readability: `auto it = rank.rbegin()`

exercises/michael_turner/ex01_basics/src/counter.cpp
```diff
@@ -0,0 +1,104 @@
...
+		outFile << std::left << std::setw(OUT_WIDTH) << it -> second << it -> first << std::endl;
```
gemmaguest: Just a minor formatting comment, but these lines will both end up being split over 2 lines by clang-format. I generally wrap the contents of `for`, `if` etc. in braces unless it's a very short one-liner. This is more of a personal preference (probably a hangover from a previous job where braces were always mandatory!), but I have seen cases where the intent was lost because braces weren't used.

Also, clang-format will remove the spaces in e.g. `it -> second`

exercises/michael_turner/ex01_basics/src/counter.h
```diff
@@ -0,0 +1,33 @@
...
+
+//Includes
+#include <iosfwd>
+#include <map>
+#include <string>
+
+const std::string::size_type  MIN_LENGTH = 5;
+const std::string PUNC = ".,?'\"!():;";
+const std::string HYPHEN = "-";
+const int OUT_WIDTH = 20;
```
gemmaguest: It's best to keep these out of the header file unless they're intended to be used elsewhere - I think they're only used in `counter.cpp` so they could go in there? It's also good to be explicit that they should use internal linkage by specifying them as `static const`, or putting them in an unnamed namespace (I think the latter is preferred). Alternatively, if they're only used in one function they could be `static const` inside that function.

exercises/michael_turner/ex01_basics/src/counter.h
```diff
@@ -0,0 +1,33 @@
...
+
+typedef std::map<std::string, int> wordMap;
+typedef std::multimap<int, std::string> wordRank;
+typedef std::string::size_type string_size;
+
```
gemmaguest: Again, I think `wordRank` and `string_size` could be moved to `counter.cpp`

exercises/michael_turner/ex01_basics/src/counter.h
```diff
@@ -0,0 +1,33 @@
...
+//Turns wordMap into wordRank [multimap]
+wordRank flipMap(wordMap& counters);
+
+//Returns true if character is whitespace or punctuations (except hyphens)
+bool isPunc(char c);
+
+//Returns true if character is hyphen
+bool isHyphen(char c);
+
```
gemmaguest: Again, I think declarations for `flipMap`, `isPunc` and `isHyphen` don't need to be in the header if they're only used in `counter.cpp` - in which case they should also be declared `static` or put in an anonymous namespace so they have internal linkage.

exercises/michael_turner/ex01_basics/src/counter.h
```diff
@@ -0,0 +1,33 @@
...
+const std::string::size_type  MIN_LENGTH = 5;
+const std::string PUNC = ".,?'\"!():;";
+const std::string HYPHEN = "-";
+const int OUT_WIDTH = 20;
+
+typedef std::map<std::string, int> wordMap;
+typedef std::multimap<int, std::string> wordRank;
+typedef std::string::size_type string_size;
+
+//Stores individual words and increments an associated counter
```
gemmaguest: I think the comments for the functions should have triple slash `///` so they get documented in doxygen.

exercises/michael_turner/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,59 @@
...
+			std::string outFilePath = argv[1];
+			outFilePath += "_Word_Rank";
+			std::ofstream outFile(outFilePath);
+
+			if(!outFile){
+				std::cerr << "The file path\"" << outFilePath << "\" does not exist" << std::endl;
+				return EXIT_FAILURE;
+			}
+			else
+				rankWords(counters, outFile);
```
gemmaguest: Very minor point, but it wasn't immediately clear to me that this is where the output file is written, so perhaps a comment or more descriptive name for this function would help.

gemmaguest (dismissed):
Very nice code :+1: :+1:. I've made lots of comments but they're mainly style points & tips. See in-line comments.

exercises/michael_turner/ex01_basics/src/counter.h
gemmaguest: Also, the naming convention for typedefs is PascalCase, e.g. `WordMap` instead of `wordMap`

exercises/michael_turner/ex01_basics/src/counter.h
michaeljturner: I think you can use either /**, /*! or /// (on every line)

exercises/michael_turner/ex02_oo_basics/src/shape_sorter.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <functional>
+#include <string>
+#include <map>
+#include <vector>
+
+/**
+ * @brief ShapeSorter::ShapeSorter      Reads in the vector of shapes to be sorted
+ * @param v         The vector of shapes
+ */
+ShapeSorter::ShapeSorter(std::vector<std::reference_wrapper<Shape> > &v): shapes(v){}
```
gemmaguest: You could add a typedef for `std::vector<std::reference_wrapper<Shape> >`. This is used in quite a lot of places so if it was replaced with, say, `ShapeVector` it would make the code more readable.

exercises/michael_turner/ex02_oo_basics/src/shape_sorter.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * @brief ShapeSorter::printRank        Prints the rank of shapes (sorted by area or perimeter)
+ *                                                          in decreasing order
+ * @param rank      The shapes in ranking order
+ */
+void ShapeSorter::printRank(std::multimap<double, std::reference_wrapper<Shape> > &rank){
+
+    //Iterate through map in reverse order and print value of area/perimeter, followed by the Shape data
+    for(std::multimap<double, std::reference_wrapper<Shape>>::const_reverse_iterator iter = rank.rbegin(); iter != rank.rend(); ++iter){
```
gemmaguest: You could use `auto` here to make it more readable. It's also good to use `crbegin` and `crend` (or `cbegin` and `cend` for forward iterators) if you want a const iterator.

exercises/michael_turner/ex02_oo_basics/src/shape_sorter.cpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+
+/**
+ * @brief ShapeSorter::printSort        Prints the data of the shapes matching a criterium
+ * @param sorted            The selected shapes
+ */
+void ShapeSorter::printSort(std::vector<std::reference_wrapper<Shape>> &sorted){
+    //Iterate through vector and print each shape's data
+    for(auto iter = sorted.begin(); iter < sorted.end(); ++iter){
```
gemmaguest: You can use a range-based for loop for better readability here and in other for loops:
```
for (const auto &shape_ref : sorted) {
  shape_ref.get().printShape();
}
```
Note that this gives you the actual values from the vector rather than iterators, so it's good practice to use `const auto &` to take a reference, otherwise you can end up with unnecessary copies.

exercises/michael_turner/ex02_oo_basics/src/shapes.h
```diff
@@ -0,0 +1,120 @@
+#ifndef SHAPES_H_
+#define SHAPES_H_
+
```
gemmaguest, line 3: In Mantid we'd have a separate header for each class, but don't worry for this small example

exercises/michael_turner/ex02_oo_basics/src/shapes.h
```diff
@@ -0,0 +1,120 @@
...
+
+//Base class
+class Shape
+{
+public:
+    ///Constructor: sets the type and number of sides
+    Shape(const std::string& name, const int &s);
+
+    ///Virtual destructor
```
gemmaguest: The virtual destructor is commented but not declared! Did you mean to add add it here? It's important to have a virtual destructor in the base class, even if you just want the default destructor - a neat way to make the default destructor virtual is `virtual ~Shape() = default;`

exercises/michael_turner/ex02_oo_basics/src/shapes.h
```diff
@@ -0,0 +1,120 @@
...
+    void printShape();
+
+    ///Returns the Shape's Type
+    std::string getType() const;
+
+    ///Returns the number of sides of the shape
+    int getSides() const;
+
+    ///Returns the Shape's Area.
+    virtual double getArea() const {return 0;}
```
gemmaguest: I think it's worth making the virtual functions pure virtual because they are required in all of the derived classes. This will make the base class an abstract class, which I think makes sense because you probably wouldn't want to create a `Shape` without knowing what type of shape it was.

exercises/michael_turner/ex02_oo_basics/src/shape_sorter.cpp
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * @param type      The type of shapes desired
+ */
+void ShapeSorter::types(const std::string& type){
+    std::vector<std::reference_wrapper<Shape>> sorted;
+    int occurences = 0;
+
+    for (auto iter = shapes.begin(); iter < shapes.end(); ++iter){
+
+        if (!(iter->get().getType().compare(type))){
```
gemmaguest: If efficiency was a concern it would be better to make the type an enum to avoid needing to do string comparisons to check the type. Your solution is fine for this example though and has the benefit that it avoids the need to convert the enum to a string for the output.

exercises/michael_turner/ex02_oo_basics/src/shapes.cpp
```diff
@@ -0,0 +1,183 @@
...
+}
+
+/**
+ * @brief Shape::printShape  Displays the Shape's type & measurements on a single line.
+ */
+void Shape::printShape(){
+    std::cout <<std::left<< std::setw(COL_WIDTH) << getType();
+    for (std::map<std::string, double>::iterator iter = dimensions.begin(); iter != dimensions.end(); ++iter){
+        std::cout << std::right <<std::setw(COL_WIDTH)<< iter->first << ": " << std::left << std::setw(COL_WIDTH) << iter->second;
+    }
```
gemmaguest, line 23: Storing the attributes like this means you have a nice simple way to print them all in a consistent format :+1:. Just for info, a common approach is to have a virtual function to print the object, where each derived class takes care of printing itself. This is common in more complex classes where it's beneficial to have direct access to the class members to do the printing.

exercises/michael_turner/ex02_oo_basics/src/shapes.h
```diff
@@ -0,0 +1,120 @@
...
+
+//Rectangle class
+class Rectangle: public Shape
+{
+public:
+    ///Constructor: sets width & height
+    Rectangle(double w, double h);
+
+    ///Returns Rectangle's Area
+    double getArea() const;
```
gemmaguest: It's good practice to add the `override` keyword when you override a function in a derived class, e.g. `double getArea() const override;`. It's functionally the same but makes the intent clear, and the compiler can do some extra checks.

exercises/michael_turner/ex02_oo_basics/src/shapes.h
```diff
@@ -0,0 +1,120 @@
...
+    ///Constructor: sets the type and number of sides
+    Shape(const std::string& name, const int &s);
+
+    ///Virtual destructor
+
+    ///Prints the Shape's Type, followed by its measurements
+    void printShape();
+
+    ///Returns the Shape's Type
+    std::string getType() const;
```
gemmaguest: `getType` and `getSides` could be made pure virtual functions, where the overrides in the derived classes directly return the hard-coded values. This would mean that you don't need to store their values in the base class, and the `CIRCLE_NAME`, `CIRCLE_SIDES` etc. definitions in the anonymous namespace would not be needed. See the note on the `type` member below.

exercises/michael_turner/ex02_oo_basics/src/shapes.cpp
```diff
@@ -0,0 +1,183 @@
...
+    const std::string TRIANGLE_NAME = "triangle";
+    const int TRIANGLE_SIDES = 3;
+
+    const std::string RECTANGLE_NAME = "rectangle";
+    const int RECTANGLE_SIDES = 4;
+
+    const std::string SQUARE_NAME = "square";
+
+    const double PI = 3.14159;
+}
```
gemmaguest: While an anonymous namespace is a good place to put these, since they are specific to particular classes they could be encapsulated further by putting this info into those classes. See the comment below on overriding the `getType` and `getSides` functions. `PI` could be made a static private member of `Circle`, and using constexpr means you can initialise it within the class definition i.e. `static constexpr double PI = 3.14159;`

exercises/michael_turner/ex02_oo_basics/src/shapes.h
```diff
@@ -0,0 +1,120 @@
...
+private:
+    //e.g. type = "square"
+    std::string type;
+
+    //e.g. numSides = 4
+    const int numSides;
+
+    ///Holds the Shape's dimensions
+    /// e.g Triangle: (base, 5), (height, 3)
+    std::map<std::string, double> dimensions;
```
gemmaguest: If you make `dimensions` `protected` instead of `private`, then it can be edited directly from the derived classes, and you can get rid of the `setDim` function. There might be situations where you'd want a base class function to do this instead, e.g. to do some common processing after setting the variable, but in this case I think it's unnecessary.

exercises/michael_turner/ex02_oo_basics/src/shapes.cpp
```diff
@@ -0,0 +1,183 @@
...
+ * @param h     Height of Rectangle
+ */
+Rectangle::Rectangle(double w, double h): Shape(RECTANGLE_NAME, RECTANGLE_SIDES), width(w), height(h){
+    std::map<std::string, double> dim;
+    dim.insert(std::pair<std::string, double>("width", width));
+    dim.insert(std::pair<std::string, double>("height", height));
+
+    //Saves measurements in 'dimensions'
+    setDim(dim);
+}
```
gemmaguest: As noted in the `Shape` class, if you make `dimensions` `protected` then it can be used directly here instead of creating the temporary `dim` object, and `setDim` could be removed. Likewise for the other derived classes.

exercises/michael_turner/ex02_oo_basics/src/shapes.cpp
```diff
@@ -0,0 +1,183 @@
...
+    std::cout << std::endl;
+}
+
+/**
+ * @brief Shape::getType        Gets the shape's type
+ * @return      std::string
+ */
+std::string Shape::getType() const{
+    return type;
+}
```
gemmaguest: As noted in the `Shape` class, `getType` could be made virtual and return the hard-coded value directly from the derived function overrides, rather than it having to be stored in the `type` member variable.

exercises/michael_turner/ex02_oo_basics/src/shapes.h
```diff
@@ -0,0 +1,120 @@
...
+protected:
+    ///Saves the Shape's dimensions
+    void setDim(std::map<std::string, double>& dim);
+
+    ///Sets the shape's type
+    void setType(const std::string &name);
+
+private:
+    //e.g. type = "square"
+    std::string type;
```
gemmaguest: As noted above, `type` and `numSides` wouldn't need to be stored if they were returned by overrides of `getType` and `getSides`. This would save a bit of memory (only a concern if there are likely to be many instances of these classes), and avoids any possibility that these values could be changed accidentally at runtime. However, polymorphism can add a performance overhead so if that were an issue then your implementation would be better. None of these issues is important for this example so I'm ambivalent if you want to make this change or not :)

gemmaguest (changes requested):
Exercise 2 looks good :+1: :+1:. I've added a few suggestions for improvements, mainly relating to style and good practice. I've also noted some alternative ways things could be implemented, although these are more for info as they're not particularly relevant to this example.

michaeljturner:
Have updated ShapeSorter exercise

exercises/michael_turner/ex02_oo_basics/src/shape_sorter.cpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+
+/**
+ * @brief ShapeSorter::printSort        Prints the data of the shapes matching a criterium
+ * @param sorted            The selected shapes
+ */
+void ShapeSorter::printSort(ShapeVector &sorted){
+    //Iterate through vector and print each shape's data
+    for(const auto &shape_ref: sorted){
```
gemmaguest, line 42: Doh, I think I suggested the name `shape_ref` (old habit) but actually the convention in Mantid is for camelCase variable names. My bad. No need to change it.

exercises/michael_turner/ex02_oo_basics/src/shape_sorter.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+
+/**
+ * @brief ShapeSorter::area     Ranks shapes in order of decreasing area
+ */
+void ShapeSorter::area(){
+    std::multimap<double, std::reference_wrapper<Shape>> rank;
+
+    for (const auto &shape_ref: shapes){
+        rank.insert(std::pair<double, std::reference_wrapper<Shape>>(shape_ref.get().getArea(), shape_ref.get()));
```
gemmaguest, line 101: No need to change it, but you could use `std::make_pair` here to improve readability (it avoids having to specify the element types).

gemmaguest (approved):
The changes look good :+1: Happy for this to go to the gatekeeper @AntonPiccardoSelg

gemmaguest:
I think we forgot to close this PR.

PR #28, opened 2017-07-07
Edward Brown Excercise 1 and 2

edwardb96:


martyngigg:
There was a configuration problem on the build server that I have now fixed. Please ignore the RHEL6 line as it is no longer valid. The other 2 failures are real.

exercises/edward_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "WordFrequencyTable.h"
+#include <iostream>
+
+/** Ensures the existance of at least one command line argument to be used as the input
+ *  file name.
+ *  @param ArgumentCount The value of the argc parameter passed to main.
+ *  @return True if the minimum number of arguments or greater are present.
+ */
+bool hasRequiredCommandLineArguments(int ArgumentCount) {
+  auto constexpr MINIMUM_COMMAND_LINE_ARGUMENTS = 1;
```
gemmaguest: The minimum here needs to be 2 (because the executable name is always the first argument)

exercises/edward_brown/ex01_basics/src/WordCharPredicates.cpp
```diff
@@ -0,0 +1,5 @@
+#include <cctype>
+
+bool isWordChar(char Candidate) { return !(std::isspace(Candidate) != 1); }
```
gemmaguest: I think there's a stray extra `!` here causing this check to fail (or maybe it was supposed to be `!= 0`?). I would actually use `isspace(Candidate) == 0` for the check because the return from `isspace` is documented only as being zero or non-zero, and to avoid a double negative.

exercises/edward_brown/ex01_basics/include/TakeWhile.h
```diff
@@ -0,0 +1,31 @@
...
+ *  If no such character exists then the substring s will we equal to the string
+ *  of all characters in the range [Begin, End)
+ *
+ *  @param Begin The const_iterator marking the begining of the range to be
+ *  searched for c.
+ *  @param End The const_iterator marking the end of the range to be
+ *  searched for c.
+ *  @param ShouldTake A callable object which returns a pair containing the
+ *  substring s and the const_iterator which points to c.
+ */
```
gemmaguest: Missing documentation for the return value

exercises/edward_brown/ex01_basics/include/Punctuation.h
```diff
@@ -0,0 +1,20 @@
...
+#define EX1_PUNCTUATION
+#include <string>
+
+/** Determines if Candidate holds any of the following character values
+ *  `.,?'"!():`.
+ *
+ *  @param The character to test.
+ *  @return True if the Candidate is a punctuation mark, false if it is not.
+ */
+bool isPunctuation(char Candidate);
```
gemmaguest: The descriptive comments should go in the cpp file to keep the header compact. You can optionally include a brief description of functions in headers (see the [coding standards](http://www.mantidproject.org/C%2B%2B_Coding_Standards#Function_headers))

exercises/edward_brown/ex01_basics/src/Punctuation.cpp
```diff
@@ -0,0 +1,16 @@
+#include "Punctuation.h"
+#include <algorithm>
+
+bool isPunctuation(char Candidate) {
+  auto const punctuation = std::string(".,?'\\\"!():");
```
gemmaguest: Would this be better as `static const std::string punctuation(".,?'\\\"!():");`?

exercises/edward_brown/ex01_basics/src/Punctuation.cpp
```diff
@@ -0,0 +1,16 @@
...
+  return std::find(punctuation.cbegin(), punctuation.cend(), Candidate) !=
+         punctuation.cend();
+}
+
+std::string stripPunctuation(std::string MaybeContainsPunctuation) {
```
gemmaguest: Could `MaybeContainsPunctuation` be passed by reference here instead of returning a copy?

exercises/edward_brown/ex01_basics/include/ForEachWord.h
```diff
@@ -0,0 +1,29 @@
...
+ *  @param DoIt A callable object to be called for each word found in the range
+ * [Begin, End).
+ */
+template <typename ConstCharInputIterator, typename Action>
+void forEachWord(ConstCharInputIterator Begin, ConstCharInputIterator End,
+                 Action DoIt) {
+  auto Current = Begin;
+  while (Current != End) {
+    auto WordPositionPair = takeWhile(Current, End, &isWordChar);
+    DoIt(std::move(WordPositionPair.first));
```
gemmaguest: Is there an advantage to using `std::move` here over passing by reference to `DoIt`?

exercises/edward_brown/ex01_basics/include/TakeWhile.h
```diff
@@ -0,0 +1,31 @@
...
+template <typename ConstCharInputIterator, typename CharPredicate>
+std::pair<std::string, ConstCharInputIterator>
+takeWhile(ConstCharInputIterator Begin, ConstCharInputIterator End,
+          CharPredicate ShouldTake) {
+  std::stringstream WordStream;
+  auto Current = Begin;
+  while (Current != End && ShouldTake(*Current)) {
+    WordStream << *Current;
```
gemmaguest: I think stringstream operations are quite slow, so it might be more efficient to find the start/end positions and then take the substring. Probably not worth worrying about for this example, though.

exercises/edward_brown/ex01_basics/include/MakeWordFrequencyMap.h
```diff
@@ -0,0 +1,45 @@
...
+makeWordFrequencyMap(ConstCharInputIterator begin, ConstCharInputIterator end) {
+  auto wordFrequencyMap = std::unordered_map<std::string, int>();
+  forEachWord(
+      begin, end, [&wordFrequencyMap](std::string WordCandidate) -> void {
+        WordCandidate = stripPunctuation(WordCandidate);
+        if (isWord(WordCandidate)) {
+          wordFrequencyMap[toLowerCase(stripPunctuation(WordCandidate))]++;
+        }
+      });
+  return std::move(wordFrequencyMap);
```
gemmaguest: I never know whether `std::move` is a good idea here. The last time I googled around I think I came to the conclusion that the compiler would do a better job at optimising out the copy (and in fact maybe `std::move` can mess up return value optimisation). However, I think this is used occasionally in Mantid, although not much. What do you think?

exercises/edward_brown/ex01_basics/src/ToLowerCase.cpp
```diff
@@ -0,0 +1,10 @@
+#include "ToLowerCase.h"
+#include <algorithm>
+#include <cctype>
+
+std::string toLowerCase(std::string MixedOrUpperCase) {
```
gemmaguest: Could you pass `MixedOrUpperCase` by reference and avoid returning a copy?

exercises/edward_brown/ex01_basics/src/ToLowerCase.cpp
```diff
@@ -0,0 +1,10 @@
...
+  std::transform(MixedOrUpperCase.cbegin(), MixedOrUpperCase.cend(),
+                 MixedOrUpperCase.begin(),
+                 [](char C) -> char { return std::tolower(C); });
```
gemmaguest: You can just pass `::tolower` directly here rather than via a lambda function

gemmaguest (commented):
Nice solution :+1: :+1: I've added lots of in-line comments, mostly minor style points or questions (for my own understanding :) ). 

On file layout, while it's good to have separate files for different classes, I'd group small utility functions together more i.e. into a single cpp file for related functions. Also minimise what goes into the header i.e. just the declarations with maybe a brief description. It's probably not worth changing for this example, though.

exercises/edward_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/// Indicates to the user that no input files were provided.
+void noInputFiles() { std::cerr << "No input file specified." << std::endl; }
+
+int main(int argc, char **argv) {
+  if (hasRequiredCommandLineArguments(argc)) {
+    try {
+      auto const InputFileName = std::string(argv[1]);
+      auto const InputString = readInputFile(InputFileName);
```
gemmaguest: If dealing with large files we'd probably want to process the file line-by-line rather than reading it all in.

exercises/edward_brown/ex01_basics/src/main.cpp
edwardb96: Fixed in ec1d827

exercises/edward_brown/ex01_basics/src/WordCharPredicates.cpp
edwardb96: Fixed in 7563492

exercises/edward_brown/ex01_basics/include/TakeWhile.h
edwardb96: Added in 702c63e

exercises/edward_brown/ex01_basics/include/TakeWhile.h
edwardb96: I did this in f7bfbd17e1eb296fa822438bbc9bc7c46daa1fe8 but had to change it back again in f4a8b5b8e33a8c2193f78109ff5b7785e29d4b5d due to the surprising way that [std::istreambuf_iterator](http://en.cppreference.com/w/cpp/iterator/istreambuf_iterator) performs comparisons and the fact that it isn't a random access iterator.

exercises/edward_brown/ex01_basics/src/Punctuation.cpp
edwardb96: Resolved in 8fe39c873488d896b759fa70f04fb7c6e4d3e922

exercises/edward_brown/ex01_basics/src/Punctuation.cpp
edwardb96: Yes, Resolved in 50ea219

exercises/edward_brown/ex01_basics/src/ToLowerCase.cpp
edwardb96: Changed it in 622cf98.

exercises/edward_brown/ex01_basics/src/main.cpp
edwardb96: I changed the approach in f4a8b5b8e33a8c2193f78109ff5b7785e29d4b5d which doesn't read the entire file at once but also not line by line. The alternative would be to make more significant changes to the interfaces of `forEachWord` and `makeWordFrequencyMap` or even write a custom file iterator.

exercises/edward_brown/ex01_basics/include/MakeWordFrequencyMap.h
edwardb96: After some research it appears it's not a good idea. I have recieved compiler warnings about unecessary `move`s on return values in the past when compiling c++14 code with `-Wall` set but didn't remember getting anything with a c++11 compiler so assumed that the copy elision rules were somehow enhanced in c++14. 

However the following paragraph from [here](http://en.cppreference.com/w/cpp/language/copy_elision) reveals otherwise. 

> If a function returns a class type by value, and the return statement's expression is the name of a non-volatile object with automatic storage duration, which isn't the function parameter, or a catch clause parameter, and which has the same type (ignoring top-level cv-qualification) as the return type of the function, then copy/move (since C++11) is omitted. When that local object is constructed, it is constructed directly in the storage where the function's return value would otherwise be moved or copied to. This variant of copy elision is known as NRVO, "named return value optimization"

I have removed the extra move in 37d7595635f1bb589548e3e2f91bfef100e2c391

exercises/edward_brown/ex01_basics/include/ForEachWord.h
edwardb96: If `DoIt` wished to modify the lifetime of the string, keep a copy or wished to modify it without affecting the caller then there may be a benefit in the form of a move instead of a copy. But I can't remember the exact details regarding the implementation of `std::unordered_map` so I changed to by reference passing instead. See 622cf98fa6936e055d3aa341a2a2522dc32aa6ff

exercises/edward_brown/ex01_basics/include/Punctuation.h
edwardb96: Have moved most of the documentation comments in 50ea21916dab29637309290bc5da97593446bec8

edwardb96: Just looking at the code for exercise two, presumably this does not apply to pure virtual functions who's declarations are only in the header and which (unlike inline or unspecialized template functions which need to be accessible from the header) don't have a separate definition.

gemmaguest: Good point. You're right - I just did a quick search and found some quite verbose descriptions in headers in this case, so it looks like that's fine :+1:

exercises/edward_brown/ex01_basics/include/TakeWhile.h
gemmaguest: Sure, that makes sense with these inputs :+1:

exercises/edward_brown/ex01_basics/src/Punctuation.cpp
gemmaguest: :+1:

Also, is there a reason for using the assignment operator rather than initialising explicitly? I've always used the latter to avoid any suggestion there could be a temporary copy (although I'm sure this would be optimised out anyway)

exercises/edward_brown/ex01_basics/src/ToLowerCase.cpp
edwardb96: I did this initially but then discovered the notes for [this](http://en.cppreference.com/w/cpp/string/byte/tolower). 

> Like all other functions from \<cctype\>, the behavior of std::tolower is undefined if the argument's value is neither representable as unsigned char nor equal to EOF. To use these functions safely with plain chars (or signed chars), the argument should first be converted to unsigned char

exercises/edward_brown/ex01_basics/src/Punctuation.cpp
edwardb96: For *local* variables I tend to prefer the use of `auto` over direct initialization for the following reasons:
1. Less issues with the [most vexing parse](https://en.wikipedia.org/wiki/Most_vexing_parse). Which is especially useful when the LLVM style guide explicitly outlaws uniform initialization for most classes. 
2. It's harder to end up with any uninitialized variables since `auto` variables must have an initializing expression.
3. Where one variable is initialized by a more complex expression or just another variable it will be updated automatically as the type/type name changes.  (If I want to commit to a specific type I can always use a `static_cast<>` or similar).

As an additional nicety when declaring multiple locals on consecutive lines, the variable names line up consistently which I feel improves readability but I accept may just be personal taste...

```
auto firstName = std::string("Joe");
auto surname = std::string("Bloggs");
auto age = 27;

// rather than

std::string firstName("Joe");
std::string surname("Bloggs");
int age = 21;
```
 However, obviously when dealing with non STL library types which may be expensive to move/copy I understand that some extra care must be taken.

gemmaguest: Good points :+1: I'm not convinced by point 3 but maybe I'm just old-fashioned ;) It looks like this style is used a bit in Mantid although not that much, so the rule of thumb is to fit in with the local style where appropriate.

exercises/edward_brown/ex01_basics/include/MakeWordFrequencyMap.h
gemmaguest: Thanks for checking that out - that's good to know :+1:

exercises/edward_brown/ex01_basics/src/ToLowerCase.cpp
gemmaguest: Good spot! But then shouldn't the argument to the lambda function be `unsigned char`?

exercises/edward_brown/ex01_basics/include/TakeWhile.h
gemmaguest: Could you use `std::string` with `append()` or `operator+=` instead of a stringstream?

exercises/edward_brown/ex01_basics/src/ToLowerCase.cpp
edwardb96: Looks like that change got caught up in f4a8b5b8e33a8c2193f78109ff5b7785e29d4b5d.

exercises/edward_brown/ex01_basics/include/TakeWhile.h
edwardb96: Changed in 8db195e065037de58530a3c70ac7d9052699d642

exercises/edward_brown/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+using ShapeVector = typename ShapeSorter::ShapeVector;
+
+ShapeVector makeShapeList() {
+  ShapeVector shapes;
+  shapes.emplace_back(std::make_unique<Circle>(10.0));
+  shapes.emplace_back(std::make_unique<Triangle>(10.0, 5.0));
+  shapes.emplace_back(std::make_unique<Rectangle>(4.0, 8.0));
+  shapes.emplace_back(std::make_unique<Square>(3.0));
+  return std::move(shapes);
```
gemmaguest: Is this an unnecessary `std::move` due to NRVO?

exercises/edward_brown/ex01_basics/include/WordFrequencyTable.h
```diff
@@ -0,0 +1,46 @@
...
+template <typename ConstWordFrequencyPairInputIterator>
+WordFrequencyTable
+makeOrderedWordFrequencyTable(ConstWordFrequencyPairInputIterator Begin,
+                              ConstWordFrequencyPairInputIterator End) {
+  auto table = WordFrequencyTable(Begin, End);
+  std::sort(table.begin(), table.end(),
+            [](WordFrequencyPair const &A, WordFrequencyPair const &B) -> bool {
+              return A.second > B.second;
+            });
+  return std::move(table);
```
gemmaguest: Unnecessary `std::move` due to NRVO?

gemmaguest (commented):
The second exercise looks great :+1: I just spotted a couple more places where I think `std::move` is unnecessary.

gemmaguest (approved):
Thanks for the changes. This looks good to me now.

Just a couple of general tips when you come to work on the main Mantid repository:
- Split long commit messages into a short summary and longer body (see [writing good commit messages](https://github.com/erlang/otp/wiki/Writing-good-commit-messages). 
- Include a description and testing instructions in the PR (there's a template prompting you to do this when you create the PR)

gemmaguest:
Hi @AntonPiccardoSelg , this is ready for gate-keeper check

exercises/edward_brown/ex01_basics/include/MakeWordFrequencyMap.h
```diff
@@ -0,0 +1,46 @@
...
+#include "ToLowerCase.h"
+#include <unordered_map>
+
+/** A predicate which determines if CandidateWord is a word for the purposes of
+ *  word frequency analysis.
+ *
+ *  @param CandidateWord The candidate to check.
+ *  @return True if CandidateWord is a 'word' false if not.
+ */
+bool isWord(std::string const &CandidateWord) {
```
AntonPiccardoSelg, line 14: In general Mantid uses `camelCase` for variable names, but I don't think it is explicitly mentioned in the style guide lines. Just something to be aware about for the future.

exercises/edward_brown/ex01_basics/include/MakeWordFrequencyMap.h
```diff
@@ -0,0 +1,46 @@
...
+ *  @param Begin The const_iterator marking the beginning of the range to search
+ * for words.
+ *  @param End The const_iterator marking the end of the range to search for
+ * words.
+ *  @return The map built from the word frequency analysis performed.
+ */
+template <typename ConstCharInputIterator>
+std::unordered_map<std::string, int>
+makeWordFrequencyMap(ConstCharInputIterator begin, ConstCharInputIterator end) {
+  auto wordFrequencyMap = std::unordered_map<std::string, int>();
```
AntonPiccardoSelg, line 35: Is there any advantage over `std::unordered_map<std::string, int> wordFrequencyMap`? The compiler will handle it in the same way, so it does not matter too much.

exercises/edward_brown/ex01_basics/include/FileHelpers.h
```diff
@@ -0,0 +1,9 @@
+#ifndef EX1_FILE_HELPERS
+#define EX1_FILE_HELPERS
+#include <fstream>
+#include <string>
+
+void openInputFile(std::string const &Filename, std::ifstream &InputFileStream);
+void openOutputFile(std::ofstream &OutputFileStream);
+std::string fileToString(std::ifstream const &File);
```
AntonPiccardoSelg, line 8: Is this function used anywhere? If not it should be removed.

exercises/edward_brown/ex01_basics/src/WordCharPredicates.cpp
```diff
@@ -0,0 +1,5 @@
+#include <cctype>
+
+bool isWordChar(char Candidate) { return std::isspace(Candidate) == 0; }
+
+bool isWhitespaceChar(char Candidate) { return std::isspace(Candidate) != 0; }
```
AntonPiccardoSelg, line 5: Do you test for the hyphen? See [here](http://www.mantidproject.org/New_Starter_C++_introduction#Exercise)

AntonPiccardoSelg (commented):
Exercise 1:
* Some minor points and info. Nothing that requires a change for this exercise.
* You don't seem to be testing hyphenated words, but maybe I missed something

Exercise 2:
Looks very nice. Good use memory management.

All looks good to me. Happy to merge

exercises/edward_brown/ex01_basics/include/MakeWordFrequencyMap.h
edwardb96: Please see the earlier discussion on this PR (Punctuation.cpp) with @gemmaguest regarding the use of auto for *local* variables. It mostly comes down to ~~three~~ four things.

1. Less issues with the [most vexing parse](https://en.wikipedia.org/wiki/Most_vexing_parse). Which is especially useful when the LLVM style guide explicitly outlaws uniform initialization in the majority of cases. 
2. It's harder to end up with any uninitialized variables since `auto` variables must have an initializing expression.
3. There is less work and changes required when refactoring functions used to initialize variables or renaming the deduced (tracked) types themselves since they are deduced rather than written explicitly. 

4. When declaring multiple locals on consecutive lines, the variable names line up consistently which I feel improves readability but I accept may just be personal taste...

```
auto firstName = std::string("Joe");
auto surname = std::string("Bloggs");
auto age = 27;

// rather than

std::string firstName("Joe");
std::string surname("Bloggs");
int age = 21;
```
 However, obviously when dealing with non STL library types which may be expensive to move/copy I understand that some extra care must be taken and as was suggested above I am of course more than happy to "fit in with the local style where appropriate".

exercises/edward_brown/ex01_basics/include/FileHelpers.h
edwardb96: Good spot.

exercises/edward_brown/ex01_basics/src/WordCharPredicates.cpp
edwardb96: I was a little confused by the wording of the sentence regarding hyphenation, specifically:
> treat hypenated words as a different word.

Taking the following as an example input `hello-world hello world`, which of the two (or neither) is/was the desired behaviour.

Hyphenated words are treated as different word**s**  producing
```
hello   2
world   2
```

or

Hyphenated words are treated as a different word (which is different from those which form it).
```
hello-word   1
hello        1
world        1
```
At the moment I believe the program exhibits the second of these behaviours since `-` is not a white space character and neither is it a 'punctuation mark' which means it won't be stripped out when inserting into the map. 

I would guess switching to the second behaviour could be achieved by adding a new function 
```
bool isWhitespaceOrHyphenChar(char Candidate) {
  return isWhitespaceChar(candidate) || candidate == '-'; 
}
``` 
and is used where `isWhitespaceChar` is currently used.

exercises/edward_brown/ex01_basics/include/MakeWordFrequencyMap.h
edwardb96: Pleased to hear it, it may be something worth adding to the wiki pages. I thought the [LLVM](http://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly) convention was a little odd but fell back on it as that's what the standards say to do.

> Variable names should be nouns (as they represent state). The name should be camel case, and start with an upper case letter (e.g. Leader or Boats).

exercises/edward_brown/ex01_basics/src/WordCharPredicates.cpp
AntonPiccardoSelg: Hmm, that is a good point. I think your implementation is correct.

exercises/edward_brown/ex01_basics/include/MakeWordFrequencyMap.h
AntonPiccardoSelg: I think you have some valid points here and I think it is ok to use that style.
As far as I am aware there is no "local style" for this.  

1. Yep, it would clarify things,  as would the use of `{...}` (which people might forget to use)
2. Agreed, you avoid things like `int k;`. For objects that should not matter though.
3. This is rather a point for using `auto` (which you should always do). In your case the refactoring effort seems about the same.
4. hmm, I agree with you on that. It does seem easier to read.

PR #29, opened 2017-07-07
Matthew bowles exercises

MatthewBowles:
## Exercise 1

A simple word counter that counts the number of times each unique word appears in a given text file.
Runs from command line with the input file name as the first argument. 
Output is saved to Output.txt.
Can be tested with [Holmes.txt](https://github.com/martyngigg/cpp-examples/raw/master/Holmes.txt) as input.

## Exercise 2

Hierarchy of shape classes, each with a type and number of sides, and each with area and perimeter methods. The main method creates eight shapes (two of each type) and demonstrates the four functions of ShapeSorter:
1. Print shapes that match a chosen type (Circle)
2. Print shapes with a chosen number of sides (4)
3. print shapes in order of area (descending)
4. print shapes in order of perimeter (descending)


exercises/matthew_bowles/ex01_basics/inc/main.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
```
AntonPiccardoSelg: I think this will work on our build servers but it is not standard. We use include guards.

exercises/matthew_bowles/ex01_basics/inc/main.h
```diff
@@ -0,0 +1,34 @@
...
+string removePunctuation(string& word);
+
+//converts a string to lowercase
+string toLower(string& word);
+
+//sorts map of words into descending order by value
+vector<pair<string, int>> sortWords(map<string, int>& words);
+
+//outputs the words to a file
+void outputCount(vector<pair<string, int>>& words);
```
AntonPiccardoSelg: new line (this is not strictly necessary, but even Github complains about it). See [here](https://stackoverflow.com/questions/72271/no-newline-at-end-of-file-compiler-warning)

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+/** Converts a string to lowercase
+*
+* @param word The string to convert to lowercase
+* @return The string in lowercase
+*
+*/
+string toLower(string& word)
+{
+	int i = 0;
+	while (i < word.length() && isupper(word[i]))
```
AntonPiccardoSelg: that assumption does not seem to be right. You can have camelCase word.

AntonPiccardoSelg: have a look at `std::transform` in combination with  `tolower`

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+* @param path The path to the input file
+* @return A map of words and their number of uses
+*
+*/
+map<string, int> loadFile(const string& path)
+{
+	map<string, int> words;
+	ifstream fileIn;
+	string line;
+	string word;
```
AntonPiccardoSelg: This does not seem to be used. Please remove

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+* @param words The map of words to hold the resulting words
+*
+*/
+void splitLine(string& line, map<string, int>& words)
+{
+	string word;
+	size_t space;
+
+	space = line.find(" ");
+	while (space != string::npos)
```
AntonPiccardoSelg: You don't seem to be considering hyphenated words.

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+	{
+		word = line.substr(0, space);
+		++space;
+		line = line.substr(space);
+		addWord(words, word);
+		space = line.find(" ");
+	}
+	addWord(words, line);
```
AntonPiccardoSelg: Does this add potentially an empty string?

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+	}
+}
+
+/** Removes punctuation from the start or end of a string
+*
+* @param word The string to have punctuation removed
+* @return The string without leading or trailing punctuation
+*
+*/
+string removePunctuation(string& word)
```
AntonPiccardoSelg: You seem to manipulate on the input, but you also provide as a return value. Your function should have only one of them.

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+{
+	//check end of word
+	while (word.length() > 1 && !isalpha(word[word.length() - 1]))
+		word = word.substr(0, word.length() - 1);
+
+	//check start of word
+	while (word.length() > 1 && !isalpha(word[0]))
+		word = word.substr(1);
+
+	return word;
```
AntonPiccardoSelg: Your code would remove characters which are valid, for example `+`. Only `.,?'"!():` are punctuation characters. Also the punctuation characters could be in the middle of the word

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+/** Adds a word to a given map if applicable
+*
+* @param words The map of words to add to
+* @param word The word to be added
+*
+*/
+void addWord(map<string, int>& words, string& word)
+{
+	map<string, int>::const_iterator existing;
```
AntonPiccardoSelg: Don't define variables in a scope that they are not required in yet (unless there is a good reason)

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+	word = removePunctuation(word);
+	if (word.length() > 4)
+	{
+		word = toLower(word);
+
+		existing = words.find(word);
+		
+		if (existing == words.end())
```
AntonPiccardoSelg: use `if (words.count(word) == 0) {...`

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+	if (word.length() > 4)
+	{
+		word = toLower(word);
+
+		existing = words.find(word);
+		
+		if (existing == words.end()) 
+		{
+			//if word isn't already in map, add it with quantity 1
+			words.insert(std::pair<string, int>(word, 1));
```
AntonPiccardoSelg: prefer `std::make_pair(word, 1)`, you will not have to specify the template arguments

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+* @param words The map of words to be sorted
+* @return A sorted vector of word-value pairs
+*
+*/
+vector<pair<string, int>> sortWords(map<string, int>& words)
+{
+	vector<std::pair<string, int>> wordsVector;
+	map<string, int>::iterator iter;
+
+	for (iter = words.begin(); iter != words.end(); ++iter)
```
AntonPiccardoSelg: prefer a range-based for loop, ie `for (const auto& element : words) { ...`

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+	return word;
+}
+
+/** Converts a map of words into a vector of pairs and sorts (descending by value)
+*
+* @param words The map of words to be sorted
+* @return A sorted vector of word-value pairs
+*
+*/
+vector<pair<string, int>> sortWords(map<string, int>& words)
```
AntonPiccardoSelg: You can use `const  &` here. you are not changing the words map.

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+{
+	return x.second > y.second;
+}
+
+/** Outputs vector of word-value pairs to a file
+*
+* @param words The vector of word-value pairs to output
+*
+*/
+void outputCount(vector<pair<string, int>>& words)
```
AntonPiccardoSelg: use `const &`

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+	ofstream output;
+	auto iter = words.cbegin();
+	string word;
+
+	output.open("Output.txt");
+	output << "Word" << setw(16) << "Usage" << "\n" << "\n";
+
+	while(iter != words.cend())
+	{
+		word = (iter->first).c_str();
```
AntonPiccardoSelg: you can use `size` on a `std::string`. There should hardly be a reason to use `c_str`.

exercises/matthew_bowles/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,188 @@
...
+void outputCount(vector<pair<string, int>>& words)
+{
+	ofstream output;
+	auto iter = words.cbegin();
+	string word;
+
+	output.open("Output.txt");
+	output << "Word" << setw(16) << "Usage" << "\n" << "\n";
+
+	while(iter != words.cend())
```
AntonPiccardoSelg: prefer a range-based for loop in this case

exercises/matthew_bowles/ex02_oo_basics/inc/Circle.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+#include "Shape.h"
+#include <math.h>
+
+class Circle :
+	public Shape
+{
+public:
+	Circle(double radius);
+	~Circle();
```
AntonPiccardoSelg: destructor is not needed here (Look out for this in other files)

exercises/matthew_bowles/ex02_oo_basics/inc/Circle.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
```
AntonPiccardoSelg: See above about pragma once

exercises/matthew_bowles/ex02_oo_basics/inc/Circle.h
```diff
@@ -0,0 +1,19 @@
...
+#include "Shape.h"
+#include <math.h>
```
AntonPiccardoSelg: Why do you need `math.h` here. Only include what is needed. (Look out for this in other files)

exercises/matthew_bowles/ex02_oo_basics/inc/Rectangle.h
```diff
@@ -0,0 +1,19 @@
...
+#include "Shape.h"
+
+class Rectangle :
+	public Shape
+{
+public:
+	Rectangle(double side1, double side2);
+	~Rectangle();
+
+	virtual double calcArea() const override;
```
AntonPiccardoSelg: `virtual` keyword is not required here. However using `override` is a good idea :+1:  (Look out for this in other files)

exercises/matthew_bowles/ex02_oo_basics/src/Shape.cpp
```diff
@@ -0,0 +1,41 @@
+#include "Shape.h"
+
+Shape::Shape()
```
AntonPiccardoSelg: why do you specify a default constructor?

exercises/matthew_bowles/ex02_oo_basics/inc/Shape.h
```diff
@@ -0,0 +1,26 @@
...
+	virtual ~Shape();
+
+	std::string getType() const;
+	int getSideCount() const;
+	virtual double calcArea() const;
+	virtual double calcPerimeter() const;
+	virtual void print() const;
+
+private:
+	int sideCount;
```
AntonPiccardoSelg: member variables should start with `m_`

exercises/matthew_bowles/ex02_oo_basics/inc/Rectangle.h
```diff
@@ -0,0 +1,19 @@
...
+	virtual double calcPerimeter() const override;
+	virtual void print() const override;
+
+private:
+	double side1;
```
AntonPiccardoSelg: should start with `m_`

exercises/matthew_bowles/ex02_oo_basics/inc/ShapeSorter.h
```diff
@@ -0,0 +1,27 @@
...
+	void sortArea(vector<Shape*>& shapes);
+	void sortPerimeter(vector<Shape*>& shapes);
+
+private:
+	void printShapes(const vector<Shape*> &shapes);
+	vector<Shape> shapes;
+};
+
+bool compareArea(const Shape* x, const Shape* y);
+bool comparePerimeter(const Shape* x, const Shape* y);
```
AntonPiccardoSelg: It makes sense to have these as private functions

exercises/matthew_bowles/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,108 @@
...
+* @param shapes The vector of shape pointers to traverse
+*
+*/
+void ShapeSorter::matchType(const string& type, vector<Shape*> shapes)
+{
+	auto iter = shapes.begin();
+	while (iter != shapes.end())
+	{
+		if (type.compare((*iter)->getType()) != 0)
+			iter = shapes.erase(iter);
```
AntonPiccardoSelg: You want to avoid erasing in a vector. See [here](https://stackoverflow.com/questions/4663455/why-is-erase-function-so-expensive).

A solution would be to populate  a new vector with pointers with the shapes of interest. Also try to avoid too many iterators. Range based for loops are more readable.

You can do simple string comparison with `==`

exercises/matthew_bowles/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,108 @@
...
+* @param shapes The vector of shape pointers to traverse
+*
+*/
+void ShapeSorter::matchSides(int sides, vector<Shape*> shapes)
+{
+	auto iter = shapes.begin();
+	while (iter != shapes.end())
+	{
+		if ((*iter)->getSideCount() != sides)
+			iter = shapes.erase(iter);
```
AntonPiccardoSelg: See above

exercises/matthew_bowles/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+/** prints out a vector of shapes
+*
+* @param shapes The vector of shape pointers to print
+*
+*/
+void ShapeSorter::printShapes(const vector<Shape*>& shapes)
+{
+	auto iter = shapes.cbegin();
+	while (iter != shapes.cend())
```
AntonPiccardoSelg: range-based foor loop

exercises/matthew_bowles/ex01_basics/inc/main.h
AntonPiccardoSelg: You don't find `main.h` often. It would be good to put your functions into a separate `.h/.cpp` , eg `WordManipulation.h` and `WordManipulation.cpp`.

AntonPiccardoSelg (changes requested):
Hi, there are some corrections for both exercises. Please let me know if something is unclear. 

I have not performed functional testing. I will do this once the code changes have been applied.

exercises/matthew_bowles/ex02_oo_basics/inc/Circle.h
MatthewBowles: Seems to be needed for M_PI constant; get errors during build without it

AntonPiccardoSelg: Yes, but nothing in the header file relies on `math.h`. You should only include it in the `.cpp` file in this case.

MatthewBowles: Oh, I see. Thank you very much for your help.
All of these changes should be done now, let me know if I missed anything.

exercises/matthew_bowles/ex01_basics/src/WordManipulation.cpp
```diff
@@ -0,0 +1,56 @@
...
+	string punctuation = ".,?'\"!():;";
+	auto iter = word.begin();
+	while (iter != word.end())
+	{
+		if (punctuation.find(*iter) != string::npos)
+			iter = word.erase(iter);
+		else
+			++iter;
+	}
+
```
AntonPiccardoSelg: You could consider [this](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom)

exercises/matthew_bowles/ex02_oo_basics/inc/Shape.h
```diff
@@ -0,0 +1,26 @@
...
+{
+public:
+	Shape(std::string type, int sideCount);
+	virtual ~Shape();
+
+	std::string getType() const;
+	int getSideCount() const;
+	virtual double calcArea() const;
+	virtual double calcPerimeter() const;
+	virtual void print() const;
```
AntonPiccardoSelg: I should have mentioned this the first time round, but this class should be pure abstract class, ie methods such as `calcArea` should not have a dummy implementation (rather they should be `virtual calcArea() const = 0;. You should not allow `Shape` to be created it self.

exercises/matthew_bowles/ex02_oo_basics/inc/main.h
```diff
@@ -0,0 +1,5 @@
+#ifndef MAIN_H
```
AntonPiccardoSelg: No `main.h` is needed. Please remove it

exercises/matthew_bowles/ex02_oo_basics/inc/Shape.h
```diff
@@ -0,0 +1,26 @@
...
+#define SHAPE_H
+
+#include <string>
+
+
+class Shape
+{
+public:
+	Shape(std::string type, int sideCount);
+	virtual ~Shape();
```
AntonPiccardoSelg: You can use the `=default` approach here.

AntonPiccardoSelg (changes requested):
Hi, there are just a few more changes which we should implement. Please see the comments. Thanks.

MatthewBowles:
@AntonPiccardoSelg I've implemented the changes.

exercises/matthew_bowles/ex02_oo_basics/src/Shape.cpp
```diff
@@ -0,0 +1,35 @@
...
+{
+	return m_type;
+}
+
+int Shape::getSideCount() const
+{
+	return m_sideCount;
+}
+
+double Shape::calcArea() const
```
AntonPiccardoSelg: This method definition is not needed.

exercises/matthew_bowles/ex02_oo_basics/src/Shape.cpp
```diff
@@ -0,0 +1,35 @@
...
+{
+	return 0;
+}
+
+double Shape::calcPerimeter() const
```
AntonPiccardoSelg: This method definition is not needed.

exercises/matthew_bowles/ex02_oo_basics/src/Shape.cpp
```diff
@@ -0,0 +1,35 @@
...
+{
+	return 0;
+}
+
+void Shape::print() const
```
AntonPiccardoSelg: This method definition is not needed.

AntonPiccardoSelg (changes requested):
Hi, just one last tiny change. The method definitions in `shape.cpp` are not required.

MatthewBowles:
@AntonPiccardoSelg Done, thanks.

AntonPiccardoSelg:
Hi @MatthewBowles, i am happy with the code changes. The summary on the top of the PR is very nice. Also providing such a good sample file helps a lot when testing :+1: 

The output of exercise 1 looked a bit odd though. There are some entries which are listed twice and the counts seem to be shifted (e.g. `holmes`):
```
Word           Usage

which             49
holmes            38
there             29
could             23
would             19
majesty           16
street            16
photograph        14
little            14
should            13
irene             13
house             13
other             12
before            12
woman             11
nothing           11
holmes
           11
myself            11
sherlock          11
indeed            10
...
```
Do you have any idea why this is the case?

MatthewBowles:
@AntonPiccardoSelg thanks again for your feedback :+1: 
That output is very unusual. When I run it locally I get this:
```
Word           Usage

which             55
holmes            49
there             32
could             25
photograph        21
would             20
street            18
majesty           16
house             15
should            14
little            14
before            14
adler             13
irene             13
minutes           13
other             12
sherlock          12
woman             12
matter            11
indeed            11
myself            11
briony            11
nothing           11
lodge             11
...
```

AntonPiccardoSelg:
Hmm, odd. Which OS are you using?

MatthewBowles:
I'm using Windows

AntonPiccardoSelg:
Hmm, I am using Ubuntu 16.04. 

@gemmaguest Could you check that exercise 1 executes correctly? Thanks

exercises/matthew_bowles/ex01_basics/src/Io.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+	output.open("Output.txt");
+	output << "Word" << setw(16) << "Usage" << "\n" << "\n";
+
+	for(auto& w : words)
+	{
+		word = w.first;
+		output << word << setw(20 - word.length()) << w.second << '\n';
+	};
+	return;
```
AntonPiccardoSelg, line 124: You don't need the `return` at this point. Also the semi-colon of the for loop is unnecessary.  A change is not required for this though. Just something to keep in mind.

gemmaguest:
@AntonPiccardoSelg I get the same output as Matt. I'm on Windows 7.

AntonPiccardoSelg (approved):
@gemmaguest Thanks for checking that :+1:

Changes look fine. Happy to pass on to gatekeeper

exercises/matthew_bowles/ex01_basics/inc/Io.h
```diff
@@ -0,0 +1,22 @@
+#ifndef IO_H
+#define IO_H
+
+#include <string>
+#include <map>
+#include <vector>
+
+using namespace std;
```
gemmaguest, line 8: No need to change it now, but note that in Mantid we avoid `using namespace std`, especially in header files. See [here](http://llvm.org/docs/CodingStandards.html#do-not-use-using-namespace-std). If this makes the code too ugly/unreadable, you can use typedefs for commonly-occuring types, e.g.
```
typedef std::map<std::string, int> WordCounts
typedef std::vector<std::pair<std::string, int>> SortedWordCounts
...
//sorts map of words into descending order by value
SortedWordCounts sortWords(const WordCounts& words);
```

gemmaguest (approved):
Looks good to me :+1:. Happy to merge. Just a couple of general tips:

- It's a good idea to set up your text editor to [use spaces instead of tabs](http://llvm.org/docs/CodingStandards.html#use-spaces-instead-of-tabs)
- Some of the commit messages could be more descriptive. It's good to get into the habit of [writing good commit messges](https://chris.beams.io/posts/git-commit/) even for small changes.

PR #30, opened 2017-07-07
Brandon Hewer: Ex01 and Ex02

brandonhewer:
## Exercise 1
A unique word counter for ASCII-formatted files. Counts the number of occurrences of each unique word (of length greater than 4). Case and punctuation insensitive; punctuation is considered to be the set of characters ".,?'"!():"

**To test:**
  1. Build the WordCounter project.
  2. a) Run the build, passing the directory of the ASCII file to count from as an argument.
      b) To check error handling, pass in an invalid directory.
      c) To check error handling, do not pass an argument when running the build.
  3. Once application has quit, open the 'results.txt' file in the build directory.
  4. Check to see that the 'results.txt' file contains a list of unique words and their total number of
      occurrences each on a new line.

## Exercise 2
An application for creating, sorting and filtering 2-dimmensional shapes. Shapes included: squares, rectangles, circles, triangles. Provides sorting by perimeter and area, as well as filtering by type (name) and number of sides.

**To test:**
  1. Build the Shapes project.
  2. a) Construct one of each shape (square, rectangle, circle and triangle), by calling their respective
          constructors with positive double values as arguments (one for circle and square, two for
          rectangle and triangle).
      b) To test error handling, try constructing the shapes by passing negative double values as arguments
           to their constructors (Note: A Runtime Exception should be thrown).
  3. Create a vector of pointers to the constructed Shapes.
  4. Construct a ShapeSorter by calling its constructor and passing in the previously created vector as
      an argument.
  5. a) Print out the returned values upon calling the filter methods, filterByType and filterBySideNumber 
          when passing in each type ('Circle', 'Square', 'Triangle, 'Rectangle') and side number (1, 3, 4) 
          respectively.
      b) Print out the returned values upong calling the same methods with types or side numbers which
          do not correspond to any shapes (should not display any shapes).
  6. Print out the returned values upon calling the sort methods, sortByPerimeter and sortByArea, when
      passing in both directional values ShapeSorter::ASCENDING (true) and ShapeSorter::DESCENDING 
      (false).


exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+/*
+ * Start of execution.
+ */
+int main(int num_args, char ** args) {
+
+	// Check whether an argument has been passed (1 default argument
+	// is always passed; working directory).
+	if (num_args == 1) {
+		cerr << "No file path was given as an argument. Program Exiting..." << endl;
+		return 0;
```
AntonPiccardoSelg: You could pass a `1` return value. Also below.

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+		// Count the number of occurrences of each unique word.
+		unordered_map<string, int> count = countUniqueWords(words, word_cond);
+
+		// Output the total occurrences of each unique word in a results file.
+		ofstream ofile("results.txt");
+		ofile << "Word\tUsage\n" << flip_map(count);
+		ofile.close();
+
+		return 1;
```
AntonPiccardoSelg: Wouldn't you return 0 here?

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+		unordered_set<char> delims = { '.', ',', '?', '!', '(', ')', ':', '\'', 
+			'"', ' ', ';', '\t'};
+		auto word_cond = [](const string s) -> const bool {
+			return s.size() >= MIN_WORD_LENGTH;
+		};
+
+		// Retrieve words from input file.
+		vector<string> words = getWords(ifile, delims);
+		ifile.close();
```
AntonPiccardoSelg: I don't think you have to close it. It uses [RAII](http://en.cppreference.com/w/cpp/language/raii), hence it's destructor will handle freeing the resource.

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+		// Check whether a file exists at the first given argument.
+		if (!ifile) {
+			cerr << "No file exists at the specified path. Program Exiting..." << endl;
+			return 0;
+		}
+
+		unordered_set<char> delims = { '.', ',', '?', '!', '(', ')', ':', '\'', 
+			'"', ' ', ';', '\t'};
+		auto word_cond = [](const string s) -> const bool {
```
AntonPiccardoSelg: take the string via `const &`. See [here](https://stackoverflow.com/questions/2627166/difference-between-const-reference-and-normal-parameter). This is quite important or else you will make unnecessary copies.

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+		vector<string> words = getWords(ifile, delims);
+		ifile.close();
+
+		// Count the number of occurrences of each unique word.
+		unordered_map<string, int> count = countUniqueWords(words, word_cond);
+
+		// Output the total occurrences of each unique word in a results file.
+		ofstream ofile("results.txt");
+		ofile << "Word\tUsage\n" << flip_map(count);
+		ofile.close();
```
AntonPiccardoSelg: See above

exercises/brandon_hewer/ex01_basics/src/main.h
```diff
@@ -0,0 +1,51 @@
...
+/// Splits a string \p str, at the delimiters specified in \p delimiters and
+/// stores the result in the specified vector \p words.
+std::vector<std::string> split(const std::string str, const std::unordered_set<char>& delimiters,
+	std::vector<std::string> words);
+
+/// Overloading the << operator to allow for inserting multimaps with int keys 
+/// and string values into output streams.
+std::ostream &operator<<(std::ostream &os, const std::multimap<int, std::string>& mp);
+
+#endif /* WORD_COUNTER_H */
```
AntonPiccardoSelg: new line

exercises/brandon_hewer/ex01_basics/src/main.h
```diff
@@ -0,0 +1,51 @@
...
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file contains the declaration for the methods contained in the
+/// main.cpp file.
+///
+//===----------------------------------------------------------------------===//
+#ifndef WORD_COUNTER_H
+#define WORD_COUNTER_H
```
AntonPiccardoSelg: :+1:

exercises/brandon_hewer/ex01_basics/src/main.h
```diff
@@ -0,0 +1,51 @@
...
+	const bool cond(std::string));
+
+/// Retrieves all words from the input stream \p in, where words
+/// are defined as the vector created when the input is split by
+/// the delimiters specified in \p delimiters.
+std::vector<std::string> getWords(std::istream& in, const std::unordered_set<char>& delimiters);
+
+/// Splits a string \p str, at the delimiters specified in \p delimiters and
+/// stores the result in the specified vector \p words.
+std::vector<std::string> split(const std::string str, const std::unordered_set<char>& delimiters,
```
AntonPiccardoSelg: as mentioned above `const &`

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+ * vector \p words.
+ *
+ * \param str			The string to split by the delimiters in \p delimiters.
+ *
+ * \param delimiters	The delimiters to use in splitting the string \p str.
+ *
+ * \param words			The vector in which to store the resulting sections of
+ *					    the split string.
+ */
+vector<string> split(const string str, const unordered_set<char>& delimiters,
```
AntonPiccardoSelg: `const &`

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	// Check whether the final string section is not empty.
+	// If true, add the final string section to \p words.
+	if (c_str.length() != 0) {
+		words.push_back(c_str);
+	}
+
+	return words;
+}
```
AntonPiccardoSelg: new line

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+ *					    the split string.
+ */
+vector<string> split(const string str, const unordered_set<char>& delimiters, 
+	vector<string> words = vector<string>()) {
+	// c_str stores the currently built string section, since the last
+	// delimiter was found.
+	string c_str;
+
+	// Iterate through all characters in the specified string \p str.
+	for (auto i = str.begin(); i != str.end(); ++i) {
```
AntonPiccardoSelg: you can use a range-based for loop here

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+ * \param delimiters	The delimiters to use in splitting the string \p str.
+ *
+ * \param words			The vector in which to store the resulting sections of
+ *					    the split string.
+ */
+vector<string> split(const string str, const unordered_set<char>& delimiters, 
+	vector<string> words = vector<string>()) {
+	// c_str stores the currently built string section, since the last
+	// delimiter was found.
+	string c_str;
```
AntonPiccardoSelg: variables in mantid are normally camel case. The only execption are member variables which have `m_...`

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * \param in			The input stream.
+ *
+ * \param delimiters	The delimiters used to split the input stream into
+ *						words. Defaults to an empty set, in which case, each
+ *                      line in the input stream will be considered a word.
+ *
+ * \returns				The vector containing all words from the input stream.
+ */
+vector<string> getWords(istream& in, const unordered_set<char>& delimiters = {}) {
```
AntonPiccardoSelg: Not sure that one would have a scenario with no delimiters. what do you think?

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+ *              total number of occurrences as the associated
+ *              values.
+ */
+unordered_map<string, int> countUniqueWords(const vector<string>& words, const bool cond(string)) {
+	unordered_map<string, int> count;
+
+	// Iterates over each value in the specified vector \p words.
+	for (auto i = words.begin(); i != words.end(); ++i) {
+		string word = *i;
+		unordered_map<string, int>::iterator m_iter = count.find(*i);
```
AntonPiccardoSelg: Make use of the methods on the data structures. Why not use ` if (count.count() > 0) {...` ?

AntonPiccardoSelg: You seem to be case sensitive and you are not ignoring the punctuation. This needs to be added.

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+ * \param mp	The multimap whose data to add to the output stream.
+ *
+ * \returns		The output stream, with the data from the output stream
+ *              added.
+ */
+ostream &operator<<(ostream &os, const multimap<int, string>& mp) {
+	multimap<int, string> mmp;
+	
+	// Iterates over each key-value entry in the specified map /p mp.
+	for (auto i = mp.rbegin(); i != mp.rend(); ++i) {
```
AntonPiccardoSelg: range based for loop `for (const auto& element : ...`

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+ * order.
+ *
+ * \param os	The output stream to add to.
+ * \param mp	The multimap whose data to add to the output stream.
+ *
+ * \returns		The output stream, with the data from the output stream
+ *              added.
+ */
+ostream &operator<<(ostream &os, const multimap<int, string>& mp) {
+	multimap<int, string> mmp;
```
AntonPiccardoSelg: Use variable names which describe what the variable refers to (not the data structure). Also try to avoid abbreviations (unless for things like `i` as a loop counter and so on which are typical use cases).

AntonPiccardoSelg (changes requested):
Note this a review for the first exercise only. Once the changes have been made, the second exercise will be reviewed.

### General
Please see the comments in the code. One general thing to be aware of is pass-by-value, pass-by-ref and pass-by-const-ref; see [here](https://stackoverflow.com/questions/2582797/why-pass-by-const-reference-instead-of-by-value). 

### Test results
I was able to read in the file and it outputs the counts . It handles bad input and provides me with a sensible error report. You could consider adding formatting to your output though.

### Errors
Note that your implementation is case-sensitive and does not consider special characters. Please see the instructions of the starter exercise for more info.

### Regarding testing
Nobody has mentioned this yet but in general you would also provide test instructions for the reviewer. In this case it would contain for example a test file and a step-by-step instruction what to test. It is then up to the reviewer then to perform additional checks.

#### Summary

Please provide the requested changes and I will have another look over the PR then. If you have any questions, come and speak to me.

exercises/brandon_hewer/ex01_basics/src/main.cpp
AntonPiccardoSelg: Actually the delimiters should be space and hyphen, no?

exercises/brandon_hewer/ex02_oo_basics/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+  src/main.cpp
+  src/Circle.cpp
+  src/Square.cpp
+  src/Rectangle.cpp
+  src/Triangle.cpp
+  src/ShapeSorter.cpp
+)
+
+set ( HDR_FILES
+  src/Circle.h
```
AntonPiccardoSelg: :+1: for registering the header files (else they won't show up in Visual Studio).
You should put them in a separate folder though, something like `inc`

exercises/brandon_hewer/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/*
+ * @return	The string representation of this circle.
+ */
+std::string Circle::toString() const {
+	std::ostringstream ss;
+	ss << "Circle | Radius: " << m_radius;
+	return ss.str();
+}
```
AntonPiccardoSelg: new line

exercises/brandon_hewer/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,72 @@
...
+		throw ss.str();
+		m_radius = radius;
+	}
+	m_radius = radius;
+}
+
+/*
+ * Destructor for this circle.
+ */
+Circle::~Circle() {
```
AntonPiccardoSelg: No need to specify the descructor

exercises/brandon_hewer/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @throw invalid_argument	Thrown if specified radius is less than zero.
+ */
+Circle::Circle(double radius) {
+	
+	if (radius < 0) {
+		std::ostringstream ss;
+		ss << "Radius of a " << name() << " must be greater than or equal"
+			"to 0";
+		throw ss.str();
```
AntonPiccardoSelg: It is very good that you checked :+1: 

You should throw however an appropriate error class, eg.

exercises/brandon_hewer/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @param radius	The radius of the circle to construct.
+ *
+ * @throw invalid_argument	Thrown if specified radius is less than zero.
+ */
+Circle::Circle(double radius) {
+	
+	if (radius < 0) {
+		std::ostringstream ss;
+		ss << "Radius of a " << name() << " must be greater than or equal"
```
AntonPiccardoSelg: Avoid using string streams. You can concatenate strings by adding them

exercises/brandon_hewer/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,72 @@
...
+int Circle::numberOfSides() const {
+	return 1;
+}
+
+/*
+ * @return	The string representation of this circle.
+ */
+std::string Circle::toString() const {
+	std::ostringstream ss;
+	ss << "Circle | Radius: " << m_radius;
```
AntonPiccardoSelg: same as above. Check out [`std::stod`](http://www.cplusplus.com/reference/string/to_string/)

exercises/brandon_hewer/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,52 @@
...
+//===----------------------------------------------------------------------===//
+class Circle :
+	public Shape
+{
+public:
+	/// Constructs a circle with a specified radius.
+	Circle(double radius);
+
+	/// Destructor.
+	~Circle();
```
AntonPiccardoSelg: not needed.

exercises/brandon_hewer/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,52 @@
...
+	int numberOfSides() const;
+
+	/// Returns the string representation of the circle.
+	std::string toString() const;
+private:
+	/// The radius of the circle.
+	double m_radius;
+};
+
+#endif /* CIRCLE_H */
```
AntonPiccardoSelg: new line

exercises/brandon_hewer/ex02_oo_basics/src/Rectangle.cpp
```diff
@@ -0,0 +1,79 @@
...
+	if (width < 0) {
+		std::ostringstream ss;
+		ss << "Width of a " << name() << " must be greater than or equal to 0.";
+		throw ss.str();
+	}
+	else if (height < 0) {
+		std::ostringstream ss;
+		ss << "Height of a " << name() << " must be greater than or equal to 0.";
+		throw ss.str();
+	}
```
AntonPiccardoSelg: see above

exercises/brandon_hewer/ex02_oo_basics/src/Rectangle.cpp
```diff
@@ -0,0 +1,79 @@
...
+	return 4;
+}
+
+/*
+* @return	The string representation of this rectangle.
+*/
+std::string Rectangle::toString() const {
+	std::ostringstream ss;
+	ss << "Rectangle | Width: " << m_width << ", Height: " << m_height;
+	return ss.str();
```
AntonPiccardoSelg: see above

exercises/brandon_hewer/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,41 @@
...
+
+	/// Returns the name of the shape.
+	virtual std::string name() const = 0;
+
+	/// Returns the number of sides of the shape.
+	virtual int numberOfSides() const = 0;
+
+	/// Returns the string representation of the shape.
+	virtual std::string toString() const = 0;
+};
```
AntonPiccardoSelg: You are using polymorphism. you require a virtual destructor. See [here](https://stackoverflow.com/questions/461203/when-to-use-virtual-destructors). Please do a bit of research about this. This is easy to get wrong but very important

exercises/brandon_hewer/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,52 @@
...
+{
+public:
+	/// Constructs a circle with a specified radius.
+	Circle(double radius);
+
+	/// Destructor.
+	~Circle();
+
+	/// Returns the perimeter of the circle.
+	double perimeter() const;
```
AntonPiccardoSelg: please use the `override` specifier. See [here](http://en.cppreference.com/w/cpp/language/override). This is true for the other classes

exercises/brandon_hewer/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,187 @@
...
+ * @param shapes	A vector containing pointers to shape objects.
+ */
+ShapeSorter::ShapeSorter(const std::vector<Shape*>& shapes) {
+	m_shapes = shapes;
+}
+
+/*
+ * Destructor for this ShapeSorter.
+ */
+ShapeSorter::~ShapeSorter() {
```
AntonPiccardoSelg: not needed

exercises/brandon_hewer/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	return ShapeSorter(filtered);
+}
+
+/*
+ * @return	The string representation of this ShapeSorter.
+ */
+std::string ShapeSorter::toString() const {
+	std::ostringstream ss;
```
AntonPiccardoSelg: don't use a string stream. Just concatenate the strings

exercises/brandon_hewer/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,187 @@
...
+ *
+ * @return			A ShapeSorter for the filtered vector of
+ *					shapes.
+ */
+ShapeSorter ShapeSorter::filterBy(std::function<bool (const Shape*)> filter) const {
+	std::vector<Shape*> filtered;
+
+	// Iterate through all shapes in this ShapeSorter's vector of
+	// shapes.
+	for (auto i = m_shapes.begin(); i != m_shapes.end(); ++i) {
```
AntonPiccardoSelg: range based for loop should work here

exercises/brandon_hewer/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,187 @@
...
+ * Filters the vector of shapes by their type (name).
+ *
+ * @param type	The type to filter by.
+ *
+ * @return		A ShapeSorter for the filtered vector of shapes.
+ */
+ShapeSorter ShapeSorter::filterByType(std::string type) const {
+
+	// Create type filter function.
+	auto filter = [&type](const Shape* s) -> bool {
```
AntonPiccardoSelg: It might have been good to use an enum class instead of a string in the shape classes. This would make the comparison much more efficient.

exercises/brandon_hewer/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,187 @@
...
+ * @param direction	The direction in which to sort:
+ *					ShapeSorter::ASCENDING (true) - Sort by increasing perimeter.
+ *					ShapeSorter::DESCENDING (false) - Sort b
+ *
+ * @return			A ShapeSorter for the sorted vector of shapes.
+ */
+ShapeSorter ShapeSorter::sortByPerimeter(bool direction) const {
+	// Copy vector of shapes.
+	std::vector<Shape*> n_shapes(m_shapes);
+	bool(*sorter) (Shape*, Shape*);
```
AntonPiccardoSelg: a lot of the code can be shared with the method above

exercises/brandon_hewer/ex02_oo_basics/src/Rectangle.h
```diff
@@ -0,0 +1,54 @@
...
+	~Rectangle();
+
+	/// Returns the perimeter of the rectangle.
+	double perimeter() const;
+
+	/// Returns the area of the rectangle.
+	double area() const;
+
+	/// Returns the name of the rectangle.
+	virtual std::string name() const;
```
AntonPiccardoSelg: virtual is not needed, but it should be marked `override` (see above0

AntonPiccardoSelg (changes requested):
Good commenting. Very structured. At times it might be a bit too detailed. I don't want to put you off of commenting (it is better to comment too much that too little), but try to consider the complexity of the piece of code and check a long or a short comment is warranted. Note that variable naming is a (good) form of commenting and that actual comments are sometimes ignore by our brains. Whenever you have time I recommend reading `Code Complete` which talks about this.

Other than that there are some points which should be addressed. Please speak to me if you have any questions.

AntonPiccardoSelg:
:+1: Good test instructions

exercises/brandon_hewer/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,259 @@
...
+ * Converts the specified string \p str to lowercase.
+ *
+ * @param str	The string to convert to lowercase.
+ */
+string toLower(const string& str) {
+	string newStr;
+
+	// Iterate over each character in the specified
+	// string.
+	for (const auto& c : str) {
```
AntonPiccardoSelg, line 184: This is fine, but you could copy the string and use `std::transform` with `tolower`. Anytime you can use something from standard library you should.

exercises/brandon_hewer/ex02_oo_basics/inc/Shape.h
```diff
@@ -0,0 +1,50 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Enumerator for the different types of shape.
+enum class ShapeType { Square, Circle, Triangle, Rectangle };
+
+class Shape
+{
+public:
+	/// Shape destructor.
+	virtual ~Shape() {};
```
AntonPiccardoSelg, line 29: Hint: You can use `=default;`

AntonPiccardoSelg (approved):
Looks fine to me. There are some minor comments to consider for the future. 
Happy to pass this on to the gate keeper. 

Hi @gemmaguest, this is ready for gate-keeping

gemmaguest (approved):
Looks good to me :+1:. Happy to merge. Just a couple of general tips:

- It's a good idea to set up your text editor to [use spaces instead of tabs](http://llvm.org/docs/CodingStandards.html#use-spaces-instead-of-tabs)
- Some of the commit messages could be more descriptive. It's good to get into the habit of [writing good commit messges](https://chris.beams.io/posts/git-commit/) even for small changes.

PR #32, opened 2018-04-25
T jubb exercises

TWJubb:
Exercise 1

### DIFF
--- a/exercises/t_jubb/ex01_basics/src/find_words.cpp
+++ b/exercises/t_jubb/ex01_basics/src/find_words.cpp
@@ -1,0 +1,114 @@
+#include <iostream>
+#include <string>
+#include <vector>
+#include <fstream>
+#include <map>
+
+using std::cout;
+using std::endl;
+
+
+std::vector<int> find_spaces(std::string input_string){
+  /*
+    Finds all the spaces in a string.
+
+    @param input_string, a string of characters.
+    @return spaces, a vector of ints, the indices of the spaces.
+  */
+  std::vector<int> spaces;
+  int index = 0;
+  spaces.push_back(0);
+  for ( std::string::iterator i = input_string.begin(); i != input_string.end(); i++ ) {
+    index++;
+    if (isspace(*i) != 0){
+      spaces.push_back(index);
+      std::cout << "Found a space!" << std::endl;
+    }
+    std::cout << (*i) << std::endl;
+  }
+  return spaces;
+}
+
+
+
+std::vector<std::string> get_words(std::string input_string) {
+  /*
+    Splits input_string up into its components words by detecting
+    any spaces in the string.
+
+    @param input_string a string with spaces.
+    @return a vector of strings (the words).
+  */
+  std::vector<std::string> words;
+  std::vector<int> space_locations;
+  space_locations = find_spaces(input_string);
+
+  for (int i = 0; i < space_locations.size(); i++) {
+    int word_length = (space_locations[i+1]-space_locations[i]);
+    if (word_length > 1) {
+      std::string word = input_string.substr(space_locations[i], word_length);
+      words.push_back(word);
+    };
+  };
+  return words;
+};
+
+std::map<std::string, int> word_counts(std::map<std::string, int> current_counts , std::vector<std::string> word_vec  ) {
+  /*
+    Adds a vector of words into a map of word:occurance pairs.
+
+    @param current_counts, a map of word:occurance pairs.
+    @param word_vec a vector of words.
+    @return current_counts after the words in word_vec have been added.
+  */
+  for ( std::vector<std::string>::iterator i = word_vec.begin(); i != word_vec.end(); i++ ) {
+      ++current_counts[(*i)];
+  };
+  return current_counts;
+};
+
+
+
+
+int main()
+{
+
+std::string file_path = "words.txt";
+
+std::fstream myFile;
+myFile.open(file_path,std::ios_base::in);
+
+if (myFile.is_open()) {
+
+  cout << "File " << file_path << " opened successfully!" << std::endl;
+  std::string file_line;
+  int line = 0;
+
+
+  while (myFile.good()) {
+    line++;
+    std::getline(myFile,file_line);
+    cout << line << " : " << file_line << endl;
+
+  }
+
+
+  myFile.close();
+}
+
+std::vector<std::string> words;
+words = get_words("hello world    ");
+// std::vector<int> spaces;
+// spaces = find_spaces("hello ");
+
+int counter = 0;
+for ( auto i = words.begin(); i != words.end(); i++ ) {
+    counter++;
+    std::cout << counter << " : " << *i << std::endl;
+}
+
+
+std::map<std::string, int> counts;
+counts = word_counts(counts , words);
+
+}
--- a/exercises/t_jubb/ex01_basics/src/find_words.cpp
+++ b/exercises/t_jubb/ex01_basics/src/find_words.cpp
@@ -68,15 +68,25 @@ int add_word_counts(std::map<std::string, int>& current_counts , std::vector<std
 
 int main()
 {
-// Map to store the word counts
+const bool verbose = false;
+
+// Map to store the words and counts and key/value pairs
 std::map<std::string, int> counts;
 
+// Prompt user for a filepath and check if file exists
 std::string file_path;
-/*
-
-Add prompt for user input of file name
-
-*/
+cout << "Please enter a valid filepath : " ;
+bool valid_file = false;
+while (!valid_file){
+  std::cin >> file_path;
+  std::ifstream infile(file_path);
+  if (!infile.good() ){
+    cout << "File does not exist, please enter a valid file : ";
+  }
+  else {
+    valid_file = true;
+  }
+}
 
 // Import test file
 file_path = "words.txt";
@@ -92,15 +102,19 @@ if (myFile.is_open()) {
   while (myFile.good()) {
     line++;
     std::getline(myFile,file_line);
-    cout << "\tFile Line " << line << " : " << file_line << endl;
+    if (verbose == true) {
+      cout << "\tFile Line " << line << " : " << file_line << endl;
+    }
 
     // count the words
     std::vector<std::string> words;
     words = get_words(file_line, 4);
-    int counter = 0;
-    for ( auto i = words.begin(); i != words.end(); i++ ) {
-        counter++;
-        std::cout << counter << " : " << *i << std::endl;
+    if (verbose == true) {
+      int counter = 0;
+      for ( auto i = words.begin(); i != words.end(); i++ ) {
+          counter++;
+          std::cout << counter << " : " << *i << std::endl;
+      }
     }
     add_word_counts(counts , words);
   }
@@ -121,19 +135,9 @@ for ( std::map<std::string, int>::iterator i = counts.begin(); i != counts.end()
 
     // Pad the index of the words
     int num_pad = 3; 
-    
-    std::cout << std::setfill('0') << std::setw(num_pad) << counter << " : ";
-    std::cout << (*i).first << pad_blank << (*i).second << std::endl;
-}
 
-// // Test code
-// std::vector<std::string> words;
-// words = get_words("hello world    ");
-// int counter = 0;
-// for ( auto i = words.begin(); i != words.end(); i++ ) {
-//     counter++;
-//     std::cout << counter << " : " << *i << std::endl;
-// }
-// counts = word_counts(counts , words);
+    cout << std::setfill('0') << std::setw(num_pad) << counter << " : ";
+    cout << (*i).first << pad_blank << (*i).second << endl;
+}
 
 }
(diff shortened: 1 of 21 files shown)


exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
+#include <iostream>
```
martyngigg, line 1: As a stylistic point it is good practice to order the includes alphabetically so that it is easy to find if one is already there or not.

exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <iomanip>
+#include <algorithm>
+#include <functional>
+
+using std::cout;
+using std::endl;
+
+
+std::vector<std::string> get_words(const std::string input_string, const int min_len) {
+  /*
```
martyngigg, line 16: This a good Doxygen documentation string 👍 

The only style point is that it should come before the function definition not after like Python.

exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <set>
+#include <iomanip>
+#include <algorithm>
+#include <functional>
+
+using std::cout;
+using std::endl;
+
+
+std::vector<std::string> get_words(const std::string input_string, const int min_len) {
```
martyngigg, line 15: C++ has something called value semantics which means that by default an argument to a function will be copied on input. To prevent this you can define the function to take a reference to an object by adding an ampersand after the type like so:
```c++
std::vector<std::string> get_words(const std::string & input_string, const int min_len) {
```
The input_string will then be a reference to the original object.

exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
...
+    @return a vector of strings (the words).
+  */
+  std::vector<std::string> words;
+  
+  int index      = -1;
+  int word_start = 0;
+  int word_end   = 0;
+  int word_len   = 0;
+
+  for ( std::string::const_iterator i = input_string.begin(); i <= input_string.end(); i++ ) {
```
martyngigg, line 31: The slightly more modern way to write this would be
```
for ( auto iter = std::cbegin(input_string); iter != std::cend(input_string); ++iter)
```
`auto` let's the compiler deduce the type rather than you having to spell it out.

exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
...
+    @return current_counts after the words in word_vec have been added.
+  */
+  for ( std::vector<std::string>::const_iterator i = word_vec.begin(); i != word_vec.end(); i++ ) {
+      ++current_counts[(*i)];
+  };
+  // return current_counts;
+  return 0;
+};
+
+int ordered_by_value(std::vector< std::pair<  int , std::string > >& vec ,std::map<std::string, int >& count_map) {
```
martyngigg, line 71: `C++11` no longer requires spaces between the angle brackets so the whitespace can be removed here to make it easier to read.

exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
...
+  std::sort(vec.begin(),vec.end());
+  return 0;
+}
+
+int main()
+{
+const bool verbose = false;
+
+// Map to store the words and counts and key/value pairs
+std::map<std::string, int> counts;
```
martyngigg, line 96: Is the order important here? `std::map` sorts on the key by default. If order is not important, which I suspect it is not, then `std::unordered_map` is better as it's just a hash table for fast look up.

exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// Order the pairs by value using default vector sort
+std::vector< std::pair<  int , std::string > > ordered;
+ordered_by_value(ordered, counts);
+
+// Display the ordered words and counts.
+cout << endl << "Ordered By Occurance : " << endl << endl;
+counter = 0;
+for ( std::vector< std::pair<  int , std::string > >::iterator i = ordered.begin(); i != ordered.end(); i++ ) {
```
martyngigg, line 172: Modern C++ would allow you to do `for ( const auto & elem : ordered )` to make this easier to read.

exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
...
+    counter++;
+
+    // Pad the gap between word and count.
+    int pad = 15 - (*i).second.size();
+    std::string pad_blank = " " ;
```
martyngigg, line 177: This could be done without the for loop using one of the string constructors (http://en.cppreference.com/w/cpp/string/basic_string):
```
std::string pad_blank(pad, ' ');
```

exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
...
+    // Pad the index of the words
+    int num_pad = 3; 
+
+    cout << std::setfill('0') << std::setw(num_pad) << counter << " : ";
+    cout << (*i).first << pad_blank << (*i).second << endl;
+}
+
+// Order the pairs by value using default vector sort
+std::vector< std::pair<  int , std::string > > ordered;
+ordered_by_value(ordered, counts);
```
martyngigg, line 167: I this would be better written as 
```
auto ordered = ordered_by_value(counts);
```
and have the function return a vector that it creates. Modern C++ has mechanisms to avoid copies in this type of case.

exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  cout << "File " << file_path << " opened successfully!" << std::endl;
+  std::string file_line;
+  int line = 0;
+
+  while (myFile.good()) {
+    // calculate line by line
+    line++;
+    std::getline(myFile,file_line);
+    if (verbose == true) {
```
martyngigg, line 126: Just `if(verbose)` is more standard for this type of check.

exercises/t_jubb/ex01_basics/src/find_words.cpp
```diff
@@ -0,0 +1,189 @@
...
+int ordered_by_value(std::vector< std::pair<  int , std::string > >& vec ,std::map<std::string, int >& count_map) {
+  /*
+    Sorts a map of <string,value> pairs by value (ascending), by inserting the <value,string> pairs into a vector
+    and then performing the default sort.
+
+    @param vec : a vector to append the entries of count_map to.
+    @param count_map : a map of <string,value> pairs.
+
+  */
+  std::pair<int,std::string> elem;
```
martyngigg, line 80: I think this would be a good use of [`std::transform`](http://en.cppreference.com/w/cpp/algorithm/transform) and a lambda:
```c++
  std::vector<std::pair<int, std::string>> ordered;
  ordered.resize(count_map.size()); // transform doesn't resize the output range
  std::transform(std::begin(count_map), std::end(count_map), std::begin(ordered), 
    [](const std::pair<std::string, int> & src_elem) { return std::make_pair(src_elem.second, src_elem.first); });
  std::sort(std::begin(ordered), std::end(ordered));
```

exercises/t_jubb/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,270 @@
+
```
martyngigg: This is the type of problem where we should start splitting code into separate header and implementation files. For instance we would have separate header files for `Shape` and all of the derived classes and also `ShapeSorter`. Usually C++ code is split between a header file  (`.h`) for declarations and an implementation file (`*.cpp`) - this helps maintain some sort of order as the code base grows.

exercises/t_jubb/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,270 @@
...
+#include <cmath>
+#include <iostream>
+#include <string>
+#include <vector>
+
+
+/*
+Shape class
+ */
+class Shape {
```
martyngigg: When deconstructing problems with O.O. it's good to think of the separation of an abstract "concept" versus a concrete realisation of that concept. For concept types you would generally define a so-called abstract base class, a class where all methods are pure virtual, that itself contains no behaviour but simply defines the interface that any realisation of that concept must adhere to.

For instance, `Shape` is a concept and not a specific realisation so I would expect `Shape` to be abstract and thus something you can't create. A realisation of this or concrete type as they are known would then be a `Square` that inherits from `Shape` and overrides the virtual methods to define its own implementation for each of the methods.

exercises/t_jubb/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,270 @@
...
+	std::string get_type() { return type; };
+	void print_shape() {  
+		int pad = 10 - type.size();
+		std::string pad_blank(pad, ' ');
+		std::cout << "Shape : " << type << pad_blank;
+		std::cout << " Area : " << area << " Perim. : " << perimeter << std::endl;
+		};
+
+protected:
+	double area, perimeter;
```
martyngigg: It's generally considered bad practice to mark member data as protected as it doesn't really encapsulate it properly. In this specific case, as discussed above, the class doesn't need an implementation, but if it did each class should ideally keep all of its data `private`.

exercises/t_jubb/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,270 @@
...
+	std::string type;
+};
+
+/*
+Square class
+*/
+class Square: public Shape {
+
+public:
+	Square() {
```
martyngigg: It is generally considered good practice to avoid having a default constructor unless it make sense for that class. In this case for example the default here produces an object in an odd state where it's a square but with zero area and perimeter. It would be better to just remove this constructor and require the height/width to be passed in.

exercises/t_jubb/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,270 @@
...
+Comparisons between shapes for area, perimeter
+*/
+bool compare_area(Shape& shape1, Shape& shape2) {
+	return shape1.get_area() < shape2.get_area();
+};
+bool compare_perimeter(Shape& shape1, Shape& shape2) {
+	return shape1.get_perimeter() < shape2.get_perimeter();
+};
+
+class ShapeSorter : public Shape {
```
martyngigg: Is `ShapeSorter` a `Shape`. Public inheritance should be thought of as an "is-a" relationship, i.e a circle is a shape in the case above but here a `ShapeSorter` is not a shape.

exercises/t_jubb/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,270 @@
...
+			(*i).print_shape();
+		};
+	};
+};
+
+/*
+Prints shapes in shape_vector with a given number of sides.
+*/
+void ShapeSorter::print_sides(int sides) {
+	for (auto i = shape_vector.begin(); i != shape_vector.end(); i++) {
```
martyngigg: This is a good place for using a range-based for loop:
```
for( const auto & iter : shape_vector )
```

exercises/t_jubb/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,270 @@
...
+/*
+
+Sort the vector of Shapes by their perimeter and print out the contents.
+
+@param ascending : Wether to sort from lowest to highst
+
+*/
+void ShapeSorter::print_by_perimeter(bool ascending) {
+
+	sort(shape_vector.begin(), shape_vector.end(), compare_perimeter);
```
martyngigg: Same comment as above.

exercises/t_jubb/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+	if (ascending) {
+		for (auto i = shape_vector.begin(); i != shape_vector.end(); i++) {
```
martyngigg: I would consider having a private member function to do the printing that can be called by the others.

exercises/t_jubb/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+	Square sq(1.0);
+	std::cout << "Area      : " << sq.get_area() << std::endl;
+	std::cout << "Perimeter : " << sq.get_perimeter() << std::endl;
+	std::cout << "Sides     : " << sq.get_sides() << std::endl;
+
+	Rectangle re(1.0, 2.0);
+	IsoscelesTriangle tri(5.7, 2.3);
+
+	std::vector<Shape> vec;
```
martyngigg: The way your code is currently structured this does work but if you change it to the abstract implementation discussed above then this won't work as expected due to something called [object slicing](https://en.wikipedia.org/wiki/Object_slicing). This is related to [polymorphism](http://www.cplusplus.com/doc/tutorial/polymorphism/) and how you can store vectors of things that are related by a similar concept; in this case Shape.

PR #33, opened 2018-06-06
PR for C++ training exercises

BhuvanBezawada:
**Report to:** @LamarMoore 

**PR Details:**
Finished both the training exercises and submitting the PR as part of training.

### DIFF
--- a/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
+++ b/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
@@ -122,10 +122,14 @@ int main(int, char **) {
     // Flip the output
     multimap<int, string> dst = flip_map(word_and_count);
 
-    // Print out the word and its count
+    // Print out the word and its count (to terminal and file)
+    ofstream out_file;
+    out_file.open("results.txt");
     for(multimap <int, string>::const_iterator it = dst.end(); it != dst.begin(); --it) {
         cout << it->second << "\t\t" << it->first << endl;
+        out_file << it->second << "\t\t" << it->first << endl;
     }
+    out_file.close();
 
     return 0;
 }
--- a/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
+++ b/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
@@ -1,0 +1,109 @@
+/**
+ * PROGRAM DESCRIPTION:
+ * 
+ * A commnad line program to read in a text file and give the occurrences of the words. 
+ *
+**/
+
+// Imports
+#include <iostream>
+#include <string>
+#include <map>
+#include <vector>
+#include <fstream>
+
+// Namespace being used
+using namespace std;
+
+
+/**
+ *  Split function to take a line of input and return words in line.
+ *  From Accelerated C++ Textbook, page 88.
+**/
+vector<string> split(const string& s) {
+    // Create return object
+    vector<string> ret;
+    typedef string::size_type string_size;
+    string_size i = 0;
+
+    // Process characters
+    while(i != s.size()) {
+        // Ignore the leading blanks
+        while(i != s.size() && isspace(s[i])) {
+            ++i;
+        }
+
+        // Find end of next word
+        string_size j = i;
+        while(j != s.size() && !isspace(s[j])) {
+            ++j;
+        }
+        
+        // Found some non-whitespace characters
+        if(i != j) {
+            ret.push_back(s.substr(i, j-i));
+            i = j;
+        }
+    }
+
+    return ret;
+}
+
+
+/**
+ * Main function.
+**/
+int main(int, char **) {
+
+    // Get file name from user and store
+    cout << "Please enter the name of the file: " << endl;
+    string file_name;
+    cin >> file_name;
+
+    // Load the file
+    ifstream input_file(file_name);
+
+    // To store line of text and words and their counts
+    string line;
+    map<string, int> word_and_count;
+
+    // Loop through each line
+    while(getline(input_file, line)) {
+        // Get the list of words in each line
+        vector<string> words = split(line);
+        
+        // Process each word
+        for (vector<string>::const_iterator it = words.begin(); it != words.end(); it++) {
+            // Dereference the iterator to get back the word
+            string word = *it;
+
+            // Loop through the word
+            for(int i = 0; i < word.size(); i++) {
+                // Deal with hyphens?
+                // Deal with allowed punctuation?
+                
+                // Remove all punctuation and white space
+                // Convert upper case letters to lowercase letters
+                if(ispunct(word[i]) || isspace(word[i])) {
+                    word.erase(i--, 1);
+                } else if(isupper(word[i])) {
+                    word[i] = tolower(word[i]);
+                }
+            }
+
+            // Only count words that are longer than 4 letters
+            if(word.length() > 4) {
+                ++word_and_count[word];
+            }
+        }
+    }
+
+    // Print out the word and it's count
+    for(map <string, int>::const_iterator it = word_and_count.begin(); it != word_and_count.end(); ++it) {
+        cout << it->first << "\t" << it->second << endl;
+    }
+
+    return 0;
+}
+
+
--- a/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
+++ b/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
@@ -72,14 +72,24 @@ vector<string> split(const string& s) {
 /**
  * Main function.
 **/
-int main(int, char **) {
+int main() {
     // Get file name from user and store
     cout << "Please enter the name of the file: " << endl;
     string file_name;
     cin >> file_name;
+	ifstream input_file(file_name);
+	
+	// Try to load the file
+	if(!input_file) {
+		cerr << "Error, there was a problem opening/reading the file." << endl;
+		exit(EXIT_FAILURE);
+	}
 
-    // Load the file
-    ifstream input_file(file_name);
+	// Check if file is empty
+	// Adapted from https://stackoverflow.com/questions/2390912/checking-for-an-empty-file-in-c
+	if(input_file.peek() == std::ifstream::traits_type::eof()) {
+		cerr << "Error, the file was empty!" << endl;
+	}
 
     // To store line of text and words and their counts
     string line;
@@ -87,9 +97,10 @@ int main(int, char **) {
 
     // Loop through each line
     while(getline(input_file, line)) {
+
         // Get the list of words in each line
         vector<string> words = split(line);
-        
+
         // Process each word
         for (vector<string>::const_iterator it = words.begin(); it != words.end(); it++) {
             // Dereference the iterator to get back the word
@@ -102,7 +113,7 @@ int main(int, char **) {
                 
                 // Remove all punctuation and white space
                 // Convert upper case letters to lowercase letters
-                if(ispunct(word[i]) || isspace(word[i])) {
+                if(ispunct(word[i])) {
                     word.erase(i--, 1);
                 } else if(isupper(word[i])) {
                     word[i] = tolower(word[i]);
@@ -115,6 +126,8 @@ int main(int, char **) {
             }
         }
     }
+
+	cout << "END..." << endl;
 
     // Header for output
     cout << "Word" << "\t\t" << "Usage" << endl;
--- a/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
+++ b/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
@@ -1,7 +1,7 @@
 /**
  * PROGRAM DESCRIPTION:
  * 
- * A commnad line program to read in a text file and give the occurrences of the words. 
+ * A command line program to read in a text file and give the occurrences of the words. 
  *
 **/
 
@@ -11,9 +11,27 @@
 #include <map>
 #include <vector>
 #include <fstream>
+#include <algorithm>
 
 // Namespace being used
 using namespace std;
+
+
+/**
+ * Code to flip the <string, int> pairs.
+ * Code adapted from: https://stackoverflow.com/questions/5056645/sorting-stdmap-using-value?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa
+**/
+template<typename A, typename B>
+pair<B, A> flip_pair(const pair<A, B> &p) {
+    return pair<B, A>(p.second, p.first);
+}
+
+template<typename A, typename B>
+multimap<B, A> flip_map(const map<A, B> &src) {
+    multimap<B, A> dst;
+    transform(src.begin(), src.end(), inserter(dst, dst.begin()), flip_pair<A, B>);
+    return dst;
+}
 
 
 /**
@@ -54,7 +72,6 @@ vector<string> split(const string& s) {
  * Main function.
 **/
 int main(int, char **) {
-
     // Get file name from user and store
     cout << "Please enter the name of the file: " << endl;
     string file_name;
@@ -98,12 +115,18 @@ int main(int, char **) {
         }
     }
 
-    // Print out the word and it's count
-    for(map <string, int>::const_iterator it = word_and_count.begin(); it != word_and_count.end(); ++it) {
-        cout << it->first << "\t" << it->second << endl;
+    // Header for output
+    cout << "Word" << "\t\t" << "Usage" << endl;
+    cout << "" << endl;
+
+    // Flip the output
+    multimap<int, string> dst = flip_map(word_and_count);
+
+    // Print out the word and its count
+    for(multimap <int, string>::const_iterator it = dst.end(); it != dst.begin(); --it) {
+        cout << it->second << "\t\t" << it->first << endl;
     }
 
     return 0;
 }
-
 
--- a/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
+++ b/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
@@ -12,6 +12,7 @@
 #include <vector>
 #include <fstream>
 #include <algorithm>
+#include <iterator>
 
 // Namespace being used
 using namespace std;
@@ -125,7 +126,7 @@ int main(int, char **) {
     // Print out the word and its count (to terminal and file)
     ofstream out_file;
     out_file.open("results.txt");
-    for(multimap <int, string>::const_iterator it = dst.end(); it != dst.begin(); --it) {
+    for(multimap <int, string>::const_reverse_iterator it = dst.rbegin(); it != dst.rend(); ++it) {
         cout << it->second << "\t\t" << it->first << endl;
         out_file << it->second << "\t\t" << it->first << endl;
     }
--- a/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
+++ b/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
@@ -2,7 +2,7 @@
  * PROGRAM DESCRIPTION:
  *
  * A command line program to read in a text file and give the occurrences of the
- *words.
+ * words.
  *
  **/
 
--- a/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
+++ b/exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
@@ -1,150 +1,151 @@
 /**
  * PROGRAM DESCRIPTION:
- * 
- * A command line program to read in a text file and give the occurrences of the words. 
  *
-**/
+ * A command line program to read in a text file and give the occurrences of the
+ *words.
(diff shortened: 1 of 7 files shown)


LamarMoore:
Doesn't build on Windows. Can you fix this error?

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+
+/**
+ * Main function.
+**/
+int main(int, char **) {
+    // Get file name from user and store
+    cout << "Please enter the name of the file: " << endl;
+    string file_name;
+    cin >> file_name;
```
LamarMoore: There isn't any error handling around this. Would be nice if the user is notified that they have entered an invalid filename or event an empty file.

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    return ret;
+}
+
+
+/**
+ * Main function.
+**/
+int main(int, char **) {
```
LamarMoore: I guess strictly speaking you would want to parse command-line arguments. For example what is shown here [Parsing Command-Line Arguments](http://www.cplusplus.com/articles/DEN36Up4/)

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,136 @@
...
+            string word = *it;
+
+            // Loop through the word
+            for(int i = 0; i < word.size(); i++) {
+                // Deal with hyphens?
+                // Deal with allowed punctuation?
+                
+                // Remove all punctuation and white space
+                // Convert upper case letters to lowercase letters
+                if(ispunct(word[i]) || isspace(word[i])) {
```
LamarMoore: Why do you need to check for spaces here? Aren't they all removed by `split()`?

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,136 @@
...
+                    word.erase(i--, 1);
```
LamarMoore: In general it may not be a good idea to change the size of an array whilst looping over it. Also, hyphenated words are treated as separate words, so do you need to remove the hyphen? What does `ispunct()` return for a hyphen? Also, shouldn't you create a two new words from words separated by punctuation?

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,136 @@
...
+                } else if(isupper(word[i])) {
+                    word[i] = tolower(word[i]);
+                }
+            }
+
+            // Only count words that are longer than 4 letters
+            if(word.length() > 4) {
+                ++word_and_count[word];
```
LamarMoore: 👍

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,136 @@
...
+        // Get the list of words in each line
+        vector<string> words = split(line);
+        
+        // Process each word
+        for (vector<string>::const_iterator it = words.begin(); it != words.end(); it++) {
+            // Dereference the iterator to get back the word
+            string word = *it;
+
+            // Loop through the word
+            for(int i = 0; i < word.size(); i++) {
```
LamarMoore: You might be able to avoid this loop since it is most likely punctuation will appear at the end of the word.

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,136 @@
...
+    multimap<int, string> dst = flip_map(word_and_count);
+
+    // Print out the word and its count (to terminal and file)
+    ofstream out_file;
+    out_file.open("results.txt");
+    for(multimap <int, string>::const_iterator it = dst.end(); it != dst.begin(); --it) {
+        cout << it->second << "\t\t" << it->first << endl;
+        out_file << it->second << "\t\t" << it->first << endl;
+    }
+    out_file.close();
```
LamarMoore: A message to the user that you have written an output file and its location would be friendlier.

LamarMoore (changes requested):

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,355 @@
...
+#include <algorithm>
+
+// Namespace to use
+using namespace std;
+
+
+/**
+ *  Base class
+**/
+class Shape {
```
LamarMoore: Normally, we like to separate the implementation from the definition. e.g `Shape.h, Shape.cpp` and then we include the  header files we need to use. You do not need to change this now but bear this in mind moving forward.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,355 @@
...
+
+    // Instance variables
+    public:
+        string shape_type;
```
LamarMoore: Do these need to be public and modifiable to the outside world? Consider making these public and readable via getter methods.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,355 @@
...
+        int num_sides;
```
LamarMoore: Can `num_sides` be < 0? Maybe `unsigned int` or even `unsigned __int16`

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,355 @@
...
+
+    // Methods
+    public:
```
LamarMoore: classes with pure virtual methods which are intended to act as base classes should have a virtual destructor.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,355 @@
...
+    // Own side length
+    private:
+        double side_length;
+
+    // Method implementations
+    public:
+        // Constructor 
+        Square(int side1) {
+            side_length = side1;
+            shape_type = "Square";
```
LamarMoore: This could be part of the `Shape` constructor. E.g `Shape(const std::string &type, int numsides)` which then sets the shape and you could do something like `Square(int side) : Shape("square", 4)`. Have you considered using an `enum class` for the type or some other type instead of a `std::string` for storing the type?

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,355 @@
...
+            Shape::print_type(shape_type);
+        }
+
+        // Prints out number of sides
+        void print_num_sides() const {
+            Shape::print_num_sides(num_sides);
+        }
+
+        // Calculates and returns the perimeter
+        double perimeter() {
```
LamarMoore: add the `override` keyword to overidden methods e.e `double perimeter() override {}`

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,355 @@
...
+class Square : public Shape {
+
+    // Own side length
+    private:
+        double side_length;
+
+    // Method implementations
+    public:
+        // Constructor 
+        Square(int side1) {
```
LamarMoore: Why capture as `int` and store as `double`? Why not capture as `double` in the first place. Also `side1` is a bad name. What about `width` or `sideLength`? You may think of something better.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,355 @@
...
+            side_length = side1;
+            shape_type = "Square";
+            num_sides = 4;
+        }
+
+        // Let's user know what type it is
+        void print_type() const {
```
LamarMoore: `print_type()` and `print_num_sides()` do not need to be redefined if they exist in the base class. These methods are essentially duplicated. Also, are these required based on the requirements in the question?

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,355 @@
...
+    triangle.print_type();
+    triangle.print_num_sides();
+    triangle.print_side_lengths();
+    cout << "Area: " << triangle.area() << endl;
+    cout << "Perimeter: " << triangle.perimeter() << endl;
+    cout << endl;
+
+
+    // Add shapes into a vector
+    vector<Shape*> shapes;
```
LamarMoore: Instead of doing this, you could have created a vector of shapes and just taken a reference to the vector in all of your `ShapeSorter` methods. Instead you are copying the pointer vector everytime you call a method, then copying again after you enter.

LamarMoore (changes requested):
Hi a number of things here I highlighted which could improve your inheritance structure and design in general.

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * Main function.
+**/
+int main() {
+    // Get file name from user and store
+    cout << "Please enter the name of the file: " << endl;
+    string file_name;
+    cin >> file_name;
+	ifstream input_file(file_name);
```
LamarMoore: indentation (and the two if blocks below this)

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+	// Try to load the file
+	if(!input_file) {
+		cerr << "Error, there was a problem opening/reading the file." << endl;
+		exit(EXIT_FAILURE);
+	}
+
+	// Check if file is empty
+	// Adapted from https://stackoverflow.com/questions/2390912/checking-for-an-empty-file-in-c
+	if(input_file.peek() == std::ifstream::traits_type::eof()) {
+		cerr << "Error, the file was empty!" << endl;
```
LamarMoore: you should return/exit here too. Maybe just return non zero

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+    // Get file name from user and store
+    cout << "Please enter the name of the file: " << endl;
+    string file_name;
+    cin >> file_name;
+	ifstream input_file(file_name);
+	
+	// Try to load the file
+	if(!input_file) {
+		cerr << "Error, there was a problem opening/reading the file." << endl;
+		exit(EXIT_FAILURE);
```
LamarMoore: return non zero?

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+            }
+
+            // Only count words that are longer than 4 letters
+            if(word.length() > 4) {
+                ++word_and_count[word];
+            }
+        }
+    }
+
+	cout << "END..." << endl;
```
LamarMoore: indentation

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+            string word = *it;
+
+            // Loop through the word
+            for(int i = 0; i < word.size(); i++) {
+                // Deal with hyphens?
+                // Deal with allowed punctuation?
+                
+                // Remove all punctuation and white space
+                // Convert upper case letters to lowercase letters
+                if(ispunct(word[i])) {
```
LamarMoore: Still not clear why you are deleting the hyphen. Are you deleting the hyphen?

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,298 @@
...
+class Shape {
+
+    // Instance variables
+    private:
+        string shape_type;
+        unsigned int num_sides;
+
+    // Methods
+    public:
+		Shape(const string &type, int sides) {
```
LamarMoore: indentation. Consider download the LLVM clang-format plugin for visual studio.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,298 @@
...
+        void print_type() const {
+            cout << "Shape type is: " << get_shape_type() << endl;
+        }
+
+        virtual void print_num_sides() const {
+            cout << "Number of sides is: " << get_num_sides() << endl;
+        }
+
+		// Getters
+		string get_shape_type() const {
```
LamarMoore: indentation. Review all your code and make sure your indentation is correct.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+    // Own side lengths
+    private:
+        double side1_len;
+        double side2_len;
+
+    // Method implementations
+    public:
+        // Constructor 
+        Rectangle(double side1, double side2) : Shape("Rectangle", 4) {
```
LamarMoore: width, height as meaningful names ?

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,298 @@
...
+            return (2*M_PI*radius_size);
+        }
+
+        // Calculates and returns the area
+        double area() override {
+            return (M_PI*radius_size*radius_size);
+        }
+
+        // Prints out the length of its radius
+        void print_radius_length() {
```
LamarMoore: Still not sure why these print methods are necessary.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,298 @@
...
+            vector<Shape*> sorted_areas = shapes;
+            sort(sorted_areas.begin(), sorted_areas.end(), ShapeSorter::compare_area);
+
+            for(vector<Shape*>::iterator it = sorted_areas.begin(); it != sorted_areas.end(); ++it) {
+                cout << (*it)->get_shape_type() << "\t\t" << (*it)->area() << endl;
+            }
+        }
+
+        // Print the shapes by perimeter descending
+        void print_per_des(vector<Shape*> shapes) {
```
LamarMoore: spell out what the method is doing in the name.

LamarMoore (commented):
Mostly minor comments surrounding indentation etc. When naming methods try to be explicit so that someone reading the interface will understand what the method is actually doing. Also, the naming styles which should be used can be found here [Coding Standards](http://developer.mantidproject.org/Standards/index.html). All developers abide by these standards so that we can have consistency in the codebase and we all know what to expect when reading each others code.

LamarMoore:
Once the RHEL build is working I will approve these changes.

LamarMoore (approved):

exercises/bhuvan_bezawada/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+  multimap<B, A> dst;
+  transform(src.begin(), src.end(), inserter(dst, dst.begin()), flipPair<A, B>);
+  return dst;
+}
+
+/**
+ *  Split function to take a line of input and return words in line.
+ *  From Accelerated C++ Textbook, page 88.
+ **/
+vector<string> split(const string &s) {
```
OwenArnold, line 40: This is acceptable for the purpose of the exercises, but you should be aware of `regular expressions` if you are not already as finding words would be much easier using them.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+    shapeType = type;
+    numSides = sides;
+  }
+
+  // Methods to be overridden
+  virtual double perimeter() = 0;
+  virtual double area() = 0;
+
+  // Methods for info
+  void printType() const {
```
OwenArnold, line 42: I would not hard-code the IO like this, I would return a string and leave it up to the client code to determine how to best present/persist it.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+    cout << "Shape type is: " << getShapeType() << endl;
+  }
+
+  virtual void printNumSides() const {
+    cout << "Number of sides is: " << getNumSides() << endl;
```
OwenArnold, line 47: Same as above

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+  int getNumSides() const { return numSides; }
+
+  // Destructor
+  virtual ~Shape() {}
+};
+
+/**
+ *  Square class
+ **/
+class Square : public Shape {
```
OwenArnold, line 62: Isn't a square a special case of a rectangle? Did you consider having Square inherit from Rectangle instead?

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+// Methods
+public:
+  Shape(const string &type, int sides) {
+    shapeType = type;
+    numSides = sides;
+  }
+
+  // Methods to be overridden
+  virtual double perimeter() = 0;
```
OwenArnold, line 38: `perimeter` and `area` should be `const`. They should guarantee not making changes to internal state.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+      }
+    }
+  }
+
+  // Method to print shapes of chosen number of sides
+  void printShapesMatchingNumSides(vector<Shape *> shapes, int numSides) {
+    for (vector<Shape *>::const_iterator it = shapes.begin();
+         it != shapes.end(); ++it) {
+      if (((*it)->getNumSides()) == numSides) {
+        cout << (*it)->getShapeType() << endl;
```
OwenArnold, line 187: As a word of warning, careful with `std::endl` in for loops. It involves an expensive flush.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+      }
+    }
+  }
+
+  // Print the shapes by area descending
+  void printAreaDescending(vector<Shape *> shapes) {
```
OwenArnold, line 193: Why pass `shapes` by value instead of by const ref?

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+    vector<Shape *> sortedAreas = shapes;
+    sort(sortedAreas.begin(), sortedAreas.end(), ShapeSorter::compareArea);
```
OwenArnold, line 195: passing the function predicate as a named function is fine, you could also have done this with a lambda, though I prefer your approach for readability.

exercises/bhuvan_bezawada/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+  Rectangle rectangle(5, 10);
+  Circle circle(5);
+  Triangle triangle(10, 5);
+
+  // Add shapes into a vector
+  vector<Shape *> shapes;
+  shapes.push_back(&square);
+  shapes.push_back(&rectangle);
+  shapes.push_back(&circle);
+  shapes.push_back(&triangle);
```
OwenArnold, line 231: Interesting, you have no memory leak here so technically what you have done is fine, but the vector `shapes` does not own the shapes, and that isn't terribly clear. Maybe these should be stored as const pointers in the vector? Otherwise you could have stored them as `std::unique_ptr`

OwenArnold (commented):
Looks pretty good to me. I've made some comments for you to think about.

PR #34, opened 2018-07-03
Samuel jones exercises

Pasarus:
First exercise completed.


exercises/Samuel_Jones/README.md
```diff
@@ -0,0 +1,3 @@
+This directory contains the template CMake projects for each of the exercises.
```
ElliotAOram: Nit: This should be updated from the template version

exercises/Samuel_Jones/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+bool sortingWords(const pair<string, int> &, const pair<string, int> &);
+
+/**
+ * Skeleton main routine
+ */
+
+int main(int, char **)
+{
+    //Get the file name from the user via command line
+    string fileName = getFileName();
```
ElliotAOram: can use const auto here: `const auto fileName`

exercises/Samuel_Jones/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    //Load the ascii file
+    vector<string> fileStrings;
+    vector<pair<string, int>> uniqueWords;
+
+    ifstream fs;
+    fs.open(fileName.c_str());
+    //Read the file words and get each unique word in
+    readWords(&fileStrings, fs);
```
ElliotAOram, line 35: Good pass by reference for large vector

exercises/Samuel_Jones/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+//Get the file name a)nd check whether or not the file exists etc.
+string getFileName()
+{
+    string fileName;
+    do
+    {
+        fileName = "";
+        cout << "Give the  file name" << endl;
+        cin >> fileName;
+    } while (!doesFileExist(fileName));
```
ElliotAOram, line 66: Good validation

exercises/Samuel_Jones/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+    {
+        cout << "File could not be found!" << endl;
+    }
+    return infile.good();
+}
+
+void readWords(vector<string> *words, ifstream &fs)
+{
+    //Create the stream buffer for content into a string.
+    string contents((istreambuf_iterator<char>(fs)), istreambuf_iterator<char>());
```
ElliotAOram: const auto contents

exercises/Samuel_Jones/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    //Create the word utilising the buffer iterators
+    string tempWord;
+    for (auto i = 0u; i < contents.length(); i++)
```
ElliotAOram: recalculating length each time here (`contents.length()` called for every loop iteration) could be improved by calculating this outside the loop:
`const auto maxCharacters = contents.length()`
Then inside the loop refer to `maxCharacters` as the stopping condition

ElliotAOram: C++ has ways of putting file content into a vector of words. Is there a reason you choose character by character instead of words? Both methods are perfectly valid just interested in your choice.

exercises/Samuel_Jones/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,204 @@
...
+            tempWord = "";
+        }
+    }
+    //Catch the final word and add it to the list as it will be forgotten.
+    words->push_back(tempWord);
+}
+
+//If the charecter is one of the delimeters return false
+bool notDelim(char c)
+{
```
ElliotAOram, line 114: The alternative to this is the following (where `c` is a single charterer):
` if(c >= 'a'&& c <= 'z')`
This will work and save you typing out the list below however you would have to be a bit careful of `\n` or similar.

exercises/Samuel_Jones/ex01_basics/src/main.cpp
Pasarus: I believe the vector of words wouldn't necessarily cut out all the things I wanted to remove so it's more efficient to just do it this way. If you check out that god awful string statement I wrote for the delimiters, it has them all there, perhaps I missed the option to add my own delimiters to it.

exercises/Samuel_Jones/ex01_basics/src/main.cpp
Pasarus: You make a fair point on this but I don't see a reason to change to get a little bit of performance out it when I would then just have to do further checks for the the end characters and null terminators.

exercises/Samuel_Jones/ex01_basics/src/main.cpp
Pasarus: Will fix that now.

exercises/Samuel_Jones/ex01_basics/src/main.cpp
Pasarus: I can't auto contents because I utilize the string specific methods/functions later on

exercises/Samuel_Jones/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+    //Sort the lists based on the number of times it is present
+    sort(uniqueWords.begin(), uniqueWords.end(), sortingWords);
+
+    //print out to file named results, could request filename using getFileName() but not required.
+    ofstream results;
+    results.open("results.txt");
+    results << "Word\t\tUsage" << endl;
+    results << endl;
+    for (auto i = 0u; i < uniqueWords.size(); i++)
```
ElliotAOram: Can do similar here with with `const auto totalWords = uniqueWords.size()`

exercises/Samuel_Jones/ex01_basics/src/main.cpp
edwardb96: Performance is not the only metric by which we measure software quality. Especially in the majority of mantid. 
```

Comment (edwardb96) [reply exercises/Samuel_Jones/ex01_basics/src/main.cpp:114]: Performance is not the only metric by which we measure software quality, especially in the majority of mantid.
```
static auto const punctuation = std::string(".,?'\\\"!():\n\r");
std::find(punctuation.cbegin(), punctuation.cend(), c) == punctuation.cend());
```
Is considerably less code, more readable and more maintainable than the original case statement.

File: exercises/Samuel_Jones/ex01_basics/src/main.cpp (hunk 5 above)
Comment (edwardb96) [reply exercises/Samuel_Jones/ex01_basics/src/main.cpp]: `auto` does not limit your ability to call methods on the variable.

File: exercises/Samuel_Jones/ex01_basics/src/main.cpp [hunk 9]
```diff
@@ -0,0 +1,205 @@
...
+void findOccurencesofWords(vector<string> words, vector<pair<string, int>> *uniqueWords)
+{
+    string currentWord = "";
+    string lastWord;
+    bool inSet = false;
+    int currentTally = 1;
+    //Make sure that the vector is sorted
+    sort(words.begin(), words.end());
+
+    for (auto i = 0u; i < words.size(); i++)
```
ElliotAOram: `const auto maxWords = words.size()`

ElliotAOram:
@Pasarus good first exercise Sam. One thing you'll come across in Mantid is that we use [Doxygen style](http://www.stack.nl/~dimitri/doxygen/) commenting for our C++ code - just something to be aware of as you'll see it everywhere in the mantid code base.

PR #35, opened 2018-07-04
Sam jenkins exercises

LowriJenkins:
It says " treat hypenated words as a different word" wasn't sure if it meant that word1-word2 means word1 and word2 are different words, or that word1-word2 is different from word1 and word2, I have currently implemented the former.


exercises/sam_jenkins/README.md
```diff
@@ -0,0 +1,3 @@
+This directory contains the template CMake projects for each of the exercises.
```
ElliotAOram: Nit:  This should be updated as it is no longer a template file

exercises/sam_jenkins/ex01_basics/src/stringManip.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+bool not_space(const char& c)
+{
+	return !(isspace(c) || c == '-');
+}
+
+void strip_string(string& input)
+{
+	//remove these delimiters from the begining and ending of words
+	string delimiters = " .,?:;\'\"()!";
```
ElliotAOram: This can be `const string delimiters`

exercises/sam_jenkins/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,43 @@
...
+		//if the stream opened properly, read the entire file to a string, then make the string all lowercase.
+		string read((istreambuf_iterator<char>(in)), (istreambuf_iterator<char>()));
+		transform(read.begin(), read.end(), read.begin(), ::tolower);
+
+		//create an unordered map to count words
+		unordered_map<string, int> count =countof(read);
+		//convert it to a sorted vector of pairs
+		vector<std::pair<string, int>> output = mapToVector(count);
+		//output to the file Out.txt
+		ofstream outputfile("Out.txt");
```
ElliotAOram: Nit: standard naming convention would probably be `outputFile` rather than `outputfile`.

exercises/sam_jenkins/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,43 @@
...
+		outputfile << std::setw(15) << "word" << std::setw(9) << "usage" << endl << endl;
+		for (vector<std::pair<string, int>>::const_iterator it = output.begin();
+			it != output.end(); ++it) {
+			outputfile <<std::setw(15) << it->first <<std::setw(6) << it->second << endl;
+		}
```
ElliotAOram: You should make sure that you close the `ofstream` using `outputFile.close()`

exercises/sam_jenkins/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,43 @@
...
+	ifstream in(argv[1]);
+	if (in) {
+		//if the stream opened properly, read the entire file to a string, then make the string all lowercase.
+		string read((istreambuf_iterator<char>(in)), (istreambuf_iterator<char>()));
+		transform(read.begin(), read.end(), read.begin(), ::tolower);
+
+		//create an unordered map to count words
+		unordered_map<string, int> count =countof(read);
+		//convert it to a sorted vector of pairs
+		vector<std::pair<string, int>> output = mapToVector(count);
```
ElliotAOram: here `output` can be `const`:
`const vector<std::pair<string, int>> output`

exercises/sam_jenkins/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,43 @@
...
+	}
+	//open the input stream
+	ifstream in(argv[1]);
+	if (in) {
+		//if the stream opened properly, read the entire file to a string, then make the string all lowercase.
+		string read((istreambuf_iterator<char>(in)), (istreambuf_iterator<char>()));
+		transform(read.begin(), read.end(), read.begin(), ::tolower);
+
+		//create an unordered map to count words
+		unordered_map<string, int> count =countof(read);
```
ElliotAOram: Here `count` can be const `const`:
`const unordered_map<string, int> count`

exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp
```diff
@@ -0,0 +1,49 @@
...
+unordered_map<string, int> countof(const string& input)
+{
+	//create a unordered map and string input stream
+	unordered_map<string, int>wordtable;
+	std::istringstream iss(input);
+	//create an iterator to loop through the string
+	auto i = input.begin();
+	while (i != input.end()) {
+		//find the first non space character
+		i = find_if(i, input.end(), not_space);
```
ElliotAOram: I believe you can use `find_if_not(i, input.end(), space);` here instead of `find_if` with the `not_space`. This will save you having to define `not_space`.

exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+
+
+using std::string; using std::unordered_map; using std::vector; using std::pair;
+
+unordered_map<string, int> countof(const string& input)
+{
+	//create a unordered map and string input stream
+	unordered_map<string, int>wordtable;
+	std::istringstream iss(input);
```
ElliotAOram: You initialise `iss` here but then don't use it. Instead you are using `input`. This is absolutely fine but you shouldn't need `iss`

ElliotAOram (changes requested):
This is a good implementation for exercise 1, well done! Good to see that you have separated the solution into separate files and nice use of STL functions for sorting/searching. I've commented on some areas of improvement.

exercises/sam_jenkins/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,44 @@
+#include <iostream>
+#include <fstream>
+#include <unordered_map>
+#include <iomanip>
+#include <vector>
+#include "mapFunctions.h"
+
+
+using std::ifstream; using std::string; using std::cout; using std::endl; using std::istreambuf_iterator; 
+using std::unordered_map; using std::vector; using std::ofstream;
```
samueljackson92: `using namespace std;` is simpler and clearer. As I mentioned to Andre, typically we just prefix standard library functions with `std::` to make it clear that is where they're from, but for this small project this is ok.

Also please use one line per using statement.

exercises/sam_jenkins/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+int main(int argc, char **argv) {
+	//check for valid number of arguements
+	if (argc !=2) {
+		cout << "This Program requires a single input" << endl;
+			return 1;
+	}
+	//open the input stream
+	ifstream in(argv[1]);
+	if (in) {
```
samueljackson92: Instead of having most of your code in an if statement, use an [early exit](https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code) instead:

```cpp
if (!in) {
    cout << "could not find file: " << argv[1] << endl;
    return 1;
}

...

File: exercises/sam_jenkins/ex01_basics/src/main.cpp [hunk 11]
```diff
@@ -0,0 +1,44 @@
...
+		//if the stream opened properly, read the entire file to a string, then make the string all lowercase.
+		string read((istreambuf_iterator<char>(in)), (istreambuf_iterator<char>()));
+		transform(read.begin(), read.end(), read.begin(), ::tolower);
+
+		//create an unordered map to count words
+		const unordered_map<string, int> count =countof(read);
```
samueljackson92: Use `auto` where you can. They type of count can be deduced from the function call. Same below.

exercises/sam_jenkins/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,44 @@
...
+		//convert it to a sorted vector of pairs
+		const vector<std::pair<string, int>> output = mapToVector(count);
+		//output to the file Out.txt
+		ofstream outputFile("Out.txt");
+		outputFile << std::setw(15) << "word" << std::setw(9) << "usage" << endl << endl;
+		for (vector<std::pair<string, int>>::const_iterator it = output.begin();
+			it != output.end(); ++it) {
+			outputFile <<std::setw(15) << it->first <<std::setw(6) << it->second << endl;
+		}
+		outputFile.close();
```
samueljackson92: At no point do you close the input file? This is an example of why [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) is so important.

exercises/sam_jenkins/ex01_basics/src/main.cpp
samueljackson92: 👍 that you're thinking of using `const` where you can.

exercises/sam_jenkins/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,44 @@
...
+		transform(read.begin(), read.end(), read.begin(), ::tolower);
+
+		//create an unordered map to count words
+		const unordered_map<string, int> count =countof(read);
+		//convert it to a sorted vector of pairs
+		const vector<std::pair<string, int>> output = mapToVector(count);
+		//output to the file Out.txt
+		ofstream outputFile("Out.txt");
+		outputFile << std::setw(15) << "word" << std::setw(9) << "usage" << endl << endl;
+		for (vector<std::pair<string, int>>::const_iterator it = output.begin();
```
samueljackson92: If you find yourself using a long type like this vector-pair in many places, consider using a [type alias](https://en.cppreference.com/w/cpp/language/type_alias) and giving it a shorter, more descriptive name.

exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp
```diff
@@ -0,0 +1,48 @@
+#include <unordered_map>
+#include <sstream>
+#include <vector>
+#include "stringManip.h"
+
+
+
+using std::string; using std::unordered_map; using std::vector; using std::pair;
+
+unordered_map<string, int> countof(const string& input)
```
samueljackson92: We generally use doxygen style comments to document functions:
```cpp
/**
* Summary of what the function does
*
* @param input what this means
* @return returns what it returns
*/
```

exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp
```diff
@@ -0,0 +1,48 @@
...
+	unordered_map<string, int>wordtable;
+	//create an iterator to loop through the string
+	auto i = input.begin();
+	while (i != input.end()) {
+		//find the first non space character
+		i = find_if_not(i, input.end(), space);
+		//find the next space from that point
+		auto j = find_if(i, input.end(), space);
+		//take the substring between i and j, strip it of punctuation and add it to the map
+		if (i != input.end()) {
```
samueljackson92: Again, consider and early exit here to reduce complexity by flipping the if statement condition and using `continue`.

exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+
+using std::string; using std::unordered_map; using std::vector; using std::pair;
+
+unordered_map<string, int> countof(const string& input)
+{
+	//create a unordered map and string input stream
+	unordered_map<string, int>wordtable;
+	//create an iterator to loop through the string
+	auto i = input.begin();
```
samueljackson92: Two problems with the name `i`:
1) It's not descriptive.
2) Worse, typically it is used as the name for an integer counter variable in a c style for loop. 

`iter` would have been a better name (suggesting it's an iterator). An even better name would have described what were iterating over. `fileIterator`? `lineIterator`?

exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp
```diff
@@ -0,0 +1,48 @@
...
+	while (i != input.end()) {
+		//find the first non space character
+		i = find_if_not(i, input.end(), space);
+		//find the next space from that point
+		auto j = find_if(i, input.end(), space);
```
samueljackson92: Same issue as above with `i` for `j`.

exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+unordered_map<string, int> countof(const string& input)
+{
+	//create a unordered map and string input stream
+	unordered_map<string, int>wordtable;
+	//create an iterator to loop through the string
+	auto i = input.begin();
+	while (i != input.end()) {
+		//find the first non space character
+		i = find_if_not(i, input.end(), space);
```
samueljackson92: 👍 excellent that you're thinking of using standard library functions.

exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp
```diff
@@ -0,0 +1,48 @@
...
+				++wordtable[current];
+			}
+			//move i to js location
+			i = j;
+		}
+	}
+	return wordtable;
+}
+
+bool compare(const pair<string, int>& left, const pair<string, int>& right)
```
samueljackson92: Compare is a very descriptive name. What are we comparing? What does true mean in this context?

exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp
```diff
@@ -0,0 +1,48 @@
...
+{
+	return left.second > right.second;
+}
+
+vector<std::pair<string, int>> mapToVector(unordered_map<string, int>map)
+{
+	//iterate across the map, copying it into a vector of pairs
+	vector<std::pair<string, int>> sortingVector(map.begin(), map.end());
+	//sort the vector via the second value descending
+	std::sort(sortingVector.begin(), sortingVector.end(), compare);
```
samueljackson92: Instead of defining the `compare` function above I would have just used a [lambda function](https://en.cppreference.com/w/cpp/language/lambda) here.

exercises/sam_jenkins/ex01_basics/src/mapFunctions.h
```diff
@@ -0,0 +1,9 @@
+#include <unordered_map>
+#include <sstream>
+#include <vector>
+#include "stringManip.h"
```
samueljackson92: In Mantid the include naming scheme is "#include StringManip.h".

samueljackson92: Also why not `StringManipulation`

exercises/sam_jenkins/ex01_basics/src/stringManip.cpp
```diff
@@ -0,0 +1,25 @@
+#include <string>
+#include <cctype>
+#include <regex>
+using std::string;
+
+
+bool space(const char& c)
+{
+	return (isspace(c) || c == '-');
```
samueljackson92: 👍 Good.

exercises/sam_jenkins/ex01_basics/src/stringManip.cpp
```diff
@@ -0,0 +1,25 @@
...
+}
+
+
+void strip_string(string& input)
+{
+	//remove these delimiters from the begining and ending of words
+	const string delimiters = " .,?:;\'\"()!";
```
samueljackson92: 👍  good.

exercises/sam_jenkins/ex01_basics/src/stringManip.cpp
```diff
@@ -0,0 +1,25 @@
...
+using std::string;
+
+
+bool space(const char& c)
+{
+	return (isspace(c) || c == '-');
+}
+
+
+void strip_string(string& input)
```
samueljackson92: This name is find, but in the project naming conventions are camel case: `stripString`

exercises/sam_jenkins/ex01_basics/src/mapFunctions.h
```diff
@@ -0,0 +1,9 @@
...
+
+
+std::unordered_map<std::string, int> countof(const std::string& input);
```
samueljackson92: Generally we should document these with a short doxygen summary comment:
```
/// Function to do foo
```

exercises/sam_jenkins/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,22 @@
+#include "Shapes.h"
+#include "Circle.h"
+#include <math.h>
+
+	Circle::Circle() :radius(1)
+	{
+		type = "Circle";
+		sides = 1;
```
samueljackson92: Why not initialise these variables like radius? Then you should be able to make them `const`

exercises/sam_jenkins/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,22 @@
...
+	}
+	Circle::Circle(double radius) :radius(radius)
+	{
+		type = "Circle";
+		sides = 1;
+	}
+	double Circle::calcPerimiter()const {
```
samueljackson92: Be careful of you use of whitespace here between the name and the `const`

exercises/sam_jenkins/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "Square.h"
+#include "Rectangle.h"
+#include "Circle.h"
+#include "Triangle.h"
+#include "ShapeSorter.h"
+using std::cout; using std::cin; using std::endl; using std::string; using std::vector;
+
+int main(int, char **)
+{
+	vector<Shapes*> shapes(10);
```
samueljackson92: I do not think we need to modify the shapes in the vector so they should be declared `vector<Shapes const *>`

exercises/sam_jenkins/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,41 @@
...
+	shapes[0] = new Square(5);
```
samueljackson92: Do not use `new` and `delete`. Either use a `unique_ptr` if you only need one reference or a `shared_ptr` if you need reference counting. Only use raw pointers for non-owning references

exercises/sam_jenkins/ex02_oo_basics/src/main.cpp
samueljackson92: Also, here it would probably be nicer to use braced initilisation:

```cpp
vector<Shapes const *> shapes = { ... };
```

exercises/sam_jenkins/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,59 @@
...
+bool comparePerimiter(const Shapes* s1, const Shapes *s2)
+{
+	return(s1->calcPerimiter() > s2->calcPerimiter());
+}
+
+void ShapeSorter::typePrint(string type, vector<Shapes*> shapes)
+{
+	for (vector<Shapes*>::size_type i = 0; i!= shapes.size(); ++i){
+		if (shapes[i]->getType() == type) {
+			cout << " Type: " << setw(9)<<shapes[i]->getType() << " Sides: " << setw(4) << shapes[i]->getSides() << " Area: " << setw(9) << shapes[i]->calcArea() << " Perimiter: " << setw(9) << shapes[i]->calcPerimiter() << endl;
```
samueljackson92: These lines are way, way to long. Just use multiple calls to cout on multiple lines to make it more readable.

exercises/sam_jenkins/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,18 @@
+#ifndef GUARD_Circle_h
```
samueljackson92: Typically the convention we use is just `NAMESPACE_CLASSNAME_H_`

exercises/sam_jenkins/ex01_basics/src/stringManip.h
```diff
@@ -0,0 +1,11 @@
+#ifndef GUARD_stringManip_h
+#define GUARD_stringManip_h
+#include <string>
+#include <cctype>
+#include <regex>
```
samueljackson92: Do you need to include `regex` and `cctype` here?

exercises/sam_jenkins/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,17 @@
+#ifndef GUARD_ShapeSorter_h
+#define GUARD_ShapeSorter_h
+#include "Shapes.h"
+#include <vector>
+class ShapeSorter {
+public:
+	static void typePrint(std::string type,std::vector<Shapes*> shapes);
```
samueljackson92: Avoid static unless its absolutely necessary. Just make then normal methods and create an instance.

samueljackson92: Please pass `std::string` and `std::vector` by reference (or `const` reference) to avoid copying.

exercises/sam_jenkins/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+	}
+	cout << endl;
+}
+
+void ShapeSorter::sidesPrint(int sides, vector<Shapes*> shapes) 
+{
+	for (vector<Shapes*>::size_type i = 0; i != shapes.size(); ++i) {
```
samueljackson92: Why not the far simpler:
```cpp
for (const auto shape : shapes) {
    shape->getSides();
    ....
}

**Review by samueljackson92 — CHANGES_REQUESTED** [review]
You said you wanted more detailed feedback on your code.

Be careful what you wish for. 😄

File: exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp (hunk 20 above)
Comment (LowriJenkins) [reply exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp]: I originally did, but some of how I used the lambda failed in older c++ versions (e.g. on rhel) I wasn't sure exactly what caused it, so I changed to not using the lambda at all

Comment (samueljackson92) [reply exercises/sam_jenkins/ex01_basics/src/mapFunctions.cpp]: You should be able to use a lambda. What was the error?

File: exercises/sam_jenkins/ex02_oo_basics/src/Circle.cpp (hunk 26 above)
Comment (LowriJenkins) [reply exercises/sam_jenkins/ex02_oo_basics/src/Circle.cpp]: I did originally but changed it due to running into problems with the linker

File: exercises/sam_jenkins/ex02_oo_basics/src/main.cpp [hunk 35]
```diff
@@ -19,16 +19,16 @@ int main(int, char **)
...
+	shapes.push_back(unique_ptr<Square>(new Square(5)));
+	shapes.push_back(unique_ptr<Circle>(new Circle(7)));
+	shapes.push_back(unique_ptr<Rectangle>(new Rectangle(10, 5)));
+	shapes.push_back(unique_ptr<Triangle>(new Triangle(10, 5)));
+	shapes.push_back(unique_ptr<Square>(new Square(9)));
+	shapes.push_back(unique_ptr<Circle>(new Circle(3)));
+	shapes.push_back(unique_ptr<Rectangle>(new Rectangle(2, 4)));
+	shapes.push_back(unique_ptr<Triangle>(new Triangle(5, 6)));
+	shapes.push_back(unique_ptr<Square>(new Square( 1)));
+	shapes.push_back(unique_ptr<Circle>(new Circle(2)));
```
samueljackson92, line 31: Fair enough, annoyingly looking at the RHEL build you're on GCC 4.8.5 which doesn't understand c++14. [4.9.0 does however...](https://godbolt.org/g/2akZC4)

samueljackson92: Note that this is worse though as the `new` could potentially throw and leak memory. Using `make_unique` avoids that possibility.

LowriJenkins: yeah I understand, but since make_unique isn't in c++ 11 I couldn't find another way to do it without implementing it myself

samueljackson92: And for the sake of this exercise I think that is acceptable.

exercises/sam_jenkins/ex02_oo_basics/src/Square.h
```diff
@@ -0,0 +1,17 @@
...
+#define GUARD_Square_h
+#include "Shapes.h"
+class Square : public Shapes {
+public:
+	Square();
+	Square(double side);
+	double calcPerimiter()const;
+	double calcArea()const;
+private:
+	double side;
```
samueljackson92: We don't need to side member variable here. We can just use `getSides` from the superclass.

exercises/sam_jenkins/ex02_oo_basics/src/Square.cpp
```diff
@@ -0,0 +1,16 @@
+#include "Shapes.h"
+#include "Square.h"
+	Square::Square() :Shapes("Square", 4), side(1)
+	{
+
+	}
+	Square::Square(double side) :Shapes("Square", 4), side(side)
```
samueljackson92: A better name would than `side` would be `length` or `edgeLength` so it's not confusing with `sides`

exercises/sam_jenkins/ex02_oo_basics/src/Square.cpp
```diff
@@ -0,0 +1,16 @@
...
+	{
+
+	}
+	double Square::calcPerimiter()const {
+		return 4 * side;
```
samueljackson92: `getSides() * side` would be better

exercises/sam_jenkins/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,24 @@
...
+#include <string>
+class Shapes {
+public:
+	
+	virtual double calcPerimiter() const =0;
+	virtual double calcArea() const =0;
+	//virtual std::string print() = 0;
+	std::string getType()const;
+	
+	int getSides()const;
```
samueljackson92: It might be an idea to `inline` this function so the compiler tries to optimise it.

exercises/sam_jenkins/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,24 @@
+//Base file for shapes to inherit from
+#ifndef GUARD_Shapes_h
+#define GUARD_Shapes_h
+#include <string>
+class Shapes {
+public:
+	
+	virtual double calcPerimiter() const =0;
+	virtual double calcArea() const =0;
+	//virtual std::string print() = 0;
```
samueljackson92: Don't leave commented out code hanging around.

samueljackson92 (changes requested):
Last few minor things

samueljackson92 (approved):
Looks good

samueljackson92 (changes requested):
Builds are failing though.

samueljackson92 (approved):

ElliotAOram (approved):

PR #36, opened 2018-07-10
Exercise 1 - Word Counter - Develop a routine for counting words in a file

robertapplin:
This PR adds a Word Counter program. 
The Word Counter program count's the usage of words, which are over 4 characters long, from a provided file and output's these words in order of most used to least used.

### DIFF
--- a/exercises/Robert_Applin/ex01_basics/src/main.cpp
+++ b/exercises/Robert_Applin/ex01_basics/src/main.cpp
@@ -88,7 +88,8 @@ void removeShorterThanFive(std::vector<std::string> &strings) {
  * @return A string transformed to lower case
  */
 std::string toLower(std::string str) {
-  std::transform(str.begin(), str.end(), str.begin(), std::tolower);
+  std::transform(str.begin(), str.end(), str.begin(),
+                 static_cast<int (*)(int)>(std::tolower));
   return str;
 }
 
--- a/exercises/Robert_Applin/ex01_basics/src/main.cpp
+++ b/exercises/Robert_Applin/ex01_basics/src/main.cpp
@@ -1,0 +1,202 @@
+/**
+ * Skeleton main routine
+ */
+
+#include <algorithm>
+#include <cctype>
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <iterator>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+using WordCount = std::vector<std::pair<std::string, std::size_t>>;
+
+/**
+ * @return An iterator pointing to the next letter contained
+ *         in the supplied set.
+ */
+std::string::const_iterator
+findNextCharacter(std::string::const_iterator fromIterator,
+                  std::string::const_iterator toIterator,
+                  const std::string &characters) {
+  for (auto it = fromIterator; it < toIterator; ++it) {
+    if (characters.find(*it) != std::string::npos)
+      return it;
+  }
+  return toIterator;
+}
+
+/**
+ * @return An iterator pointing to the next letter not
+ *         contained in the supplied set.
+ */
+std::string::const_iterator
+findNextNotCharacter(std::string::const_iterator fromIterator,
+                     std::string::const_iterator toIterator,
+                     const std::string &characters) {
+  for (auto it = fromIterator; it < toIterator; ++it) {
+    if (characters.find(*it) == std::string::npos)
+      return it;
+  }
+  return toIterator;
+}
+
+/**
+ * @return A vector of sub strings split around delimiters
+ */
+std::vector<std::string> splitBy(std::string::const_iterator fromIterator,
+                                 std::string::const_iterator toIterator,
+                                 const std::string &delimiters) {
+  std::vector<std::string> substrings;
+  auto iter = findNextNotCharacter(fromIterator, toIterator, delimiters);
+
+  while (iter != toIterator) {
+    const auto nextIter = findNextCharacter(iter, toIterator, delimiters);
+    substrings.emplace_back(iter, nextIter);
+    iter = findNextNotCharacter(nextIter, toIterator, delimiters);
+  }
+  return substrings;
+}
+
+/**
+ * @return A true if count1 is larger than count2
+ */
+bool compareCounts(const std::pair<std::string, std::size_t> &count1,
+                   const std::pair<std::string, std::size_t> &count2) {
+  return count1.second > count2.second;
+}
+
+/**
+ * @return True if the string is shorter than 5 characters
+ */
+bool isShorterThanFive(const std::string &str) { return str.length() < 5; }
+
+/**
+ * Removes any words shorter than five from the vector
+ */
+void removeShorterThanFive(std::vector<std::string> &strings) {
+  auto it = std::remove_if(strings.begin(), strings.end(), isShorterThanFive);
+  // Remove invalid indices
+  strings.erase(it, strings.end());
+}
+
+/**
+ * @return A string transformed to lower case
+ */
+std::string toLower(std::string str) {
+  std::transform(str.begin(), str.end(), str.begin(), std::tolower);
+  return str;
+}
+
+/**
+ * Add one to the count for a string, or creates a new mapping for a
+ * new string and sets its count to one.
+ */
+void addToCount(std::unordered_map<std::string, std::size_t> &count,
+                const std::string &str) {
+  auto it = count.find(str);
+  if (it == count.end())
+    count[str] = 1;
+  else
+    ++it->second;
+}
+
+/**
+ * Loops through the strings in substrings and records the
+ * number of occurences for each string.
+ */
+void countWords(const std::vector<std::string> &substrings,
+                std::unordered_map<std::string, std::size_t> &count) {
+  for (auto const &str : substrings)
+    addToCount(count, str);
+}
+
+/**
+ * Splits the string into each word, removes short words and changes
+ * each word to lower case before counting the number of occurances.
+ */
+void countWords(const std::string &str, const std::string &delimiters,
+                std::unordered_map<std::string, std::size_t> &count) {
+  auto words = splitBy(str.begin(), str.end(), delimiters);
+  removeShorterThanFive(words);
+  std::transform(words.begin(), words.end(), words.begin(), toLower);
+  countWords(words, count);
+}
+
+/**
+ * @return A vector pair containing each word paired to their usage count.
+ */
+WordCount countWords(std::istream &start, const std::string &delimiters) {
+  std::unordered_map<std::string, std::size_t> count;
+  for (std::string line; std::getline(start, line);)
+    countWords(line, delimiters, count);
+  return std::vector<std::pair<std::string, std::size_t>>(count.begin(),
+                                                          count.end());
+}
+
+/**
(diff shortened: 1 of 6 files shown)


exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <iterator>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+using WordCount = std::vector<std::pair<std::string, std::size_t>>;
```
samueljackson92: Good. Using a type alias here makes a lot of sense. 👍

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * @return An iterator pointing to the next letter contained
+ *         in the supplied set.
```
samueljackson92: Good that you described the `@return` of the function but what about the `@params`s? And a summary of what the function does.

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,203 @@
...
+ */
+std::string::const_iterator
+findNextCharacter(std::string::const_iterator fromIterator,
+                  std::string::const_iterator toIterator,
+                  const std::string &characters) {
+  for (auto it = fromIterator; it < toIterator; ++it) {
```
samueljackson92, line 30: Perhaps use `const auto` for clearer intent?

exercises/Robert_Applin/ex01_basics/src/main.cpp
samueljackson92: Also update other places where this is true.

exercises/Robert_Applin/ex01_basics/src/main.cpp
samueljackson92: Also update other places where this is true.

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,203 @@
...
+ */
+bool compareCounts(const std::pair<std::string, std::size_t> &count1,
+                   const std::pair<std::string, std::size_t> &count2) {
+  return count1.second > count2.second;
+}
+
+/**
+ * @return True if the string is shorter than 5 characters
+ */
+bool isShorterThanFive(const std::string &str) { return str.length() < 5; }
```
samueljackson92: This is pretty specific. Perhaps just move this to be a lambda function:

```cpp
auto it = std::remove_if(strings.begin(), strings.end(), [](const std::string& str) {
    return str.length() < 5;
});
```

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,203 @@
...
+  // Remove invalid indices
+  strings.erase(it, strings.end());
+}
+
+/**
+ * @return A string transformed to lower case
+ */
+std::string toLower(std::string str) {
+  std::transform(str.begin(), str.end(), str.begin(),
+                 static_cast<int (*)(int)>(std::tolower));
```
samueljackson92: Prefer a lambda over this weird cast to a function pointer:
```cpp
 std::transform(str.begin(), str.end(), str.begin(), [](const char c){ std::tolower(c); });
```

It should equate to the same thing in any decent compiler.

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Splits the string into each word, removes short words and changes
+ * each word to lower case before counting the number of occurances.
+ */
+void countWords(const std::string &str, const std::string &delimiters,
+                std::unordered_map<std::string, std::size_t> &count) {
+  auto words = splitBy(str.begin(), str.end(), delimiters);
+  removeShorterThanFive(words);
+  std::transform(words.begin(), words.end(), words.begin(), toLower);
```
samueljackson92: You wrote a `toLower` function, but you haven't used it here.

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/**
+ * @return A vector pair containing each word paired to their usage count.
+ */
+WordCount countWords(std::istream &start, const std::string &delimiters) {
+  std::unordered_map<std::string, std::size_t> count;
+  for (std::string line; std::getline(start, line);)
+    countWords(line, delimiters, count);
+  return std::vector<std::pair<std::string, std::size_t>>(count.begin(),
```
samueljackson92: `return WordCount(...)` instead?

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * @return A vector pair containing each word paired to their usage count.
+ */
+WordCount countWordsInFile(const std::string &filename,
+                           const std::string &delimiters) {
+  std::ifstream filestream(filename);
+  if (!filestream)
+    throw std::runtime_error("Error occured when attempting to load file: " +
+                             filename);
```
samueljackson92: You did not remember to close the file stream before throwing an exception. While this is a trivial toy example I would suggest using [RAII](https://en.cppreference.com/w/cpp/language/raii) instead.

Maybe write yourself a tiny file reading class to do this.

exercises/Robert_Applin/ex01_basics/src/main.cpp
samueljackson92: I would consider perhaps even type aliasing the `std::pair` as well as this is used in several places.

samueljackson92 (commented):
Generally very good. I have a few minor comments listed below. 

Nice to see a good use of standard algorithms throughout the code. It is also generally clean and well formatted. Good work.

exercises/Robert_Applin/ex01_basics/src/main.cpp
robertapplin: Ahh yes, that would have been much better!

exercises/Robert_Applin/ex01_basics/src/main.cpp
robertapplin: I think I have used it there

samueljackson92: Yes but I think you could have done:
```cpp
words = toLower(words)
```

?

exercises/Robert_Applin/ex01_basics/src/main.cpp
robertapplin: That looks much better!

exercises/Robert_Applin/ex01_basics/src/main.cpp
robertapplin: Yes, you are right!

exercises/Robert_Applin/ex01_basics/src/main.cpp
robertapplin: Good point, but it will be changing, so it can't be const. Is that right?

samueljackson92: Ah apologies. I think you're correct. The iterator needs to be mutable here. Forget this comment.

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -13,11 +13,15 @@
 #include <unordered_set>
 #include <vector>
 
-using WordCount = std::vector<std::pair<std::string, std::size_t>>;
+using WordCountPair = std::pair<std::string, std::size_t>;
+using WordCount = std::vector<WordCountPair>;
```
samueljackson92: 👍

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -13,11 +13,15 @@
...
 
 /**
+ * function finds the end point of a word.
+ * @params Iterators representing the start and end points of an
+ *         iteration. The characters used to split up each word.
```
samueljackson92: 👍

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -63,39 +73,42 @@ std::vector<std::string> splitBy(std::string::const_iterator fromIterator,
...
- * Removes any words shorter than five from the vector
+ * function removes any words shorter than five from the vector
+ * @params A vector of words
+ * @return void
  */
 void removeShorterThanFive(std::vector<std::string> &strings) {
-  auto it = std::remove_if(strings.begin(), strings.end(), isShorterThanFive);
+  auto it =
+      std::remove_if(strings.begin(), strings.end(),
+                     [](const std::string str) { return str.length() < 5; });
```
samueljackson92: 👍

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -63,39 +73,42 @@ std::vector<std::string> splitBy(std::string::const_iterator fromIterator,
...
 /**
+ * function transforms a string to lower case
+ * @params A string representing a word
  * @return A string transformed to lower case
  */
 std::string toLower(std::string str) {
   std::transform(str.begin(), str.end(), str.begin(),
-                 static_cast<int (*)(int)>(std::tolower));
+                 [](const char str2) { return std::tolower(str2); });
   return str;
```
samueljackson92: 👍

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -117,44 +131,57 @@ void countWords(const std::vector<std::string> &substrings,
...
+ *         around and a map containing each word and their count
+ * @return void
  */
 void countWords(const std::string &str, const std::string &delimiters,
                 std::unordered_map<std::string, std::size_t> &count) {
   auto words = splitBy(str.begin(), str.end(), delimiters);
   removeShorterThanFive(words);
-  std::transform(words.begin(), words.end(), words.begin(), toLower);
+  for (auto &word : words) {
+    word = toLower(word);
```
samueljackson92: I take it back. I thought here we were operating on a string, not a vector. You were correct to prefer the `std::transform`. My apologies it was late on Friday.

exercises/Robert_Applin/ex01_basics/src/main.cpp
```diff
@@ -117,44 +131,57 @@ void countWords(const std::vector<std::string> &substrings,
...
+ * function attempts to access the provided file
+ * @params The file name, and the characters each word will be split around.
  * @return A vector pair containing each word paired to their usage count.
  */
 WordCount countWordsInFile(const std::string &filename,
                            const std::string &delimiters) {
   std::ifstream filestream(filename);
-  if (!filestream)
+  if (!filestream) {
+    filestream.close();
```
samueljackson92: 👍

samueljackson92 (approved):

PR #37, opened 2018-07-12
Exercise 2 - Shape Sorter - Develop a program for sorting shapes and searching for shapes

robertapplin:
This PR adds a Shape Sorter program.
The Shape Sorter program sorts different shapes in order of volume and perimeter. It also allows you to search for a specific shape based on its type or number of edges.
Follows on from PR #36

### DIFF
--- a/exercises/Robert_Applin/ex02_oo_basics/inc/Circle.h
+++ b/exercises/Robert_Applin/ex02_oo_basics/inc/Circle.h
@@ -13,9 +13,9 @@ public:
   double getArea() const override;
 
 private:
-  double const m_radius;
-  double const m_area;
-  double const m_perimeter;
+  const double m_radius;
+  const double m_area;
+  const double m_perimeter;
 };
 
 #endif
--- a/exercises/Robert_Applin/ex02_oo_basics/inc/Circle.h
+++ b/exercises/Robert_Applin/ex02_oo_basics/inc/Circle.h
@@ -1,0 +1,21 @@
+#ifndef CIRCLE_H_
+#define CIRCLE_H_
+
+#include "Shape.h"
+
+class Circle : public Shape {
+public:
+  Circle(double radius);
+
+  std::string getType() const override;
+  std::size_t getNumberOfEdges() const override;
+  double getPerimeter() const override;
+  double getArea() const override;
+
+private:
+  double const m_radius;
+  double const m_area;
+  double const m_perimeter;
+};
+
+#endif
--- a/exercises/Robert_Applin/ex02_oo_basics/inc/Rectangle.h
+++ b/exercises/Robert_Applin/ex02_oo_basics/inc/Rectangle.h
@@ -1,0 +1,24 @@
+#ifndef RECTANGLE_H_
+#define RECTANGLE_H_
+
+#include "Shape.h"
+
+class Rectangle : public Shape {
+public:
+  Rectangle(double width, double height);
+
+  double getWidth() const;
+  double getHeight() const;
+  std::string getType() const override;
+  std::size_t getNumberOfEdges() const override;
+  double getPerimeter() const override;
+  double getArea() const override;
+
+private:
+  double const m_width;
+  double const m_height;
+  double const m_area;
+  double const m_perimeter;
+};
+
+#endif
--- a/exercises/Robert_Applin/ex02_oo_basics/inc/Rectangle.h
+++ b/exercises/Robert_Applin/ex02_oo_basics/inc/Rectangle.h
@@ -15,10 +15,10 @@ public:
   double getArea() const override;
 
 private:
-  double const m_width;
-  double const m_height;
-  double const m_area;
-  double const m_perimeter;
+  const double m_width;
+  const double m_height;
+  const double m_area;
+  const double m_perimeter;
 };
 
 #endif
--- a/exercises/Robert_Applin/ex02_oo_basics/inc/Shape.h
+++ b/exercises/Robert_Applin/ex02_oo_basics/inc/Shape.h
@@ -1,0 +1,14 @@
+#ifndef SHAPE_H_
+#define SHAPE_H_
+
+#include <string>
+
+class Shape {
+public:
+  virtual std::string getType() const = 0;
+  virtual std::size_t getNumberOfEdges() const = 0;
+  virtual double getPerimeter() const = 0;
+  virtual double getArea() const = 0;
+};
+
+#endif
--- a/exercises/Robert_Applin/ex02_oo_basics/inc/ShapeSorter.h
+++ b/exercises/Robert_Applin/ex02_oo_basics/inc/ShapeSorter.h
@@ -1,0 +1,29 @@
+#ifndef SHAPESORTER_H_
+#define SHAPESORTER_H_
+
+#include <ostream>
+#include <vector>
+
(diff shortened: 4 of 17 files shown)


exercises/Robert_Applin/ex02_oo_basics/inc/Circle.h
```diff
@@ -0,0 +1,21 @@
...
+public:
+  Circle(double radius);
+
+  std::string getType() const override;
+  std::size_t getNumberOfEdges() const override;
+  double getPerimeter() const override;
+  double getArea() const override;
+
+private:
+  double const m_radius;
```
samueljackson92: Using `const` is good 👍. But Mantid code style is west const and [west end functions](https://levelofindirection.com/blog/east-end-functions.html).

exercises/Robert_Applin/ex02_oo_basics/inc/ShapeSorter.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <ostream>
+#include <vector>
+
+// Forward declare Shape
+class Shape;
+
+class ShapeSorter {
+public:
+  ShapeSorter(std::vector<Shape const *> &&shapes);
```
samueljackson92, line 12: In c++11 (and upwards) you should almost never need to use a raw pointer. Prefer a `std::unique_ptr` and if that is completely impractical use a `std::shared_ptr`.

exercises/Robert_Applin/ex02_oo_basics/inc/Circle.h
```diff
@@ -0,0 +1,21 @@
+#ifndef CIRCLE_H_
+#define CIRCLE_H_
+
+#include "Shape.h"
+
+class Circle : public Shape {
+public:
+  Circle(double radius);
+
+  std::string getType() const override;
```
samueljackson92, line 10: I know this is a toy problem, but doxygen comments are required for methods.

exercises/Robert_Applin/ex02_oo_basics/inc/Shape.h
```diff
@@ -0,0 +1,14 @@
+#ifndef SHAPE_H_
+#define SHAPE_H_
+
+#include <string>
+
+class Shape {
+public:
+  virtual std::string getType() const = 0;
```
samueljackson92, line 8: Good using `const` on method definitions

exercises/Robert_Applin/ex02_oo_basics/inc/Rectangle.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include "Shape.h"
+
+class Rectangle : public Shape {
+public:
+  Rectangle(double width, double height);
+
+  double getWidth() const;
+  double getHeight() const;
+  std::string getType() const override;
```
samueljackson92, line 12: Good: using override  👍

exercises/Robert_Applin/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "Circle.h"
+#include "Rectangle.h"
+#include "Shape.h"
+#include "ShapeSorter.h"
+#include "Square.h"
+#include "Triangle.h"
+
+std::string toLower(std::string str) {
+  std::transform(str.begin(), str.end(), str.begin(),
+                 static_cast<int (*)(int)>(std::tolower));
```
samueljackson92: Again, same comment as in the first exercise. Prefer a lambda.

exercises/Robert_Applin/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+std::size_t getUnsignedFrom(std::istream &input) {
+  double value;
+  input >> value;
+
+  if (!input || value <= 0 || value - std::floor(value) != 0) {
+    input.clear();
+    throw std::runtime_error(
+        "Invalid input supplied: not a whole number >= 0.");
+  }
+  return static_cast<std::size_t>(value);
```
samueljackson92, line 43: :+1: not using a c style cast.

exercises/Robert_Applin/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+ShapeSorter
+ShapeSorter::filterByNumberOfEdges(std::size_t numberOfEdges) const {
+  return ShapeSorter(filterShapes(m_shapes, [&](Shape const *shape) {
+    return shape->getNumberOfEdges() == numberOfEdges;
+  }));
+}
+
+ShapeSorter ShapeSorter::sortByVolume() const {
+  return ShapeSorter(sortShapes(m_shapes, compareVolume));
```
samueljackson92: Perhaps just use a lambda instead of a separate member function.

exercises/Robert_Applin/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+ShapeSorter ShapeSorter::sortByPerimeter() const {
+  return ShapeSorter(sortShapes(m_shapes, comparePerimeter));
```
samueljackson92: Interesting choice to return a new shape sorter every time. I personally quite like the idea of immutable objects, but be careful about performance.

exercises/Robert_Applin/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,89 @@
...
+    : m_shapes(shapes) {}
+
+ShapeSorter ShapeSorter::filterByType(std::string const &type) const {
+  return ShapeSorter(filterShapes(
+      m_shapes, [&](Shape const *shape) { return shape->getType() == type; }));
+}
+
+ShapeSorter
+ShapeSorter::filterByNumberOfEdges(std::size_t numberOfEdges) const {
+  return ShapeSorter(filterShapes(m_shapes, [&](Shape const *shape) {
```
samueljackson92: Do you need to capture everything here? Perhaps just capture `this`

samueljackson92 (commented):
Again, generally very good. 👍

exercises/Robert_Applin/ex02_oo_basics/inc/ShapeSorter.h
robertapplin: I've been told that it is a raw non-owning pointer and that issues arise with bare new and delete and so in this circumstance it is okay to be used. std::unique_ptr would make immutability impossible here and std::shared_ptr would add reference counting, Does that sound right, because I don't know?

samueljackson92: I added this comment before seeing how you used it further down in the code. In this instance this works well because (as you say) `ShapeSorter` doesn't really own the pointers and you've been careful to make sure the lifetime of the shapes is longer than the `ShapeSorter`. And you are right that it avoid ref counting where (in this case) it's not needed.

I think you are correct to use a raw pointer here. 

I think the general advice still stands. But I will now caveat my advice "with unless you need a non-owning pointer to observe an object". 😄

PR #38, opened 2018-07-16
Solutions to exercise 1, Andre

aybamidele:


samueljackson92:
@aybamidele Would you like this to be reviewed or are you still working on it?

aybamidele:
@samueljackson92, yes I would like to have it reviewed

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include <iomanip>
+#include <utility>
+
+
+using std::string; using std::max; using std::find_if_not;
+using std::ifstream; using std::fstream;
+using std::getline; using std::map; using std::pair; using std::transform;
+ using std::endl; using std::vector;
+using std::cout; using std::find; using std::ofstream;
+using std::find_if; using std::vector; using std::setw;
```
samueljackson92: Two things here:
 - If you're going to use a lot of stuff from the same namespace, you can just use `using namespace std;`
 - In general it's bad style to put multiple using statements on the same line.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+bool pass(string str, string sample, size_t limit);
```
samueljackson92: I see what you were going for by forward declaring this function here, but I don't think we gain anything from defining it below the main function.

samueljackson92: `pass` is not a very descriptive name, it does not descriptive what this function actually does and doesn't hint at what the semantics of the parameters might be.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+int main(){
+	// vector of pairs to sort
+	vector < pair < string, int > > v;
```
samueljackson92: Please try to avoid single letter variable names (unless it's domain specific and understandable from context). We should try to make code as self documenting as possible. `v` is not a descriptive. I can see here it will contain and `string` and an `int` but the name `v` gives no hint as to what it will be used for.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    // max word length
+    size_t size = 0;
```
samueljackson92: 👍 initialising variables to something sensible on creation.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    //write each word into line
```
samueljackson92: 👍 Good that you're thinking about commenting your code. However, I would add that we don't need a comment for every line of code. Prefer self documenting code with clear, descriptive names and good code organisation. Inline comments are best used to explain something that isn't obvious from reading the code itself.

To quote from the book clean code by Bob Martin:
> A comment is a failure to express yourself in code.  If you fail, then write a comment; but try not to fail.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+    string line;
+
+    // create input file object
+    ifstream infile("Holmes.txt");
+
+	// create output file
+	ofstream outfile("Holmes Counter.txt");
```
samueljackson92: Be careful with your whitespace formatting. This should be indented to the same level as the line above. On the main project we actually use the `clang-format` tool to standardise code formatting across developers. 

However ensuring that you're code is properly formatted and indented to the right level will help you (and other people!) reason better about your program because it clearly signifies the scope of the current code block.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+    map<string, int> occurences;
+    
+    // open file for reading
+    if (infile.is_open()){
+
+        // read line
+        while(getline(infile, line)){
+            
+            // iterator to loop through each line
+            auto i = line.begin();
```
samueljackson92: 👍 for using `auto`.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+            while(i != line.end()){
+				
+                // ignore leading blanks
+				i = find_if_not(i, line.end(), ::isspace);
```
samueljackson92: We don't need the `::` in front of the `::isspace`, we can just use isspace. Same applies for else where in this program.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
samueljackson92: `i` is not descriptive. Worse, `i` is often used as the name for the integer counter variable in for loops. A better name would have been `iter` which suggests that it might be an iterator. An even better name would be something that indicates what we're iterating over. `lineIterator` perhaps?

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+				// ignore leading characters
+				i = find_if(i, line.end(), ::isalpha);
+
+				// if line is all whitespaces or has reached the end
+				if (i == line.end()) {
+					break;
+				}
+
+                // find end of next word
+                auto j = find_if(i, line.end(), ::isspace);
```
samueljackson92: Again, same problem as with `i` here.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
samueljackson92: 👍 for thinking about using standard library algorithms where you can.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+                // copy the characters in [i, j) and increment the counter
+                // or add to occurences after checking conditions
+				if (i != line.end()) {
+
+					// word
+					string word = string(i, j);
```
samueljackson92: Could have used `auto` here.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+					// make word lowercase
+					transform(word.begin(), word.end(), word.begin(), ::tolower);
+
+					// remove non alphabetic characters at the end of word
+					while (!isalpha(word.back())) {
```
samueljackson92: This approach works, but a better approach would probably have been to not include non-alphanumerics in the work in the first place. e.g. iterate over the line until you reach whitespace or punctuation, count the word, then skip over the whitespace/punctuation until you find the next valid character.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+						word = word.substr(0, word.size() - 1);
+					}
+				
+                    // check conditions and add to occurence
+                    if (!pass(word, ".,?'\"!():", 4)){
```
samueljackson92: I would make the string `".,?'\"!():"` a constant variable with a clear name:
```cpp
const auto punctuation {".,?'\"!():"};
if (!pass(word, punctuation, 4))
```

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+		infile.close();
+    
+    }else{
+        cout << "File could not be opened" << endl;
+    }
+
+	// sort occurences using vector of pairs (v)
+	for (auto i : occurences) {
+		v.push_back(i);
+	}
```
samueljackson92: 👍 that you're using the foreach loop.

samueljackson92: This can be better by making `i` a const reference (and using a better name for `i`):
```cpp
for (const auto& frequency : occurrences) {
    v.push_back(frequency);
}
```

samueljackson92: In fact, as you're just copying stuff to a vector you could have use the `std::copy` algorithm instead and copy over the `std::pair`s.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+	sort(v.begin(), v.end(), 
+		[](pair <string, int> a, pair <string, int> b){
+			return a.second > b.second;
+		});
```
samueljackson92: 👍 for using a lambda function here.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    // write the occurences from vector of pairs to a file
+	outfile << "Word" << setw(size - 2) << "Occurence" << setw(4) << endl;
+    for(auto i : v){
```
samueljackson92: `for (const auto& ...`

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+	outfile.close();
+
+	cout << "Done";		
+
+    return 0;
+}
+
+// check if a character in sample is in word
+// and see if the number of characters is > limit
+bool pass(string str, string sample, size_t limit) {
```
samueljackson92: We generally document cpp functions with doxygen style comments:
```
/**
 * Summary of what the function does
 *
 * @param str what this parameter does
 * @param sample what this parameter does
 * @param limit what this parameter does
 * @return what the function returns and what that means
 */
```

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+		for (auto i : str) {
+			if (sample.find(i) && !::isalpha(i)) {
+				return false;
+			}
+		}
+		return true;
+	}
+	else {
+		return false;
+	}
```
samueljackson92: Generally for C++ code with follow the [LLVM style guide](https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code). Which advocates early exits. We could rewrite this as follows:

```cpp
if (str.size() <= limit) {
    return false;
}

for (auto i : str) {
    if (sample.find(i) && !::isalpha(i)) {
        return false;
    }
}

return true;
```

Notice that this is ever so slightly less complex (one less level of indentation)

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+using std::string; using std::max; using std::find_if_not;
+using std::ifstream; using std::fstream;
+using std::getline; using std::map; using std::pair; using std::transform;
+ using std::endl; using std::vector;
+using std::cout; using std::find; using std::ofstream;
+using std::find_if; using std::vector; using std::setw;
+
+bool pass(string str, string sample, size_t limit);
+
+int main(){
```
samueljackson92: This function is way, way too long. It just about fits on my screen, but that's only because I have a 4K monitor. Consider how you can break down parts of the program into functions.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
samueljackson92: Also `std::find` is not used.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
samueljackson92: Also, `isalpha` returns an `int` which here is implictly casted to a `bool` we should explictly cast it with a `static_cast`:
```cpp
static_cast<bool>(isalpha(word.back())
```

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
samueljackson92: If also, if you know the size of the vector beforehand you should preallocate its size before you `push_back`. Here you know the size of occurrences so you should be able to do `v.reserve(occurences.size());` before the loop

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
samueljackson92: Also, to avoid a copy we can use constant references to the strings:
```cpp
bool pass(const string& str, ...)
```

samueljackson92 (changes requested):
I've made quite a few comments here. Feel free to come talk to me anytime about any of them if you're unsure about what I've said (or if you disagree!).

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
aybamidele: Do you mean instead of using string?

samueljackson92: Yes. On the right you are constructing a string, so the compiler should be able to deduce the type of `word` from this.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
aybamidele: I get this when I try to implement the static_cast:
Warning	C4800	'int': forcing value to bool 'true' or 'false' (performance warning)

aybamidele (commented):
I tried to include all the changes. I might change some things to pass the Jenkins checks

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include <cctype>
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <map>
+#include <string>
+#include <utility>
+#include <vector>
+
+using namespace std;
```
samueljackson92, line 11: 👍

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * Checks if a string is at least sizeLimit long and only has characters in
+ * allowedPunctuations
+ *
+ * @param word a string
+ * @param allowedPunctuations a string of not alphanumeric characters
+ * @param siseLimit size_t integer
+ * @return returns false if word has less than sizeLimit characters or if a non
+ * alphanumeric character in word is in punctuations
+ */
```
samueljackson92: This is much, much clearer 👍

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+bool checkSizeAndPunctuation(const string &word,
+                             const string &allowedPunctuations,
+                             const size_t &sizeLimit) {
```
samueljackson92: If `allowedPunctuations` is only used in this function then we can just move it inside here and have it as a local constant.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * @param character a character type
+ * @return returns true if character is whitespace or non alphanumeric
+ */
+bool isSpaceOrPunctuation(char character) {
+
+  if (isspace(character) || !isalpha(character))
+    return true;
+
+  return false;
```
samueljackson92: Why not just:

```cpp
return (isspace(character) || !isalpha(character))
```

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+  // vector of pairs to sort
+  vector<pair<string, int>> sortVector(map.begin(), map.end());
+
+  sort(sortVector.begin(), sortVector.end(),
+       [](pair<string, int> a, pair<string, int> b) {
+         return a.second > b.second;
+       });
+
+  return sortVector;
```
samueljackson92, line 63: This is very nice. 👍

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+ * Takes a map of <string, int> and returns a vector according to the values in
+ * the map.
+ *
+ * @param map a map of <string, int>
+ * @return returns the sorted vector of pair<string, int> according to the
+ * values in the map
+ */
+vector<pair<string, int>> maptoVecSorter(const map<string, int> &map) {
+
+  // vector of pairs to sort
```
samueljackson92: I would remove this comment. The names & documentation clearly convey what this is doing. This comment adds nothing.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+ * @param vecPairs vector of <string, int> pairs
+ * @param maxWordLength length of longest word, used to determine the space used
+ * in output stream
+ * @return returns 0 if successful
+ */
+
+int writeToFile(ostream &outfile, const vector<pair<string, int>> &vecPairs,
+                size_t maxWordLength) {
+
+  outfile << "Word" << setw(maxWordLength - 2) << "Occurrence" << setw(4)
```
samueljackson92: Here's an example of where a comment might be useful. Why do we subtract 2 here?

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+          << endl;
+
+  for (const auto &occurencePair : vecPairs) {
+    outfile << occurencePair.first
+            << setw(maxWordLength - occurencePair.first.size() + 2)
```
samueljackson92: Likewise why do we add 2 here?

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+  size_t maxWordLength = 0;
+  ifstream inFile("Holmes.txt");
+  ofstream outFile("Holmes Counter.txt");
+  map<string, int> occurences;
+  string lineOfText;
+
+  if (inFile.is_open()) {
+    while (getline(inFile, lineOfText)) {
+      transform(lineOfText.begin(), lineOfText.end(), lineOfText.begin(),
+                ::tolower);
```
samueljackson92, line 102: The `::` still needs to go.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+        // start of first word
+        lineIterator =
+            find_if_not(lineIterator, lineOfText.end(), isSpaceOrPunctuation);
+
+        // start of next word
+        auto nextWordIterator =
+            find_if(lineIterator, lineOfText.end(), isSpaceOrPunctuation);
+
+        // add word to occurrences
+        if (lineIterator != lineOfText.end()) {
```
samueljackson92: I think we can drop a level of indentation by flipping the condition and using a `continue`:

```cpp
if (lineIterator == lineOfTest.end())
    continue;

...
```

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+using namespace std;
+
+/**
+ * Checks if a string is at least sizeLimit long and only has characters in
+ * allowedPunctuations
+ *
+ * @param word a string
+ * @param allowedPunctuations a string of not alphanumeric characters
+ * @param siseLimit size_t integer
```
samueljackson92: Typo: sise

samueljackson92 (changes requested):

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
aybamidele: This creates an error on the RHEL though

samueljackson92: Ah sorry. You do need to do this here because of [this](https://stackoverflow.com/a/5539302)

aybamidele: Okay. Thanks.

exercises/ayomide_bamidele/ex01_basics/src/main.cpp
aybamidele: Moving ```allowedPunctuations``` forces me to use `string` instead of auto. This is okay?

samueljackson92: As discussed: Yes.

aybamidele:
I don't know why the windows check has failed. I just renamed a variable and it seems to have stopped passing the test.

aybamidele:
I reverted the changes and it does not seem to be a problem with changing the variable name.

aybamidele:
It works now

exercises/ayomide_bamidele/ex02_oo_basics/src/Circle.cpp
```diff
@@ -1,9 +1,7 @@
 #include "Circle.h"
 #include <iostream>
 
-Circle::Circle(double r) : radius(r) {
-  std::cout << "Circle Created" << std::endl;
-}
+Circle::Circle(double r) : radius(r) {}
 
 double Circle::area() const { return 3.14 * radius * radius; }
```
samueljackson92: There is a `M_PI` macro defined in c++ for the value of Pi that you can use. You'll need the following includes:
```cpp
#define _USE_MATH_DEFINES
#include <cmath>
```

exercises/ayomide_bamidele/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,16 @@
...
+class Shape {
+private:
+  std::string type = "Shape";
+
+public:
+  virtual std::string getType() const = 0;
+  virtual int getNoOfSides() const = 0;
+  virtual double area() const = 0;
+  virtual double perimeter() const = 0;
+  ~Shape() = default;
```
samueljackson92: We don't need to default the constructor here as it should already be set to the default

exercises/ayomide_bamidele/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,50 @@
...
+                << std::endl;
+    }
+  }
+  return 0;
+}
+
+int ShapeSorter::printByVolume(std::vector<Shape *> shapes) {
+
+  std::sort(shapes.begin(), shapes.end(),
+            [](Shape *&a, Shape *&b) { return a->area() < b->area(); });
```
samueljackson92: Good use of a lambda 👍

samueljackson92: And good use of std library algorithms

exercises/ayomide_bamidele/ex02_oo_basics/src/Square.cpp
```diff
@@ -0,0 +1,12 @@
+#include "Square.h"
+#include <iostream>
+
+Square::Square(double x) : length(x) {}
+
+double Square::area() const { return length * length; }
+
+double Square::perimeter() const { return 4 * length; }
```
samueljackson92: Sides is defined to be 4. Why not use `sides * length`

exercises/ayomide_bamidele/ex02_oo_basics/src/Square.h
```diff
@@ -0,0 +1,22 @@
+#ifndef SQUARE_H
+#define SQUARE_H
+#include "Shape.h"
+#include <string>
+
+class Square : public Shape {
+private:
+  std::string type = "Square";
+  int sides = 4;
+  static int id;
```
samueljackson92: Why does it need a `static id`?

samueljackson92: We should be able to make many of these member variables `const`

exercises/ayomide_bamidele/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+int main() {
+
+  std::cout << "Create 4 Shapes" << std::endl;
+  std::cout << "Circle: radius = 1" << std::endl;
+  std::cout << "Triangle: base = 1, height = 1" << std::endl;
+  std::cout << "Square: length = 1" << std::endl;
+  std::cout << "Rectangle: length = 1, breadth = 1\n" << std::endl;
+
+  auto circle = Circle(1.0);
```
samueljackson92: `Circle circle(1.0);` is shorter and simpler.

exercises/ayomide_bamidele/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,12 @@
+#include "Circle.h"
+#include <iostream>
+
+Circle::Circle(double r) : radius(r) {}
+
+double Circle::area() const { return 3.14 * radius * radius; }
```
samueljackson92: Member function should be comment with doxygen style comments much like free functions

exercises/ayomide_bamidele/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,50 @@
...
+                             const std::string &type) {
+  for (const auto &shape : shapes) {
+    if (shape->getType() == type) {
+      std::cout << shape->getType() << " has the right type\n" << std::endl;
+    }
+  }
+  return 0;
+}
+
+int ShapeSorter::printBySide(const std::vector<Shape *> &shapes,
```
samueljackson92: Why do these functions need to return `int`? Why not just `void`?

samueljackson92 (changes requested):
Generally pretty good. A few comments:

exercises/ayomide_bamidele/ex02_oo_basics/src/Circle.cpp
aybamidele: Alright.

exercises/ayomide_bamidele/ex02_oo_basics/src/ShapeSorter.cpp
aybamidele: I'll implement this

exercises/ayomide_bamidele/ex02_oo_basics/src/main.cpp
aybamidele: Alright

exercises/ayomide_bamidele/ex02_oo_basics/src/Square.h
aybamidele: It doesn't. I forgot to remove that. Yes they could be `const`

exercises/ayomide_bamidele/ex02_oo_basics/src/Square.cpp
aybamidele: Well, that's true. I guess I like it this way because it's slicker mathematically: You only use one variable.

aybamidele: Is using `sides * length` better ?

exercises/ayomide_bamidele/ex02_oo_basics/src/Shape.h
aybamidele: Okay. I didn't know that. In my defense, I was just doing what David said.

exercises/ayomide_bamidele/ex02_oo_basics/src/Circle.cpp
aybamidele: Yes, I thought about that to be honest. I thought it'd be best to implement the core of the problem first before worrying about `pi`

aybamidele: I'll implement it

exercises/ayomide_bamidele/ex02_oo_basics/src/Shape.h
samueljackson92: The rules for what is declared by default are actually quite complicated depending on what you explicitly declare. The following table is a good reference for what you define vs. what the state of the special member function is:

screen shot 2018-07-20 at 08 56 19

exercises/ayomide_bamidele/ex02_oo_basics/src/Circle.cpp
samueljackson92: >  I thought it'd be best to implement the core of the problem first before worrying about pi

No worries. I'm only pointing things out to make you aware 😄

exercises/ayomide_bamidele/ex02_oo_basics/src/Square.cpp
samueljackson92: > I guess I like it this way because it's slicker mathematically: You only use one variable.

You might think so, but I would argue that this is not true. Compilers are very smart and good at optimising your code. With optimisation flags turned on a good compiler is going to see that variable and optimise it away. 

Don't believe me? Here's your [code vs. mine](https://godbolt.org/g/E5iMem) in the extremely useful [compiler explorer](https://godbolt.org/). The assembly on the right shows that clang 6.0.0 outputs the same code regardless of whether we use a literal or a const member variable.

So I would argue the we should use `sides * length` because it's more descriptive (even thought yes I can guess 4 means the number of sides).

Even if this was not true and using the member variable was slightly slower I would still argue that you use the variable over the literal. In general prefer to write good readable code over "fast" code. Then *measure* the performance and think about how you can improve speed *if it matters*. If code isn't in a critical high performance area then it doesn't need to be the fastest possible code; it is more important that it is clear, readable and maintainable.

exercises/ayomide_bamidele/ex02_oo_basics/src/Shape.h
aybamidele: And here, did you mean constructor or destructor? I thought the tilde meant that it's a destructor...

exercises/ayomide_bamidele/ex02_oo_basics/src/Square.cpp
aybamidele: Alright, I understand this.

exercises/ayomide_bamidele/ex02_oo_basics/src/Shape.h
samueljackson92: I meant destructor. Sorry.

aybamidele:
Please, could you look at the changes especially in the lambda function, and help me understand why making those changes helped me pass the RHEL check? I sort of understand pointers and referencing but why is it that using them both here created an error on RHEL? @samueljackson92

samueljackson92 (approved):

samueljackson92 (changes requested):
Accidentally approved.

samueljackson92 (approved):
I think this is looking good now.

PR #39, opened 2018-10-10
Dolica's C++ exercises

DolicaAkelloEgwel:


exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+	return words;
+}
+std::string PrepareWord(std::string rawWord)
+{
+	// Declare a string of punctuation characters
+	std::string unwantedChars = ".,?'\"!():;";
+
+	// Remove the unwanted characters from the string
+	for (int i = 0; i < unwantedChars.length(); i++)
```
OwenArnold: This would probably work better with iterators given the remove-erase below.

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+	{
+		rawWord.erase(std::remove(rawWord.begin(), rawWord.end(), unwantedChars[i]), rawWord.end());
```
OwenArnold: Good use of erase-remove idiom

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+		// Add the word to the vector
+		words.push_back(segment);
+	}
+
+	return words;
+}
+std::string PrepareWord(std::string rawWord)
+{
+	// Declare a string of punctuation characters
+	std::string unwantedChars = ".,?'\"!():;";
```
OwenArnold: Could you do this using `boost::regex` i.e.`regex_search` I think this could be a useful further exercise. Would be a better solution that the character-by character match.

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+	// Remove the unwanted characters from the string
+	for (int i = 0; i < unwantedChars.length(); i++)
+	{
+		rawWord.erase(std::remove(rawWord.begin(), rawWord.end(), unwantedChars[i]), rawWord.end());
+	}
+
+	// Convert the word to lower case
+	for (unsigned int i = 0; i < rawWord.length(); ++i)
+	{
+		rawWord[i] = std::tolower(rawWord[i]);
```
OwenArnold: Why not do this first with the incoming raw string? Reduce allocations and would also make the pattern matching above easier to write.

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return rawWord;
+}
+void SaveWordCountToFile(std::map<std::string, int> wordMap, std::string outputFilename)
```
OwenArnold: Looks like map is read-only, should pass as const-ref.

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+	// Check that the file was opened successfully
+	if (!outFile)
+	{
+		std::cout << "Error: Unable to open the file " << outputFilename << " for writing." << std::endl;
+		exit(1);
+	}
+
+	// Declare a comparator for sorting the map by word count
+	auto cmp = [](const auto &p1, const auto &p2)
```
OwenArnold: I would extract this as a named function at the point I was assigning the lambda to a variable. Things start to get messy otherwise.

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	inFile.close();
+
+	return wordMap;
+}
+int main(int argc, char *argv[])
+{
+	// Declare a map for storing the word counts
+	std::map<std::string, int> wordMap;
```
LamarMoore: Declare variables where they will be used.

OwenArnold (changes requested):
Feedback from partial review of changes. I need to find time for a more thorough look through it before I approve it. Here are some changes I request you make.

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	// Declare a comparator for sorting the map by word count
+	auto cmp = [](const auto &p1, const auto &p2)
+	{
+		return p2.second < p1.second || !(p1.second < p2.second) && p1.first < p2.first;
+	};
+
+	// Declare a set for storing words by word count
+	std::set < std::pair<std::string, size_t>, decltype(cmp)> s(wordMap.begin(), wordMap.end(), cmp);
```
LamarMoore: Could you just sorted the map?

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+		outFile << p.first << "\t" << p.second << std::endl;
+	}
+
+	outFile.close();
+
+}
+std::map<std::string, int> CountWords(std::string inputFilename)
+{
+	// Declare a map for storing the word counts
+	std::map<std::string, int> wordMap;
```
LamarMoore: Again declare variables at point of use.

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+	std::string words;
+
+	// Attempt to open the text file
+	std::ifstream inFile;
+	inFile.open(inputFilename);
+
+	// Check that the file was opened successfully
+	if (!inFile)
+	{
+		std::cout << "Error: Unable to open the file " << inputFilename << " for reading." << std::endl;
```
LamarMoore, line 87: 👍 error handling

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+			// Remove dashes from the words if any are present
+			splitWords = SplitWords(words);
+		}
+		else
+		{
+			// Place the single word in the vector when no dashes are present
+			splitWords.push_back(words);
+		}
+
+		for (auto word = begin(splitWords); word != end(splitWords); ++word)
```
LamarMoore, line 121: Do you think there is any way this step and the previous step could be combined? i.e could you call `PrepareWord()` and add to the map as you read the file?

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+		if (segment.length() == 0)
+			continue;
+
+		// Add the word to the vector
+		words.push_back(segment);
+	}
+
+	return words;
+}
+std::string PrepareWord(std::string rawWord)
```
LamarMoore: consider passing by const ref `...(const std::string &rawWord)` in cases where you only need an immutable reference.

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include <iostream>
+#include <fstream>
+#include <sstream>
+#include <map>
+#include <cctype>
+#include <algorithm>
+#include <vector>
+#include <set>
+
+bool ContainsDashes(std::string compositeWord)
```
LamarMoore: const ref argument.

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+{
+	return compositeWord.find("-") != std::string::npos;
+}
+std::vector<std::string> SplitWords(std::string compositeWord)
```
LamarMoore: const ref argument.

LamarMoore (commented):
Here is my code review of exercise 1. Overall very good solution, nicely laid out and very readable. Is there a reason you chose not to use a `WordProcessor` class for example? (Probably a bad name). You do not have to change that but your thoughts on that would be good.

exercises/dolica_akelloegwel/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,31 @@
...
+#include <string>
+#include <iostream>
+#include <math.h>
+#include <vector>
+#include <algorithm>
+
+class Shape
+{
+protected:
+	double area;
```
LamarMoore: In our coding standards all class member variables are prefixed `m_`

exercises/dolica_akelloegwel/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,31 @@
...
+	std::string printMessage;
+	void SetPerimeter(double p);
+	void SetArea(double a);
+	void SetSides(int s);
+	std::string CreateBaseMessage();
+	void SetMessage(std::string msg);
+
+public:
+	Shape(std::string t, int n);
+	double GetPerimeter() const;
```
LamarMoore: Did you consider using virtual methods so that you can reap the benefits of polymorphism? (Talk to me about this if you need more explanation).

exercises/dolica_akelloegwel/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include <string>
+#include <iostream>
+#include <math.h>
+#include <vector>
+#include <algorithm>
+
+class Shape
```
LamarMoore, line 9: Shape should have a virtual destructor.

exercises/dolica_akelloegwel/ex02_oo_basics/src/Shape.cpp
```diff
@@ -0,0 +1,62 @@
...
+	// Compute during subclass initialisation
+	perimeter = 0;
+	area = 0;
+	printMessage = "";
+}
+double Shape::GetPerimeter() const
+{
+	return perimeter;
+}
+void Shape::SetPerimeter(double p)
```
LamarMoore: In general we stay away from single letter variables in an api. However, in cases like these where it is painfully obvious it's fine.

LamarMoore (commented):
Once again, well structured and documented. I noticed you avoided using virtualisation completely. Might be worth thinking about the pros and cons of using pure virtual methods in a your base class in this situation. You can leave a note here as a response to this review or we can talk about it.

matthew-d-jones:
This looks good :)
I notice that there are a lot of comments in the code though. Comments should generally be used to explain _why_ something is done a particular way (in cases that it is not obvious) rather than _what_ the code is doing, which should be self-evident. A code block with an attached comment is usually a sign that the code should be extracted as a separate method, using a descriptive function name instead of the comment.

exercises/dolica_akelloegwel/ex01_basics/src/main.cpp
```diff
@@ -105,13 +93,25 @@ std::map<std::string, int> CountWords(std::string inputFilename)
...
+	std::regex unwantedChars("[.,?'\"!():;]");
 
 	// Traverse through the file
 	while (!inFile.eof())
 	{
 		// Store the word in a string
 		inFile >> words;
 
+		// Convert the word(s) to lower case
+		for (unsigned int i = 0; i < words.length(); ++i)
```
LamarMoore: this seems like a good candidate for `std::transform` see str_tolower [here](https://en.cppreference.com/w/cpp/string/byte/tolower)

exercises/dolica_akelloegwel/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,44 @@
...
+	Circle myCircle = Circle(circleRadius);
+
+	double triangleBase = 3;
+	double triangleHeight = 4;
+	Triangle myTriangle = Triangle(triangleBase, triangleHeight);
+
+	// Create a vector and place the shapes in it
+	std::vector<Shape> myShapes;
+
+	myShapes.push_back(mySquare);
```
OwenArnold: Your solution relies of slicing. The way this is implemented is not good (see below) If you really want to go down this route, you need to go the full template or CRTP route. Runtime virtualisation is an acceptable solution though. 

For your ref: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-slice they even use the same example with shape!

exercises/dolica_akelloegwel/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,32 @@
...
+
+public:
+	Shape(std::string t, int n);
+	~Shape();
+	double GetPerimeter() const;
+	double GetArea() const;
+	std::string GetType() const;
+	int GetSides() const;
+	std::string GetMessage() const;
+	friend std::ostream &operator<<(std::ostream & os, const Shape &shape);
```
OwenArnold: Interesting solution here. You survive the slicing because you perform the calculations in the constructor. Here are a few problems:

1. Your solution would not work if say your shape was mutable. i.e. change radius of circle shape then get area.
1. You can construct a "Shape" stand alone because it has a public constructor and is not pure-virtual. What does an object of type "Shape" mean? Can I have a Shape that isn't really any Shape?

Also your Shape is type is not just contractural "fulfil these methods I don't care how", but implicitly implies how it can be done.

exercises/dolica_akelloegwel/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,60 @@
...
+void ShapeSorter::PrintBasedOnNSides(std::vector<Shape> shapes, int numSides)
+{
+	// Print shapes that have a given number of sides
+	for (std::vector<Shape>::iterator it = shapes.begin(); it != shapes.end(); it++)
+	{
+		if (it->GetSides() == numSides)
+			std::cout << *it << std::endl;
+	}
+}
+void ShapeSorter::SortByAreaDesc(std::vector<Shape> shapes)
```
OwenArnold: Your function does two things. It sorts and prints. Avoid giving functions multiple responsibilities. The name also does not imply that it's doing two things.

exercises/dolica_akelloegwel/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+	double circleRadius = 5;
+	Circle myCircle = Circle(circleRadius);
+
+	double triangleBase = 3;
+	double triangleHeight = 4;
+	Triangle myTriangle = Triangle(triangleBase, triangleHeight);
+
+	// Create a vector and place the shapes in it
+	std::vector<Shape> myShapes;
```
OwenArnold: For a large vector this could be problematic. The vector will be 24 bytes + n*size_of(Shape). Also copies of vectors will force deep copies of your shape.

OwenArnold (changes requested):

exercises/dolica_akelloegwel/ex02_oo_basics/src/Shape.h
```diff
@@ -9,24 +9,15 @@
...
-	void SetPerimeter(double perimeter);
-	void SetArea(double area);
-	void SetSides(int sides);
-	std::string CreateBaseMessage();
-	void SetMessage(std::string msg);
+	std::string CreateBaseMessage() const;
 
 public:
-	Shape(std::string t, int n);
+	Shape();
```
OwenArnold: Minor point, this constructor is useless and possibly misleading. I would remove it.

exercises/dolica_akelloegwel/ex02_oo_basics/src/Rectangle.cpp
```diff
@@ -1,32 +1,41 @@
...
-}
\ No newline at end of file
+}
+std::string Rectangle::GetMessage() const
+{
+	// Construct and set the message that will be displayed when the object is printed
+	std::string rectMessage = "Rectangle Properties: \n";
+	rectMessage += "  Width: " + std::to_string(m_width) + "\n";
+	rectMessage += "  Height: " + std::to_string(m_height);
+	rectMessage += CreateBaseMessage();
```
OwenArnold: `CreateBaseMessage` comes from the base `Shape`. I think it's worth separating where you are using a base class to provide common implementation (as here), and where you are using a base class to provide polymorphism. These are completely different reasons. Note that in the latter your concrete classes are only tied to the signature of the pure virtual methods, in the former there is also implications on HOW the internals should be performed. 

Try moving the `CreateBaseMessage` into a helper method. Maybe in a new header `ShapeImplementationCommon.h`?

exercises/dolica_akelloegwel/ex02_oo_basics/src/main.cpp
```diff
@@ -25,7 +25,7 @@ int main(int, char **)
 	Triangle myTriangle = Triangle(triangleBase, triangleHeight);
 
 	// Create a vector and place the shapes in it
-	std::vector<Shape> myShapes;
+	std::vector<std::reference_wrapper<Shape> > myShapes;
```
OwenArnold: Your `reference_wrapper` is safe, as currently written, and an interesting approach, but could result in unusual/undefined behaviour with some minor modification. Try storing the `Shape` pointers as `std::unique_ptr<Shape>` ask me for help if you have any questions.

exercises/dolica_akelloegwel/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+	return copyShapes;
+}
+std::vector<std::reference_wrapper<Shape>> ShapeSorter::SortByPerimeterDesc(std::vector<std::reference_wrapper<Shape>> shapes)
+{
+	// Create a copy of the shapes vector
+	std::vector<std::reference_wrapper<Shape>> copyShapes(shapes);
+
+	// Create a lambda expression for sorting by perimeter values
+	auto perimCompare = [](const Shape &s1, const Shape &s2)
```
OwenArnold: Would be better as named free function `bool PerimeterComparitor(const Shape &s1, const Shape &s2)` would improve readability in my opinion.

exercises/dolica_akelloegwel/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -1,5 +1,13 @@
 #include "ShapeSorter.h"
 
+bool ShapeSorter::AreaComparator(const Shape &s1, const Shape &s2)
```
OwenArnold: Excellent. Now it would be possible to write unit tests for the comparators to check that they work without testing everything else.

exercises/dolica_akelloegwel/ex02_oo_basics/Shapes.cbp
```diff
@@ -0,0 +1,111 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
OwenArnold: Please don't commit non-cmake project files (i.e. generated files to the project) adding these to a [gitignore](https://git-scm.com/docs/gitignore) file is would be a good idea. Also make sure any generated file go to a separate top-level directory. i.e. use "out of source" builds.

exercises/dolica_akelloegwel/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -1,17 +1,18 @@
...
 {
 public:
-	static bool AreaComparator(const Shape &s1, const Shape &s2);
-	static bool PerimeterComparator(const Shape &s1, const Shape &s2);
-	void PrintBasedOnType(std::vector<std::reference_wrapper<Shape>> shapes, std::string type);
-	void PrintBasedOnNSides(std::vector<std::reference_wrapper<Shape>> shapes, int numSides);
-	std::vector<std::reference_wrapper<Shape>> SortByAreaDesc(std::vector<std::reference_wrapper<Shape>> shapes);
-	std::vector<std::reference_wrapper<Shape>> SortByPerimeterDesc(std::vector<std::reference_wrapper<Shape>> shapes);
-	void PrintShapes(std::vector<std::reference_wrapper<Shape>> shapes);
+	static bool AreaComparator(const std::unique_ptr<Shape> &s1, const std::unique_ptr<Shape> &s2);
```
OwenArnold: 1. I think that we should probably change all of these to be `std::unique_ptr<const Shape>` since we do not want to allow shape to be modified once created?
1. use of `std::unique_ptr<Shape>` (at present) is common and a little long-winded. how about declaring 
```using Shape_uptr = std::unique_ptr<Shape>;``` or ```using Shape_cuptr = std::unique_ptr<const Shape>;``` to clean up?

File: exercises/dolica_akelloegwel/ex02_oo_basics/src/main.cpp [hunk 31]
```diff
@@ -11,26 +11,22 @@ int main(int, char **)
...
 
 	// Create a vector and place the shapes in it
-	std::vector<std::reference_wrapper<Shape> > myShapes;
+	std::vector<std::unique_ptr<Shape>> myShapes;
 
-	myShapes.push_back(mySquare);
-	myShapes.push_back(myRectangle);
-	myShapes.push_back(myCircle);
-	myShapes.push_back(myTriangle);
+	myShapes.push_back(std::make_unique<Square>(squareSide));
```
OwenArnold: I would either change the declarations above to be `const`, i.e. `circleRadius`, or better yet, change the push_backs to include in-line comments:
```cpp
myShapes.push_back(std::make_unique<Rectangle>(2 /*recWidth*/, 5 /*recHeight*/))```

File: exercises/dolica_akelloegwel/ex02_oo_basics/tests/CircleTests.cpp [hunk 32]
```diff
@@ -0,0 +1,24 @@
+//
+// Created by dolica on 16/10/18.
+//
+
+#include <gtest/gtest.h>
+#include "../src/Circle.h"
+
+class CircleTests : public ::testing::Test
```
OwenArnold, line 8: Well done!

OwenArnold (commented):
Good work @DolicaAkelloEgwel

dtasev:
I feel this PR has run its course so I am closing it to keep the open PRs relevant

PR #40, opened 2019-01-17
exercise 1 solution Alice Russell

alicerussell1:
Can someone provide a code review for my solution. I have a test file in the test directory named Holmes.

### DIFF
--- a/exercises/alice_russell/ex01_basics/src/main.cpp
+++ b/exercises/alice_russell/ex01_basics/src/main.cpp
@@ -11,6 +11,7 @@
 #include <iterator>
 #include <iomanip>
 
+using WordFrequency = std::pair<std::string, int>;
 
  //returns true if the character is punctuation
 bool find_punctuation(char c)
@@ -65,7 +66,7 @@ bool compare(std::pair<std::string, int> a, std::pair<std::string, int> b)
 }
 
 //counts and sorts the words
-std::vector<std::pair<std::string, int> > count_words(std::vector<std::string> s)
+std::vector<std::pair<std::string, int> > count_words(const std::vector<std::string>& s)
 {
 
 	std::map<std::string, int> count;
@@ -82,6 +83,46 @@ std::vector<std::pair<std::string, int> > count_words(std::vector<std::string> s
 	return ordered;
 }
 
+void count_words(std::istream& in, std::ostream& out)
+{
+	std::string s;
+	std::vector<std::string> find;
+	while (std::getline(in, s))
+	{
+		auto line = find_words(s);
+		find.insert(find.end(), line.cbegin(), line.cend());
+	}
+	std::vector<WordFrequency> words = count_words(find);
+
+	//formatting output
+	//find longest word
+	unsigned int max_length = 0;
+	for (const auto &wordAndCount : words)
+	{
+		max_length = (wordAndCount.first.length() > max_length) ? wordAndCount.first.length() : max_length;
+	}
+
+	out << std::left << std::setw(max_length) << "Word" << "\t" << "Usage" << std::endl << std::endl;
+	for (const auto &wordAndCount : words)
+	{
+		out << std::left << std::setw(max_length) << wordAndCount.first << "\t" << wordAndCount.second << std::endl;
+	}
+}
+
+void count_words_in_file(const std::string & in_filename, const std::string & out_filename)
+{
+	std::ifstream infile(in_filename);
+	std::ofstream outfile(out_filename);
+	if (infile.is_open())
+	{
+		count_words(infile, outfile);
+	}
+	else
+	{
+		throw std::invalid_argument("Unable to open input file: " + in_filename);
+	}
+}
+
 int main(int argc, char **argv)
 { 
 
@@ -91,41 +132,7 @@ int main(int argc, char **argv)
 	}
 	else
 	{
-		std::ifstream infile(argv[1]);
-		std::ofstream outfile("results");
-		if (infile)
-		{
-			std::string s;
-			std::vector<std::string> find;
-			while (std::getline(infile, s))
-			{
-				std::vector<std::string> line = find_words(s);
-				for (std::vector<std::string>::const_iterator it = line.begin(); it != line.end(); ++it)
-				{
-					find.push_back(*it);
-				}
-			}
-			std::vector<std::pair<std::string, int> > words = count_words(find);
-
-			//formatting output
-			//find longest word
-			int max_length = 0;
-			for (std::vector<std::pair<std::string, int> >::const_iterator i = words.begin(); i != words.end(); ++i)
-			{
-				max_length = (i->first.length() > max_length) ? i->first.length() : max_length;
-			}
-
-
-			outfile << std::left << std::setw(max_length) << "Word" << "\t"  << "Usage" << std::endl << std::endl;
-			for (std::vector<std::pair<std::string, int> >::const_iterator i = words.begin(); i != words.end(); ++i)
-			{
-				outfile << std::left << std::setw(max_length) << i->first << "\t"  << i->second << std::endl;
-			}
-		}
-		else
-		{
-			std::cerr << "cannot open file " << argv[1] << std::endl;
-		}
+		count_words_in_file(argv[1], "results");
 	}
 	return 0;
 }  
--- a/exercises/alice_russell/ex01_basics/src/main.cpp
+++ b/exercises/alice_russell/ex01_basics/src/main.cpp
@@ -1,0 +1,131 @@
+/**
+ * Skeleton main routine
+ */
+
+#include <iostream>
+#include <vector>
+#include <string>
+#include <algorithm>
+#include <fstream>
+#include <map>
+#include <iterator>
+#include <iomanip>
+
+
+ //returns true if the character is punctuation
+bool find_punctuation(char c)
+{
+	static const std::string punc = ".,?'\"!():";
+	return (std::find(punc.begin(), punc.end(), c) != punc.end());
+}
+
+std::vector<std::string> find_words(const std::string& s)
+{
+	std::vector<std::string> ret;
+	typedef std::string::size_type string_size;
+	string_size i = 0;
+
+	
+	while (i != s.size())
+	{
+		//finds first character which is not a space or hyphen
+		while (i != s.size() && (isspace(s[i]) || s[i] == '-'))
+		{
+			++i;
+		}
+		//finding the end of the word
+		string_size j = i;
+		while (j != s.size() && !isspace(s[j]) && s[j] != '-')
+		{
+			++j;
+		}
+
+		std::string word = s.substr(i, j - i);
+		//removing any punctuation within the word
+		word.erase(std::remove_if(word.begin(), word.end(), find_punctuation), word.end());
+
+		
+
+		//checks if the word is longer than 4 characters 
+		if (word.length() > 4)
+		{
+			//make word lower case
+			std::transform(word.begin(), word.end(), word.begin(), tolower);
+
+			ret.push_back(word);
+		}
+		i = j;
+	}
+	return ret;
+}
+
+bool compare(std::pair<std::string, int> a, std::pair<std::string, int> b)
+{
+	return a.second > b.second;
+}
+
+//counts and sorts the words
+std::vector<std::pair<std::string, int> > count_words(std::vector<std::string> s)
+{
+
+	std::map<std::string, int> count;
+	for (std::vector<std::string>::size_type i = 0; i != s.size(); ++i)
+	{
+		++count[s[i]];
+	}
+
+	std::vector<std::pair<std::string, int> > ordered;
+	std::copy(count.begin(), count.end(), back_inserter(ordered));
+
+	std::sort(ordered.begin(), ordered.end(), compare);
+
+	return ordered;
+}
+
+int main(int argc, char **argv)
+{ 
+
+	if (argc > 2)
+	{
+		std::cerr << "too many files" << std::endl;
+	}
+	else
+	{
+		std::ifstream infile(argv[1]);
+		std::ofstream outfile("results");
+		if (infile)
+		{
+			std::string s;
+			std::vector<std::string> find;
+			while (std::getline(infile, s))
+			{
+				std::vector<std::string> line = find_words(s);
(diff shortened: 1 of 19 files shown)


exercises/alice_russell/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+int main(int argc, char **argv)
+{ 
+
+	if (argc > 2)
+	{
+		std::cerr << "too many files" << std::endl;
+	}
+	else
+	{
+		std::ifstream infile(argv[1]);
```
martyngigg: I would consider putting all of the code in this `else` block in a function, something like `count_words_in_file(const std:string &in_filename, const std::string & out_filename)`. This way `main` is simple to read as it's just check arguments and call this function with the argument.

exercises/alice_russell/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+			outfile << std::left << std::setw(max_length) << "Word" << "\t"  << "Usage" << std::endl << std::endl;
+			for (std::vector<std::pair<std::string, int> >::const_iterator i = words.begin(); i != words.end(); ++i)
+			{
+				outfile << std::left << std::setw(max_length) << i->first << "\t"  << i->second << std::endl;
+			}
+		}
+		else
+		{
+			std::cerr << "cannot open file " << argv[1] << std::endl;
```
martyngigg: Good checking that the input can be opened as a file. Had you considered that the output file may not open correctly too?

In Mantid we would use an exception here rather than printing to the screen. Something like `throw std::invalid_argument("Unable to open input file: " + std::to_string(argv[1]))`.

exercises/alice_russell/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+	{
+		std::cerr << "too many files" << std::endl;
+	}
+	else
+	{
+		std::ifstream infile(argv[1]);
+		std::ofstream outfile("results");
+		if (infile)
+		{
+			std::string s;
```
martyngigg: It is generally considered good practice, once a `if/else` block becomes more than a few lines to abstract this code to a function. In C++ you can overload functions based on argument type so you could have a function `count_words(std::istream& in, std::ostream& out)` so that `count_words_in_file` just opens and checks the files are okay then passes them to `count_words`.

This way it separates out the concerns of each part of the code and allows you to focus on what a particular part of  the code is doing in isolation. This is also important for testing when it comes to that in Mantid.

exercises/alice_russell/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+			std::vector<std::string> find;
+			while (std::getline(infile, s))
+			{
+				std::vector<std::string> line = find_words(s);
+				for (std::vector<std::string>::const_iterator it = line.begin(); it != line.end(); ++it)
```
martyngigg: It would be a bit more concise (and more performance) here to write `find.insert(line.cbegin(), line.cend());`

exercises/alice_russell/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+	{
+		std::ifstream infile(argv[1]);
+		std::ofstream outfile("results");
+		if (infile)
+		{
+			std::string s;
+			std::vector<std::string> find;
+			while (std::getline(infile, s))
+			{
+				std::vector<std::string> line = find_words(s);
```
martyngigg: In "modern" C++there is a keyword called `auto` that can be used to deduce the type of an expression so this could be written as
```
auto line = find_words(s);
```
This style is called [almost always auto](http://cginternals.github.io/guidelines/articles/almost-always-auto/) and can drastically reduce the amount of boilerplate code and increase readability.

exercises/alice_russell/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+				{
+					find.push_back(*it);
+				}
+			}
+			std::vector<std::pair<std::string, int> > words = count_words(find);
+
+			//formatting output
+			//find longest word
+			int max_length = 0;
+			for (std::vector<std::pair<std::string, int> >::const_iterator i = words.begin(); i != words.end(); ++i)
```
martyngigg: :+1: for the const iterator!

Unfortunately the good introductory C++ books have not yet caught up with recent advances in the language and this can be written far more concisely these days using a [range-based for loop](https://docs.microsoft.com/en-us/cpp/cpp/range-based-for-statement-cpp?view=vs-2017)
```c++
for(const auto &wordAndCount: words) {
    max_length = (wordAndCount.first.length() > max_length) ? wordAndCount.length() : max_length;
}

File: exercises/alice_russell/ex01_basics/src/main.cpp [hunk 7]
```diff
@@ -0,0 +1,131 @@
...
+			{
+				max_length = (i->first.length() > max_length) ? i->first.length() : max_length;
+			}
+
+
+			outfile << std::left << std::setw(max_length) << "Word" << "\t"  << "Usage" << std::endl << std::endl;
+			for (std::vector<std::pair<std::string, int> >::const_iterator i = words.begin(); i != words.end(); ++i)
```
martyngigg: Again this can also use a range-based for loop.

exercises/alice_russell/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+	return ret;
+}
+
+bool compare(std::pair<std::string, int> a, std::pair<std::string, int> b)
+{
+	return a.second > b.second;
+}
+
+//counts and sorts the words
+std::vector<std::pair<std::string, int> > count_words(std::vector<std::string> s)
```
martyngigg: In C++ functions should generally use whats called "pass by reference" as the default in C++ is to copy the function arguments and particularly for things like vectors this is expensive. The function signature would just become
```c++
 std::vector<std::pair<std::string, int> > count_words(const std::vector<std::string>& s)
```

exercises/alice_russell/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+
+	std::map<std::string, int> count;
+	for (std::vector<std::string>::size_type i = 0; i != s.size(); ++i)
+	{
+		++count[s[i]];
+	}
+
+	std::vector<std::pair<std::string, int> > ordered;
+	std::copy(count.begin(), count.end(), back_inserter(ordered));
```
martyngigg, line 79: Good use of standard algorithms :+1:

exercises/alice_russell/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+			std::vector<std::string> find;
+			while (std::getline(infile, s))
+			{
+				std::vector<std::string> line = find_words(s);
+				for (std::vector<std::string>::const_iterator it = line.begin(); it != line.end(); ++it)
+				{
+					find.push_back(*it);
+				}
+			}
+			std::vector<std::pair<std::string, int> > words = count_words(find);
```
martyngigg: The type `std::pair<std::string, int>` is used quite frequently so it could be good to consider creating a typedef, basically giving it an alias name. If you put something like `using WordFrequency =  std::pair<std::string, int>;` just underneath your includes at the top then you can use this name to refer to it and increase the readability of some expressions.

martyngigg (changes requested):
This is a great start, we've got some functions and good use of the standard library.

I've added some inline comments where we can tidy up a bit.

exercises/alice_russell/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include <string>
+
+class Shape {
```
martyngigg, line 9: This is a pretty good class definition. It is generally a good idea to avoid prefixing method names with redundant words such as `get` and in the case where methods will just give properties, like these, then just use things like `perimeter()`, `area()`, `sides()`, `type()`.

`print_details` is an example where the qualifier is necessary to understand that there will be a side effect or some computation will probably need to happen. `find_area` could be `calculate_area` if we thought that some significant calculation could be involved.

exercises/alice_russell/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
```
martyngigg: This works well in Visual Studio but it's not what you'll see in Mantid as it's not guaranteed to work in all of our environments. In Mantid you'll see the use of 
```c++
#ifndef SHAPE_H_
#define SHAPE_H_

... class and function definitions

#endif //SHAPE_H_

File: exercises/alice_russell/ex02_oo_basics/src/Circle.cpp [hunk 13]
```diff
@@ -0,0 +1,24 @@
+#include "Circle.h"
+
+static const double PI = 3.141592653589793238463;
+
+double Circle::find_perimeter() const
```
martyngigg: In Mantid you will also see methods and functions documented using [Doxygen](https://wiki.scilab.org/Doxygen%20documentation%20Examples).

exercises/alice_russell/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,24 @@
+#include "Circle.h"
+
+static const double PI = 3.141592653589793238463;
```
martyngigg: Good extraction of a constant out to a name and this is good practice to get into generally. Just as note that in Mantid we would use the definition of pi called `M_PI` which you can get from `#include <cmath>`.

exercises/alice_russell/ex02_oo_basics/src/Square.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+#include "Shape.h"
+
+class Square : public Shape
```
martyngigg, line 6: Good. Some people inherit rectangle here but that's not the correct way to go.

exercises/alice_russell/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,51 @@
+#include "ShapeSorter.h"
+
+bool ShapeSorter::compare_area(Shape* &s1, Shape* &s2)
```
martyngigg: A more standard signature for this type of method would be `bool ShapeSorter::compare_area(const Shape &s1, const Shape &s2)`. It reads as two references to a const `Shape`" object so `compare_area` can't change either of them. You would generally not see a signature such as `Shape * &` as copying pointers is basically free and a pointer allows you to change the object it points to.

martyngigg: This is also the type of method where documentation is crucial so that it can be described what compare means in this context, i.e. compare returns true iff the first shape has a greater area than the second.

exercises/alice_russell/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,12 @@
...
+
+#include <string>
+
+class Shape {
+public:
+	virtual double find_perimeter() const = 0;
```
martyngigg: Good use of `const` on these member functions.

exercises/alice_russell/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,16 @@
...
+
+
+
+class Circle : public Shape
+{
+private:
+	const double radius;
+public:
+	Circle(double r) :radius(r) {}
+	double find_perimeter() const;
```
martyngigg: In modern C++ you will see compilers warn that methods like this can be marked "override". This essentially means you would write `double find_perimeter() const override;` to indicate that you are intending to override the method specified in `Shape`. That way the compiler can spot occasions where parameters might be incorrect and a function does not override as intended.

exercises/alice_russell/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+	return s1->find_area() > s2->find_area();
+}
+
+bool ShapeSorter::compare_perimeter(Shape* &s1, Shape* &s2)
+{
+	return s1->find_perimeter() > s2->find_perimeter();
+}
+
+void ShapeSorter::print_type(std::ostream &out, std::vector<Shape*> &all, std::string &type) const
```
martyngigg: `const` is also important in function definitions when using `&` to ensure that a function does not change that parameter. For example, `type` would be written `const std::string &type` as it shouldn't change that. `out` however has to be changed by that function so it is not `const` and can't be.

martyngigg: There are similar cases in all of the following methods.

martyngigg (commented):
This is generally a good solution for exercise 2. I've again made some inline comments again.

Something generally to be aware of is that for Mantid we use a system called CMake. When we add files, such as header or source files, we also need to update files called [`CMakeLists.txt`](https://github.com/mantidproject/newstarter/blob/alice_russell_exercises/exercises/alice_russell/ex02_oo_basics/CMakeLists.txt) with references to these new files.

If you look at the main page of the pull request then at the bottom you will see that some status checks have failed. On Mantid we will need to make these green before a piece of work would go up for review.

exercises/alice_russell/ex02_oo_basics/src/ShapeSorter.cpp
alicerussell1: Removing the pointer leads to an error using sort in print_area_descending. Is there a way to make it so the object can't be change but keeping the pointer? Would 
```
bool ShapeSorter::compare_area(const Shape * const s1, const Shape * const s2)
``` 
be sufficient?

martyngigg: Ah yes of course - I'd missed the `std::sort` call. Your suggested solution would also work.

martyngigg (approved):
This looks like a good set of solutions now.

martyngigg:
Closing without merging so the solutions don't end up in the starter code for future.

PR #42, opened 2019-06-13
added classes for triangle, circle. square, and rectangle, with base …

TakudzwaMakoni:
[Following exercise description (exercise 2)](https://www.mantidproject.org/New_Starter_C%2B%2B_introduction#Exercise)

**For tester:**
please review code.
do not merge until passes CI.

### DIFF
--- a/exercises/takudzwa_makoni/ex02_oo_basics/CMakeLists.txt
+++ b/exercises/takudzwa_makoni/ex02_oo_basics/CMakeLists.txt
@@ -1,0 +1,28 @@
+cmake_minimum_required ( VERSION 2.8.5 )
+
+###############################################################################
+# Setup
+###############################################################################
+# Name
+project ( Shapes )
+
+# Find custom cmake modules
+set ( CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../../../CMake")
+
+# Common setup
+include ( CommonSetup )
+
+###############################################################################
+# Executable
+###############################################################################
+
+set ( SRC_FILES
+  src/main.cpp
+  # add any additional source files here
+)
+
+set ( HDR_FILES
+  # add any additional header files here
+)
+
+add_executable ( Shapes ${SRC_FILES} ${HDR_FILES} )
--- a/exercises/takudzwa_makoni/ex02_oo_basics/CMakeLists.txt
+++ b/exercises/takudzwa_makoni/ex02_oo_basics/CMakeLists.txt
@@ -23,6 +23,14 @@ set ( SRC_FILES
 
 set ( HDR_FILES
   # add any additional header files here
+  src/shape.h
+  src/square.h
+  src/rectangle.h
+  src/isoTriangle.h
+  src/circle.h
+  src/shapeSorter.h
+
+
 )
 
 add_executable ( Shapes ${SRC_FILES} ${HDR_FILES} )
--- a/exercises/takudzwa_makoni/ex02_oo_basics/CMakeLists.txt
+++ b/exercises/takudzwa_makoni/ex02_oo_basics/CMakeLists.txt
@@ -23,7 +23,7 @@ set ( SRC_FILES
   src/rectangle.cpp
   src/isoTriangle.cpp
   src/shapeSorter.cpp
-  src/shape.cpp
+
   # add any additional source files here
 )
 
--- a/exercises/takudzwa_makoni/ex02_oo_basics/CMakeLists.txt
+++ b/exercises/takudzwa_makoni/ex02_oo_basics/CMakeLists.txt
@@ -18,6 +18,12 @@ include ( CommonSetup )
 
 set ( SRC_FILES
   src/main.cpp
+  src/circle.cpp
+  src/square.cpp
+  src/rectangle.cpp
+  src/isoTriangle.cpp
+  src/shapeSorter.cpp
+  src/shape.cpp
   # add any additional source files here
 )
 
--- a/exercises/takudzwa_makoni/ex02_oo_basics/src/circle.h
+++ b/exercises/takudzwa_makoni/ex02_oo_basics/src/circle.h
@@ -8,7 +8,7 @@ class circle : public shape { // circle class
 private:
 
 	
-	const std::string m_name = "CIRCLE"; // make static
+	static std::string m_name;
 	const int m_numOfSides = 1;
 	const double m_pi = 3.14159;
 	double m_radius;
--- a/exercises/takudzwa_makoni/ex02_oo_basics/src/circle.h
+++ b/exercises/takudzwa_makoni/ex02_oo_basics/src/circle.h
@@ -8,7 +8,7 @@ class circle : public shape { // circle class
 private:
 
 	
-	static std::string m_name;
+	static const std::string m_name;
 	const unsigned int m_numOfSides = 1;
 	const double m_pi = 3.14159;
 	double m_radius;
--- a/exercises/takudzwa_makoni/ex02_oo_basics/src/circle.h
+++ b/exercises/takudzwa_makoni/ex02_oo_basics/src/circle.h
@@ -1,0 +1,31 @@
+#ifndef CIRCLE_H
+#define CIRCLE_H
+
+class circle : public shape { // circle class
+
+private:
+
+	
+	const std::string name;
+	const int numOfSides;
+	const double pi = 3.14159;
+
+public:
+
+	double radius;
+
+	virtual int getNumOfSides() const override { return numOfSides; }
+	virtual std::string getName() const override { return name; }
+
+	virtual double getArea() const override { return pi * radius * radius; }
+	virtual double getPerimeter() const override { return pi * 2 * radius; }
+
+	//constructor
+	circle::circle() : name("circle"), numOfSides(1) {
+
+		radius = 1;
+	};
+
+};
+
+#endif
--- a/exercises/takudzwa_makoni/ex02_oo_basics/src/circle.h
+++ b/exercises/takudzwa_makoni/ex02_oo_basics/src/circle.h
@@ -1,29 +1,32 @@
 #ifndef CIRCLE_H
 #define CIRCLE_H
 
+
+
 class circle : public shape { // circle class
 
 private:
 
 	
-	const std::string name;
-	const int numOfSides;
-	const double pi = 3.14159;
+	const std::string m_name = "CIRCLE"; // make static
+	const int m_numOfSides = 1;
+	const double m_pi = 3.14159;
+	double m_radius;
+
 
 public:
 
-	double radius;
+	 
+	virtual int numOfSides() const override { return m_numOfSides; }
+	virtual std::string name() const override { return m_name; }
 
-	virtual int getNumOfSides() const override { return numOfSides; }
-	virtual std::string getName() const override { return name; }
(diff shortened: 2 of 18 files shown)


exercises/takudzwa_makoni/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,163 @@
+
+//#include "pch.h"
+#include <iostream>
+#include <stdio.h>
+#include <math.h>
+#include <vector>
+#include <string>
+
+const double pi = (double)3.14159; // constant Pi casted to double
```
OwenArnold: c-style cast `(double)` does nothing useful. In fact avoid c-style casts in all future code.

exercises/takudzwa_makoni/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+class shape {
+
+public:
+	virtual void getNumOfSides()	{ cout << "override" << endl; }
+	virtual void getName()			{ cout << "override" << endl; }
+
+	virtual void getArea()			{ cout << "override" << endl; }
+	virtual void getPerimeter()		{ cout << "override" << endl; }
+};
```
OwenArnold: `shape` should be pure virtual IMHO - Unless you have a good reason otherwise?

OwenArnold: member functions should not mutate `shape`, should be marked `const`.

exercises/takudzwa_makoni/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+
+//--------------------------------------------------------------------------
+
+//------------------------- shape subclasses -------------------------------
+
+class square : public shape {
+
+private:
```
OwenArnold: 👍 for keeping internal data private

exercises/takudzwa_makoni/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+	double sideLength;
+	string name;
+	int numOfSides;
+
+public:
+
+	virtual void getNumOfSides() { cout << numOfSides << endl; }
```
OwenArnold: You have function called `getNumOfSides()` but it doesn't return what is expected and is mixed up with IO. Use separate `printNumOfSides` to do the IO.

exercises/takudzwa_makoni/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,163 @@
...
+	virtual void getName() { cout << name << endl; }
+
+	virtual void getArea()		{ cout << sideHeight * sideWidth << endl; }
+	virtual void getPerimeter()	{ cout << (2 * sideWidth) + (2 * sideHeight) << endl; }
+
+	rectangle::rectangle() {
```
OwenArnold: Prefix `rectangle::` in the declaration is a compiler error (see CI).

exercises/takudzwa_makoni/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,163 @@
...
+	virtual void getNumOfSides()	{ cout << numOfSides << endl; }
+	virtual void getName()			{ cout << name << endl; }
+
+	virtual void getArea()		{ cout << 0.5 * base* height << endl; }
+	virtual void getPerimeter() { cout << 2 * (sqrt(height * height) + ((base * base) / 4)) << endl; }
+
+	//constructor
+	isoTriangle::isoTriangle() {
+		base	= 1;
+		height	= 2;
```
OwenArnold: How did you decide that the `base` and `height` were 1 and 2 respectively - if I were to use your `isoTriangle` how would I be able to provide `base`, `height`?

exercises/takudzwa_makoni/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+int main() {
+
+
+	square x;
+	circle y;
+	rectangle z;
+	isoTriangle p;
+
+	vector<shape*> shapeVector({ &x, &y, &z, &p });
```
OwenArnold: Interesting approach - technically not wrong but you should ensure you are familiar with `smart pointers` (I think Scott Meyers has a chapter specifically on the `vector` problem you encounter here) so that you can code correctly where scope changes mean that ownership transfer is required.

OwenArnold (changes requested):
Quick initial review. A lot of what you have done is in the right direction.

exercises/takudzwa_makoni/ex02_oo_basics/src/main.cpp
TakudzwaMakoni: ive modified to return instead of cout

exercises/takudzwa_makoni/ex02_oo_basics/src/circle.h
```diff
@@ -0,0 +1,31 @@
...
+	double radius;
+
+	virtual int getNumOfSides() const override { return numOfSides; }
+	virtual std::string getName() const override { return name; }
+
+	virtual double getArea() const override { return pi * radius * radius; }
+	virtual double getPerimeter() const override { return pi * 2 * radius; }
+
+	//constructor
+	circle::circle() : name("circle"), numOfSides(1) {
```
OwenArnold: `circle::circle()` ... `rectangle::rectangle` are causing compiler errors on RHEL7. We discussed already.

OwenArnold: `circle::circle()` ... `rectangle::rectangle` are causing compiler errors on RHEL7. We discussed already.

OwenArnold: I want to set the radius of my circle. I think that's a reasonable thing to do. Please change the constructor to allow the variable parts of the type to be "injected" so constructor should look like `circle(double rad) : radius(rad) {...}` or similar

exercises/takudzwa_makoni/ex02_oo_basics/src/isoTriangle.h
```diff
@@ -0,0 +1,31 @@
+#ifndef ISOTRIANGLE_H
+#define ISOTRIANGLE_H
+
+class isoTriangle : public shape { // isosceles triangle class
+
+private:
+
+	
+	const std::string name;
```
OwenArnold: Consider making shape names static - they do not change from instance to instance.

exercises/takudzwa_makoni/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "shape.h"
+#include "square.h"
+#include "rectangle.h"
+#include "isoTriangle.h"
+#include "circle.h"
+#include "shapeSorter.h"
+
+
+square s; circle c; rectangle r; isoTriangle t;
+std::vector<shape*> shapeVector({ &s, &c, &r, &t });
```
OwenArnold: Already discussed `std::vector<unique_ptr<shape>>`. Also should not be in defined global scope.

exercises/takudzwa_makoni/ex02_oo_basics/src/rectangle.h
```diff
@@ -0,0 +1,29 @@
+#ifndef RECTANGLE_H
+#define RECTANGLE_H
+
+class rectangle : public shape {
+
+private:
+
+	
+	const std::string name;
+	const int numOfSides;
```
OwenArnold: Can your rectangle have a negative number of sides? I would recommend using `unsigned int` for these cases (see other shape subtypes with similar problems too).

exercises/takudzwa_makoni/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,15 @@
+#ifndef SHAPE_H
+#define SHAPE_H
+
+class shape {
+
+public:
+	virtual int getNumOfSides() const = 0;
```
OwenArnold: This looks good now.

OwenArnold (changes requested):

exercises/takudzwa_makoni/ex02_oo_basics/src/circle.h
TakudzwaMakoni: i've made radius in circle class private instead of public, and written get/set member functions for all shapes. circle::circle() changed to circle() etc. ive put default values in the initialiser list.

exercises/takudzwa_makoni/ex02_oo_basics/src/shapeSorter.h
```diff
@@ -0,0 +1,93 @@
...
+
+		std::cout << "\n\n";
+
+	};
+
+	void printMatchForType(std::string chosenType) {
+
+		std::cout << "\nMATCHES FOR TYPE \"" << chosenType << "\"\n";
+		for (size_t i = 0; i != shapes.size(); i++){
+		std::transform(chosenType.begin(), chosenType.end(), chosenType.begin(), std::toupper);
```
OwenArnold: @Milli-Lab just to help you a bit with the GCC errors. change `std::toupper` to `::toupper`. The std::toupper overload does not have the signature we need.

exercises/takudzwa_makoni/ex02_oo_basics/src/isoTriangle.h
```diff
@@ -8,12 +8,12 @@ class isoTriangle : public shape { // isosceles triangle class
 private:
 
 	static std::string m_name;
```
OwenArnold: Unless you want the name to change, i would make this `const` too.

exercises/takudzwa_makoni/ex02_oo_basics/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+set ( HDR_FILES
+  # add any additional header files here
+  src/shape.h
+  src/square.h
+  src/rectangle.h
+  src/isoTriangle.h
+  src/circle.h
+  src/shapeSorter.h
+
```
OwenArnold, line 38: I like the matching now between physical and logical units in your codebase 👍 

Note that you should put implementations in `cpp` files.  Please attempt this for say `class circle`, keep declarations in the header, but move the definitions into a new `src/circle.cpp`.

TakudzwaMakoni: will add cpp files for the classes👍

OwenArnold: Just doing one is fine.

exercises/takudzwa_makoni/ex02_oo_basics/src/shapeSorter.h
TakudzwaMakoni: replaced it👍

exercises/takudzwa_makoni/ex02_oo_basics/src/isoTriangle.h
TakudzwaMakoni: replaced with member function shapeType()

OwenArnold:
Considered complete

PR #43, opened 2019-07-12
Phil Colebrooke Exercises

### LINKED ISSUE #46: Complete Exercise 2
Mantid is in desperate need of a way to store and view different shapes.

Write a command line program that:

1. Has classes to allow number of shapes to be defined: square (side1), rectangle(side1, side2), circle(radius), triangle(height, base).
    1. Each shape class should know it's type ("Square"), how many sides it has.
    2. Each shape needs to be able to calculate it's perimeter and area. For the triangle you can assume it is isoceles and the perimeter can be computed using p = b + 2\sqrt{h^2+(b^2/4)}, where b is the base and h is the height.
2. Within the Main method create a variety of the shapes and put them in a std::vector
3. Create a class ShapeSorter which should contain four methods
    1. Print out the Shapes that match a chosen type
    2. Print out the Shapes that match a chosen number of sides
    3. Print out the Shapes in order of volume descending
    4. Print out the Shapes in order of perimeter descending

PhilColebrooke:
**Description of Work**
Solution to Exercise 1. Takes a text file as input and creates a file called output.txt which lists every word in the inputted text file (with length greater than 4) and the number of times it occurs in the file. Provide one text file as a command-line argument.

Test file: [Holmes.txt](https://github.com/martyngigg/cpp-examples/raw/master/Holmes.txt)

**Description of Work 2**
Solution to Exercise 2. Includes classes that allow various shapes to be defined, populates a vector with shapes and outputs information about them. No arguments required to run.

Fixes #46.


exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
+#include "main.h"
+
+int main(int argc, char* argv[])
+{
+    std::string filename{};
+    std::map<std::string, int> wordCounter{};
+    std::vector<std::pair<int, std::string>> sorted{};
+    const int correctNumberArguments{ 2 };
```
ghost: Constants like this should be in all capitals and have words separated with underscores. This lets us know these are constants at a glance. See our [C++ coding standards](http://developer.mantidproject.org/Standards/CPPStandards.html#naming-conventions).

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
+#include "main.h"
+
+int main(int argc, char* argv[])
+{
+    std::string filename{};
+    std::map<std::string, int> wordCounter{};
```
ghost: I don't know if `Counter` is the right terminology here. This variable is where you store your counts, it's not an object or function that's doing counting. A more appropriate name may be something like `wordCounts` or `wordsAndCounts`

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+    std::vector<std::pair<int, std::string>> sorted{};
+    const int correctNumberArguments{ 2 };
+
+    // checks that the user has provided only one filename
+    if (argc != correctNumberArguments) {
+        std::cerr << "Wrong number of arguments provided. "
+            "Provide only one filename.";
+        return 0;
```
ghost: By convention, if a function or program exits having failed, we want to return some non-zero error code

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+    }
+    else {
```
ghost: The `else` here is not actually needed since the function exits when the `if` clause evaluates to `true`. This is a bit nit-picky, but it saves a couple of lines.

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    writeToFile(sorted);
+
+    return 0;
+}
+
+// opens file with the provided filename
+// populates a map with each word in the file
+// and the number of occurences
+void countWords(std::string filename,
```
ghost: We can pass `filename` by reference here, I think (you'll have to change the function definition in the header as well)

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+    std::map<std::string, int>& wordCounter) {
+
+    std::ifstream file{ filename };
+
+    // checks the file has been opened
+    if (!file.is_open()) {
+        std::cerr << "Unable to open file.";
```
ghost: When printing to standard out, it's nearly always a good idea to end the string with a new line. This separates the statement from any other statements that may follow.

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+    countWords(filename, wordCounter);
+
+    sorted = orderByUsage(wordCounter);
+
+    writeToFile(sorted);
+
+    return 0;
+}
+
+// opens file with the provided filename
```
ghost: Nice descriptive doc-string 👍. At Mantid, we use Doxygen to auto-generate documentation from our doc-strings. This means we need to follow the Doxygen standard, see our [Function Header Standards](http://developer.mantidproject.org/Standards/CPPStandards.html#function-headers).

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+    // checks the file has been opened
+    if (!file.is_open()) {
+        std::cerr << "Unable to open file.";
+        return;
+    }
+
+    // goes through each word in the file
+    std::string word;
+    while (file >> word) {
+        const int minWordLength{ 5 };
```
ghost: You're declaring `minWordLength` on every iteration of this loop. I would personally have this variable as an argument to `countWords`, this would make the function more general, and therefore more reusable

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+            ++wordCounter[word];
+        }
+    }
+
+    file.close();
+}
+
+// given a word, removes punctuation and converts to lowercase
+// returns the changed word
+std::string cleanUpWord(std::string word) {
```
ghost: We can pass by reference here as well, I think

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+    std::string newWord{};
+    std::string punctuation{ ".,?'\"!():" };
+
+    // iterates through each character in the word
+    for (char c : word) {
```
ghost: Nice range based `for` loop 👍. 

Note you can actually loop over `word` by reference - `for (char &c : word){}` - and use `std::tolower()`, to convert the `char` to lowercase. So this loop could be condensed to:
```cpp
for (char &c : word) {
  if (punctuation.find(c) == std::string::npos)
    newWord += std::tolower(c);
}
```

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+        newWord += c;
+    }
+
+    return newWord;
+}
+
+// converts the map to a vector of pairs so it can be sorted
+std::vector<std::pair<int, std::string>>
+orderByUsage(std::map<std::string, int>& wordCounter) {
```
ghost: Nice uses of `std::map` and `std::pair` 👍

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+    if (argc != correctNumberArguments) {
+        std::cerr << "Wrong number of arguments provided. "
+            "Provide only one filename.";
+        return 0;
+    }
+    else {
+        filename = argv[1];
+    }
+
+    countWords(filename, wordCounter);
```
ghost: Currently, if `countWords` fails, we're still running the rest of `main` and eventually writing `output.txt` with no words in it. If we return 0/1 or `true`/`false` (i.e. success/failure) from `countWords`, we can exit main early on a failure.

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+// converts the map to a vector of pairs so it can be sorted
+std::vector<std::pair<int, std::string>>
+orderByUsage(std::map<std::string, int>& wordCounter) {
+
+    std::vector<std::pair<int, std::string>> pairs{};
+
+    // each word-usage pair in the map is added to the vector
+    // as a usage-word pair (swapped for easy sorting)
+    for (auto word : wordCounter) {
```
ghost: We can loop by reference here; avoiding making a copy of every `word` in `wordCounter`

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+    return pairs;
+}
+
+// writes each word-usage pair to a file
+void writeToFile(std::vector<std::pair<int, std::string>> wordCounter) {
+    std::ofstream outputFile{ "output.txt" };
+
+    // checks the file has been opened
+    if (!outputFile.is_open()) {
+        std::cerr << "Unable to open file.";
```
ghost: This error message could be a little more descriptive. What file can't you open and are you trying to write to it or read it?

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+        return;
+    }
+
+    // headings are written to the file
+    outputFile << "Word\t\t\t\tUsage\n\n";
+
+    // iterates through the vector of pairs
+    for (std::pair<int, std::string> wordAndUsage : wordCounter) {
+        const int paddedWordLength{ 32 };
```
ghost: Variable declarations should generally not be inside `for` loops, especially `const` ones

exercises/Phil_Colebrooke/ex01_basics/src/main.h
```diff
@@ -0,0 +1,21 @@
...
+void countWords(std::string filename, std::map<std::string, int>& wordCounter);
+
+std::string cleanUpWord(std::string word);
+
+std::vector<std::pair<int, std::string>>
+orderByUsage(std::map<std::string, int>& wordCounter);
+
+void writeToFile(std::vector<std::pair<int, std::string>> wordCounter);
+
+#endif EXERCISE1_H
```
ghost: `#endif` doesn't take any arguments, so this gives me a compiler warning on GCC. It is generally good practice to specify which `if` is ending via a comment though:
```cpp
#endif // EXERCISE1_H
```

exercises/Phil_Colebrooke/ex01_basics/src/main.h
```diff
@@ -0,0 +1,21 @@
+#ifndef EXERCISE1_H
+#define EXERCISE1_H
+
+#include <iostream>
+#include <string>
+#include <fstream>
+#include <map>
+#include <vector>
+#include <algorithm>
+#include <functional>
```
ghost: At Mantid, we always alphabetize our includes http://developer.mantidproject.org/Standards/CPPStandards.html#layout-of-header-file

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,135 @@
...
+    // sorted in descending order of word usage
+    std::sort(pairs.begin(), pairs.end(),
+        std::greater<std::pair<int, std::string>>());
+
+    return pairs;
+}
+
+// writes each word-usage pair to a file
+void writeToFile(std::vector<std::pair<int, std::string>> wordCounter) {
+    std::ofstream outputFile{ "output.txt" };
```
ghost: For reusability, I would have `filename` as a function argument here, it could even be an optional argument with a default value of `"output.txt"`

ghost (changes requested):
This a very good first attempt. I like your approach using `std::map` and `std::pair`. I've left some comments below, let me know if you have any questions.

For formatting I highly recommend getting a `clang-format` extension on whatever IDE you're using (you can find the one for Visual Studio [here](https://marketplace.visualstudio.com/items?itemName=LLVMExtensions.ClangFormat)). On Mantid, we use clang-format to *try* and achieve consistent formatting across our code base; nothing that isn't "clang-formatted" goes into master. If you haven't already, give our [C++ coding standards](http://developer.mantidproject.org/Standards/CPPStandards.html) a read.

PhilColebrooke:
With the improved solution, provide 3 command-line arguments to run the program:
- The name of the input text file.
- The name of the output text file.
- The minimum word length.

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -1,135 +1,167 @@
...
-
-    writeToFile(sorted);
+int main(int argc, char *argv[]) {
+  const int CORRECT_NUMBER_ARGUMENTS{4};
+  std::string inputFilename;
+  std::string outputFilename;
+  int minWordLength;
+  std::map<std::string, int> wordsAndCounts;
+  std::vector<std::pair<int, std::string>> sorted;
+  bool success{false};
```
ghost: In general, I'd recommend declaring variables as close to their first usage as possible. This makes it easier to read (as you're not scrolling up the page to find the type of the variable you're using later) and means you're not creating objects that may not be used. For example, if the below check of `CORRECT_NUMBER_ARGUMENTS` fails, then you've created a bunch of objects that aren't ever called. 

In most of these cases declaring on assignment is fine, especially if the variable should be a `const` anyway. e.g. line 21 could be `const std::string inputFilename = argv[1];`. Declaring on assignment also means you can use `auto` instead of having to write out long templated types like `std::vector<std::map<string, int>>`

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -1,135 +1,167 @@
...
-    // each word-usage pair in the map is added to the vector
-    // as a usage-word pair (swapped for easy sorting)
-    for (auto word : wordCounter) {
-        std::pair<int, std::string> pair;
-        pair.second = word.first;
-        pair.first = word.second;
+  // each word-usage pair in the map is added to the vector
+  // as a usage-word pair (swapped for easy sorting)
+  for (auto &word : wordsAndCounts) {
+    std::pair<int, std::string> pair;
```
ghost: I'd move this declaration up one line, so it's outside the loop

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -1,135 +1,167 @@
...
+                std::map<std::string, int> &wordsAndCounts,
+                const int &minWordLength) {
+
+  std::ifstream file{inputFilename};
+
+  // checks the file has been opened
+  if (!file.is_open()) {
+    std::cerr << "Unable to open input file for reading. Check that the file "
+                 "exists and its name is spelt correctly."
+              << "/n";
```
ghost: Wrong kind of slash here, I think

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -1,135 +1,167 @@
...
-    }
+/** Converts a word to lowercase and removes punctuation
+ *
+ * @param word The word to be cleaned up
+ * @return The cleaned up word
+ *
+ */
+std::string cleanUpWord(std::string &word) {
+  std::string newWord{};
+  std::string punctuation{".,?'\"!():"};
```
ghost: This is a compile time constant, so I'd have this variable in captials too (and make it `const`!)

ghost (changes requested):
This is looking excellent to me now, just a few more small comments. 

The Jenkins builds seem to be failing, the build script is looking for the `template` folder which you've understandably deleted. Putting the `newstarter/exercises/template` folder back should fix this. Then, if the build succeeds, you should get some nice green ticks instead of red crosses.

exercises/Phil_Colebrooke/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,25 @@
...
+
+#define _USE_MATH_DEFINES
+
+#include "Shape.h"
+
+#include <cmath>
+
+class Circle : public Shape {
+private:
+  const double m_radius;
```
alicerussell1, line 10: Well done for using `const` and prefixing with `m_`

exercises/Phil_Colebrooke/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,27 @@
...
+  const int m_sidesNumber;
+
+public:
+  /// Constructor
+  Shape(std::string type, int sidesNumber);
+
+  /// Pure abstract functions
+  /// Shape cannot be instansiated
+  virtual double getPerimeter() = 0;
+  virtual double getArea() = 0;
```
alicerussell1: Good use of purely virtual functions, these could also be made `const` since the shape should not change

exercises/Phil_Colebrooke/ex02_oo_basics/src/main.h
```diff
@@ -0,0 +1 @@
+#pragma once
```
alicerussell1: If the file is not needed you do not need to create it.

exercises/Phil_Colebrooke/ex02_oo_basics/src/Shape.h
alicerussell1: There are a number of methods here that could be `const`, I'll leave you to go through them all.

exercises/Phil_Colebrooke/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @param sides The number of sides
+ *
+ */
+void ShapeSorter::printBySides(int sides) {
+  for (Shape *shape : m_shapes) {
+    if (shape->getSides() == sides) {
+      std::cout << shape->getType() + " with perimeter "
+                << shape->getPerimeter() << " and area " << shape->getArea()
+                << "\n";
```
alicerussell1: To reduce the number of lines of code and so you don't keep repeating this, you could consider have a print method that is called by each of these methods

exercises/Phil_Colebrooke/ex02_oo_basics/src/main.h
alicerussell1: Just so you are aware, it is worth noting that `#pragma once` works on visual studio but not guaranteed to work everywhere which is why we use 
```
#ifndef MAIN_H
#define MAIN_H

#endif // MAIN_H
```

alicerussell1 (changes requested):
This is a good approach to exercises 2 - well done. I've put a few comments below

exercises/Phil_Colebrooke/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,25 @@
+#ifndef CIRCLE_H
+#define CIRCLE_H
+
+#define _USE_MATH_DEFINES
```
alicerussell1: This is giving warnings and I think it can be removed

alicerussell1 (changes requested):
This is looking great to me, I've just added one more comment so it will pass the tests

alicerussell1:
Exercises 2 are now looking good to me 👍

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -18,16 +12,18 @@ int main(int argc, char *argv[]) {
...
   }
 
-  inputFilename = argv[1];
-  outputFilename = argv[2];
+  std::string inputFilename{argv[1]};
+  std::string outputFilename{argv[2]};
 
   // the inputted min word length is converted to int
-  minWordLength = atoi(argv[3]);
+  int minWordLength{atoi(argv[3])};
```
ghost: The three variables above should probably be `const`

exercises/Phil_Colebrooke/ex01_basics/src/main.cpp
```diff
@@ -18,16 +12,18 @@ int main(int argc, char *argv[]) {
...
 
+  bool success{false};
+  std::map<std::string, int> wordsAndCounts;
   success = countWords(inputFilename, wordsAndCounts, minWordLength);
 
   if (success) {
-    sorted = orderByUsage(wordsAndCounts);
+    auto sorted{orderByUsage(wordsAndCounts)};
```
ghost: `sorted` is an OK name, but it could be a little more descriptive. What exactly is sorted here? Container names should ideally describe what they contain

ghost (changes requested):
This looks great to me; a good approach and very well written :+1:. Just a couple of (very picky) comments below.

ghost (approved):
Code all looks good to me 🎉.

My final comment would be that I feel your git commit messages could be a little more specific about the changes made. Don't worry about updating the messages on this project, but for the future, [this article](https://chris.beams.io/posts/git-commit/) was very useful to me when I first started using git.

PR #44, opened 2019-07-12
Conor Finn Exercises

cailafinn:
**Description of Work**
Solution to exercise 01 of the new starter C++ exercises. 

**To test:**
1. Create binaries
2. Run from command line. 
3. When prompted, enter the path for the file to count.
4. When prompted, enter the path for an output file. 
5. Check the output file specified for accuracy. 

2019-07-16:
Completes #45

### DIFF
--- a/exercises/conor_finn/ex01_basics/src/main.cpp
+++ b/exercises/conor_finn/ex01_basics/src/main.cpp
@@ -10,40 +10,6 @@ using std::ifstream;    using std::ofstream;
 using std::map;
 
 /**
- * @brief Determine if a character is an invalid punctuation character.
- * 
- * @param c The character to check.
- * @return true If invalid.
- * @return false if valid or alphanumeric.
- */
-bool isBadPunct(char c) {
-    static const string badPunct = ".,?'\"!():";
-    return !(isalnum(c) || 
-        find(badPunct.begin(), badPunct.end(), c) != badPunct.end());
-}
-
-/**
- * @brief Cleans a word by excluduing invalid punctuation,
- * and changing its case. 
- * 
- * @param s The string to clean. 
- */
-void cleanWord(string& s) {
-    s.erase(std::remove_if(s.begin(), s.end(), isBadPunct));
-}
-
-/**
- * @brief Counts the number of occurances of >4 character words 
- * in a file stream. 
- * 
- * @param in The input stream.
- * @return map<string, int> A map of the word and number of occurances. 
- */
-map<string, int> countWords(ifstream& in) {
-    
-}
-
-/**
  * @brief Creates a results file containing the number of
  * counts of words of >4 characters in a user defined file.
  * 
@@ -65,6 +31,61 @@ int wordCounterInterface() {
             counts = countWords(filein);
     } else {
         cout << "Invalid path or blank file." << endl;
+    }
+}
+
+/**
+ * @brief Determine if a character is an invalid punctuation character.
+ * 
+ * @param c The character to check.
+ * @return true If invalid.
+ * @return false if valid or alphanumeric.
+ */
+bool isBadPunct(char c) {
+    static const string badPunct = ".,?'\"!():";
+    return !(isalnum(c) || 
+        find(badPunct.begin(), badPunct.end(), c) == badPunct.end());
+}
+
+/**
+ * @brief Convert a string to be all lower case. 
+ * 
+ * @param s The string to convert. 
+ */
+void toLowerCase(string& s) {
+    for(string::size_type i = 0; i < s.size(); ++i) {
+        s[i] = tolower(s[i]);
+    }
+}
+
+/**
+ * @brief Cleans a word by excluduing invalid punctuation,
+ * and changing its case. 
+ * 
+ * @param s The string to clean. 
+ */
+void cleanWord(string& s) {
+    s.erase(std::remove_if(s.begin(), s.end(), isBadPunct));
+    toLowerCase(s);
+}
+
+/**
+ * @brief Counts the number of occurances of >4 character words 
+ * in a file stream. 
+ * 
+ * @param in The input stream.
+ * @return map<string, int> A map of the word and number of occurances. 
+ */
+map<string, int> countWords(ifstream& in) {
+    map<string, int> ret;
+    string s;
+    // Read strings, using whitespace as a separator. 
+    while(in >> s) {
+        cleanWord(s);
+        // Check that clean string is >4 characters. 
+        if(s.size > 4) {
+            ++ret[s];
+        }
     }
 }
 
--- a/exercises/conor_finn/ex01_basics/src/main.cpp
+++ b/exercises/conor_finn/ex01_basics/src/main.cpp
@@ -9,7 +9,7 @@
  * @return int 0 on success, >0 otherwise.
  */
 int wordCounterInterface() {
-  const int MIN_WORD_LENGTH = 5;
+  const int MIN_WORD_LENGTH{5};
   // Ask for a file path.
   std::cout << "Please enter the file path:";
   std::string path;
@@ -21,34 +21,44 @@ int wordCounterInterface() {
   // Check if the file exists.
   if (fileIn.is_open()) {
     std::map<std::string, int> counts = countWords(fileIn, MIN_WORD_LENGTH);
-    std::vector<std::pair<int, std::string>> sortedByWordCount{sortMap(counts)};
+    auto sortedByWordCount{sortMap(counts)};
     fileIn.close();
-    std::cout
-        << "Please enter the output path (paths that do not reach existing "
-           "files will create new ones):";
-    std::cin >> path;
-    std::ofstream fileOut{path.c_str()};
-    if (fileOut.is_open()) {
-      // Write headings to the file.
-      std::string::size_type maxWordLength =
-          findMaxStringSize(sortedByWordCount);
-      fileOut << "Word" + std::string(maxWordLength - 4, ' ') + "\tCounts"
-              << std::endl;
-      // Write the list of words to the file.
-      for (auto &wordCountPair : sortedByWordCount) {
-        fileOut << wordCountPair.second +
-                       std::string(
-                           maxWordLength - (wordCountPair.second.size()), ' ') +
-                       "\t"
-                << wordCountPair.first << std::endl;
-      }
-      fileOut.close();
-      return 0;
-    } else {
-      std::cout << "Invalid output path." << std::endl;
-    }
+    return writeToFile(sortedByWordCount);
   } else {
     std::cout << "Invalid input path or blank file." << std::endl;
+    return 1;
+  }
+}
+
+/**
+ * @brief Format and write a vector of pairs to a file supplied by the user.
+ *
+ * @param words Vector to write.
+ * @return int 0 on success, 1 otherwise.
+ */
+int writeToFile(std::vector<std::pair<int, std::string>> words) {
+  std::cout << "Please enter the output path (paths that do not reach existing "
+               "files will create new ones):";
+  std::string path;
+  std::cin >> path;
+  std::ofstream fileOut{path.c_str()};
+  if (fileOut.is_open()) {
+    // Write headings to the file.
+    std::string::size_type maxWordLength = findMaxStringSize(words);
+    fileOut << "Word" + std::string(maxWordLength - 4, ' ') + "\tCounts"
+            << std::endl;
+    // Write the list of words to the file.
+    for (auto &wordCountPair : words) {
+      fileOut << wordCountPair.second +
+                     std::string(maxWordLength - (wordCountPair.second.size()),
+                                 ' ') +
+                     "\t"
+              << wordCountPair.first << std::endl;
+    }
+    fileOut.close();
+    return 0;
+  } else {
+    std::cout << "Invalid output path." << std::endl;
     return 1;
   }
 }
@@ -62,9 +72,9 @@ int wordCounterInterface() {
 std::string::size_type
 findMaxStringSize(const std::vector<std::pair<int, std::string>> &pairs) {
   std::string::size_type max = 0;
-  for (const std::pair<int, std::string> &p : pairs) {
-    if (p.second.size() > max) {
-      max = p.second.size();
+  for (const auto &pair : pairs) {
+    if (pair.second.size() > max) {
+      max = pair.second.size();
     }
   }
   return max;
@@ -81,11 +91,11 @@ findMaxStringSize(const std::vector<std::pair<int, std::string>> &pairs) {
 std::map<std::string, int> countWords(std::ifstream &in,
                                       const int &minWordLength) {
   std::map<std::string, int> words;
-  std::string s;
-  while (in >> s) {
-    cleanWord(s);
-    if (s.size() >= minWordLength) {
-      ++words[s];
+  std::string word;
+  while (in >> word) {
+    cleanWord(word);
+    if (word.size() >= minWordLength) {
+      ++words[word];
     }
   }
   return words;
@@ -113,11 +123,11 @@ bool sortInRev(const std::pair<int, std::string> &a,
 std::vector<std::pair<int, std::string>>
 sortMap(const std::map<std::string, int> &wordMap) {
   std::vector<std::pair<int, std::string>> pairs;
-  std::pair<int, std::string> p;
-  for (auto i = wordMap.begin(); i != wordMap.end(); ++i) {
-    p.first = i->second;
-    p.second = i->first;
-    pairs.push_back(p);
+  std::pair<int, std::string> pair;
+  for (auto const word : wordMap) {
+    pair.first = word.second;
+    pair.second = word.first;
+    pairs.push_back(pair);
   }
   std::sort(pairs.begin(), pairs.end(), sortInRev);
   return pairs;
@@ -130,10 +140,10 @@ sortMap(const std::map<std::string, int> &wordMap) {
  * @return true If invalid.
  * @return false if valid or alphanumeric.
  */
-bool isBadPunct(char &c) {
+bool isBadPunct(char &character) {
   static const std::string badPunct = ".,?'\"!():";
-  return !(isalnum(c) ||
-           std::find(badPunct.begin(), badPunct.end(), c) == badPunct.end());
+  return !(isalnum(character) || std::find(badPunct.begin(), badPunct.end(),
+                                           character) == badPunct.end());
 }
 
 /**
@@ -141,9 +151,9 @@ bool isBadPunct(char &c) {
  *
  * @param s The string to convert.
  */
-void toLowerCase(std::string &s) {
-  for (std::string::size_type i = 0; i < s.size(); ++i) {
-    s[i] = tolower(s[i]);
+void toLowerCase(std::string &str) {
+  for (std::string::size_type i = 0; i < str.size(); ++i) {
+    str[i] = tolower(str[i]);
   }
 }
 
@@ -153,9 +163,9 @@ void toLowerCase(std::string &s) {
  *
  * @param s The string to clean.
  */
-void cleanWord(std::string &s) {
-  s.erase(std::remove_if(s.begin(), s.end(), isBadPunct), s.end());
-  toLowerCase(s);
+void cleanWord(std::string &word) {
+  word.erase(std::remove_if(word.begin(), word.end(), isBadPunct), word.end());
+  toLowerCase(word);
 }
 
 /**
--- a/exercises/conor_finn/ex01_basics/src/main.cpp
+++ b/exercises/conor_finn/ex01_basics/src/main.cpp
@@ -1,7 +1,79 @@
 /**
  * Skeleton main routine
  */
+#include "main.h"
+#include <algorithm>
 
-int main(int, char **)
-{
+using std::cin;         using std::cout;
+using std::string;      using std::endl;
+using std::ifstream;    using std::ofstream;
+using std::map;
+
+/**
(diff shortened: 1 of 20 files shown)


exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+using std::cin;
+using std::cout;
+using std::endl;
+using std::ifstream;
+using std::map;
+using std::ofstream;
+using std::pair;
+using std::string;
+using std::vector;
```
ghost: This is fine in small projects, but in Mantid we prefer to keep our namespace clean and don't use `using` on anything in the `std` namespace. Whenever we need a vector, we just write `std::vector`

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * @brief Creates a results file containing the number of
+ * counts of words of >4 characters in a user defined file.
+ *
+ * @return int 0 on success, >0 otherwise.
+ */
+int wordCounterInterface() {
+  // Number of characters a word must have to be counted.
+  const int characters = 5;
```
ghost: In Mantid constants like this are written in all capitals with underscores separating words, see our [naming conventions](http://developer.mantidproject.org/Standards/CPPStandards.html#naming-conventions).

This variable's name could be more descriptive; maybe something like `MIN_WORD_LENGTH`. This is more self documenting and the comment becomes unnecessary.

Note you can also initialize variables using curly braces: `const int MIN_WORD_LENGTH {5};`. Using curly braces is more modern and preferred (unless you're using `auto`) since it stops you [assigning things to types you shouldn't](https://stackoverflow.com/questions/18222926/why-is-list-initialization-using-curly-braces-better-than-the-alternatives).

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+  // Ask for a file path.
+  cout << "Please enter the file path:";
+  string path;
+  cin >> path;
+  ifstream filein;
+  filein.open(path.c_str());
+  // Create a file in the calling directory.
+  ofstream fileout;
```
ghost: I would declare this variable a bit later, only when you know you'll need it. If you've failed to open `filein` later you've created this object for no reason.

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  map<string, int> counts; // Map to hold the counts of each word.
+
+  // Check if the file exists.
+  if (filein.is_open()) {
+    map<string, int> counts = countWords(filein, characters);
+    vector<pair<int, string>> sorted{sortMap(counts)};
+    cout << "Please enter the output path (paths that do not reach existing "
```
ghost: You're done with `filein` at this point, so I would call `filein.close()` here.

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+  filein.open(path.c_str());
+  // Create a file in the calling directory.
+  ofstream fileout;
+
+  map<string, int> counts; // Map to hold the counts of each word.
+
+  // Check if the file exists.
+  if (filein.is_open()) {
+    map<string, int> counts = countWords(filein, characters);
+    vector<pair<int, string>> sorted{sortMap(counts)};
```
ghost: `sorted` could have a more descriptive name here; sorted what? Maybe `sortedWordCounts`?

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+ */
+int wordCounterInterface() {
+  // Number of characters a word must have to be counted.
+  const int characters = 5;
+  // Ask for a file path.
+  cout << "Please enter the file path:";
+  string path;
+  cin >> path;
+  ifstream filein;
+  filein.open(path.c_str());
```
ghost: An alternative to the above two lines is to open the file in the initializer, i.e.: `ifstream fileIn {path.c_str()};`. Also `filein`'s second "I" should be capitalized

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+    vector<pair<int, string>> sorted{sortMap(counts)};
+    cout << "Please enter the output path (paths that do not reach existing "
+            "files will create new ones):";
+    cin >> path;
+    fileout.open(path.c_str());
+    // Write headings to the file.
+    string::size_type maxLength = findMaxSize(sorted);
+    fileout << "Word" + string(maxLength - 4, ' ') + "\tCounts" << endl;
+    // Write the list of words to the file.
+    for (auto it = sorted.begin(); it != sorted.end(); ++it) {
```
ghost: This is good, but a range based `for` loop would be the preferred way to do this. Something like:
```C++
for (auto &wordCountPair : sorted) { fileOut << wordCountPair.second ... }
```

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+  if (filein.is_open()) {
+    map<string, int> counts = countWords(filein, characters);
+    vector<pair<int, string>> sorted{sortMap(counts)};
+    cout << "Please enter the output path (paths that do not reach existing "
+            "files will create new ones):";
+    cin >> path;
+    fileout.open(path.c_str());
+    // Write headings to the file.
+    string::size_type maxLength = findMaxSize(sorted);
+    fileout << "Word" + string(maxLength - 4, ' ') + "\tCounts" << endl;
```
ghost: We haven't checked if `fileout` is open before trying to write to it. If I enter an invalid file name, the program exits without error but I have no file!

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * @brief Finds the maximum length string in vector of pairs.
+ *
+ * @param v The vector to check.
+ * @return string::size_type The max size.
+ */
+string::size_type findMaxSize(const vector<pair<int, string>> v) {
+  string::size_type max = 0;
+  for (pair<int, string> p : v) {
```
ghost: Looping by reference is better here as you avoid making copies: `for (pair<int, string> &p : v)`

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+  }
+}
+
+/**
+ * @brief Finds the maximum length string in vector of pairs.
+ *
+ * @param v The vector to check.
+ * @return string::size_type The max size.
+ */
+string::size_type findMaxSize(const vector<pair<int, string>> v) {
```
ghost: Passing by reference here would be more memory efficient, especially since `v` is constant so we couldn't accidentally change it anyway.

Also, try to avoid single character variable names, even for small functions like this. The only exception being integer loop counters where using `i`,`j`,`k`, etc. is fine.

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/**
+ * @brief Count the number of words with a number characters.
+ * Excludes punctuation and ignores case.
+ *
+ * @param in Input file stream of words.
+ * @return map<string, int> Map of words and counts.
+ */
+map<string, int> countWords(ifstream &in, int chars) {
```
ghost: `maxWordLength` might be a better variable name than `chars` in this case. The argument is also missing in your doc-string

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * @brief Sorting for pairs based on the first value.
+ *
+ * @param a First pair to compare.
+ * @param b Second pair to compare.
+ * @return true First element of a > the first element of b
+ * @return false First elemet of a <= first element of b
+ */
+bool sortInRev(const pair<int, string> &a, const pair<int, string> &b) {
+  return (a.first > b.first);
```
ghost, line 114: Nice!

exercises/conor_finn/ex02_oo_basics/src/circle.h
```diff
@@ -0,0 +1,16 @@
...
+#include "shape.h"
+
+class Circle : public Shape {
+private:
+  int radius;
+
+public:
+  Circle(int radius);
+  double perimiter();
+  double area();
```
alicerussell1: In C++ you should write `override` after methods like this to indicate you are overriding methods in shape. If override is written the compiler can check if any parameters are incorrect and doesn't override as intended.
This should be done for all the overridden methods

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/**
+ * @brief Cleans a word by excluduing invalid punctuation,
+ * and changing its case.
+ *
+ * @param s The string to clean.
+ */
+void cleanWord(string &s) {
+  s.erase(std::remove_if(s.begin(), s.end(), isBadPunct), s.end());
```
ghost: Nice use of `erase` :+1:

exercises/conor_finn/ex01_basics/src/main.h
```diff
@@ -0,0 +1,26 @@
...
+
+int wordCounterInterface();
+
+int main(int, char **);
+
+void cleanWord(std::string &s);
+
+void toLowerCase(std::string &s);
+
+std::vector<std::pair<int, std::string>> sortMap(std::map<std::string, int>);
```
ghost: [Always specify the variable name as well as the type in function prototypes](http://developer.mantidproject.org/Standards/CPPStandards.html#functions-and-variables). With any luck the variable will have a descriptive name and will help self-document the function.

ghost (changes requested):
### Note, the following reviews Exercise 1 only

This is a good approach to the problem and you have excellent doc-strings! :tada:  Good stuff.

I've left some comments below. 

I'd have a good look through your functions and make sure you're passing arguments by reference where you can. Unless you mean to make a copy of whatever is being passed to the function, you should pass by reference. This is much more memory efficient, especially when dealing with larger objects. 

Also, your variable and function names could be a bit more descriptive. Names like `maxLength` can be a bit vague (length of what?), `maxWordLength` is only four extra characters but makes it much clearer what the variable is.

exercises/conor_finn/ex02_oo_basics/src/shapeSorter.cpp
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * @brief Print out the details for all shapes with a given name.
+ *
+ * @param shapes Vector containing pointers to the shapes.
+ * @param n The name to match.
+ */
+void ShapeSorter::printMatchingName(const std::vector<Shape *> &shapes,
+                                    const std::string &n) {
+  int counter = 1;
+  for (auto i = shapes.begin(); i != shapes.end(); i++) {
```
alicerussell1: I don't think these are in the book as it is a fairly recent addition but loops like this can be written as more concisely using a range-base for loop. So it would be:
`for (const auto &i : shapes) {
... }`

exercises/conor_finn/ex02_oo_basics/src/circle.h
```diff
@@ -0,0 +1,16 @@
+#ifndef CIRCLE_H_
+#define CIRCLE_H_
+
+#include "shape.h"
+
+class Circle : public Shape {
+private:
+  int radius;
```
alicerussell1: I think this could be `const`. It is better to make everything `const` if it can be

exercises/conor_finn/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,53 @@
...
+  shapes.push_back(t2);
+  shapes.push_back(s1);
+  shapes.push_back(c1);
+
+  // Printing squares.
+  std::cout << "Squares: \n\n";
+  ShapeSorter::printMatchingName(shapes, "Square");
+
+  // Printing triangles.
+  std::cout << std::endl << "Triangles: \n\n";
```
alicerussell1: It is better for performance to use the newline character `\n` rather than `std::endl` so it is best to avoid `endl` when possible.
This is because `std::endl` flushes the output buffer whereas `\n` simply outputs a newline.

exercises/conor_finn/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,23 @@
...
+class Shape {
+private:
+  int sides;
+  std::string name;
+
+public:
+  Shape(int s, std::string n);
+
+  // Pure virtual functions. Overidden.
+  virtual double perimiter() = 0;
```
alicerussell1: These member functions should not mutate shape, so they can be made const so this would be 
`virtual double perimeter() const = 0;`

exercises/conor_finn/ex02_oo_basics/src/shape.cpp
```diff
@@ -0,0 +1,37 @@
...
+  sides = s;
+  name = n;
+}
+
+/**
+ * @brief Accessor method for number of sides.
+ *
+ * @return int Number of sides.
+ */
+int Shape::getSides() { return sides; }
```
alicerussell1: I think it would be clearer if the name of this was something like `getNumberOfSides`

exercises/conor_finn/ex02_oo_basics/src/shapeSorter.cpp
```diff
@@ -0,0 +1,72 @@
...
+ * @param shapes A vector of shape pointers.
+ */
+void ShapeSorter::printOrderedByPerimiter(std::vector<Shape *> shapes) {
+  std::sort(shapes.begin(), shapes.end(),
+            [](Shape *a, Shape *b) { return a->perimiter() > b->perimiter(); });
+  int counter = 1;
+  for (auto i = shapes.begin(); i != shapes.end(); i++) {
+    std::cout << "Shape " << counter << ":\n";
+    (*i)->printInfo();
+    ++counter;
```
alicerussell1, line 71: Rather than repeating this code 4 times, it would be better if you separated it out and then called it for each of these methods

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -2,58 +2,53 @@
...
 /**
  * @brief Creates a results file containing the number of
  * counts of words of >4 characters in a user defined file.
  *
  * @return int 0 on success, >0 otherwise.
  */
 int wordCounterInterface() {
-  // Number of characters a word must have to be counted.
-  const int characters = 5;
+  const int MIN_WORD_LENGTH = 5;
```
ghost: I know this is a bit picky, but using curly braces instead of equals is preferred here: `const int MIN_WORD_LENGTH {5};`. You will see `=` used a lot in Mantid, but curly braces is better practice.

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -64,9 +59,10 @@ int wordCounterInterface() {
  * @param v The vector to check.
  * @return string::size_type The max size.
  */
-string::size_type findMaxSize(const vector<pair<int, string>> v) {
-  string::size_type max = 0;
-  for (pair<int, string> p : v) {
+std::string::size_type
+findMaxStringSize(const std::vector<std::pair<int, std::string>> &pairs) {
+  std::string::size_type max = 0;
+  for (const std::pair<int, std::string> &p : pairs) {
```
ghost: Using `const auto` here would neaten up this `for` loop

exercises/conor_finn/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,53 @@
...
+int main(int, char **) {
+  std::vector<Shape *> shapes;
+
+  // Make some shapes.
+  Rectangle *r1 = new Rectangle(5, 10);
+  Rectangle *r2 = new Rectangle(50, 10);
+  Triangle *t1 = new Triangle(10, 23);
+  Triangle *t2 = new Triangle(2, 60);
+  Square *s1 = new Square(67);
+  Circle *c1 = new Circle(42);
```
alicerussell1, line 19: Using `new` isn't ideal as you then have to delete the pointers at the end (well done for remembering to do this though). You could consider using smart pointers, the standards say to "replace uses of `new` by `std::make_unique`, where possible"

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -75,18 +71,20 @@ string::size_type findMaxSize(const vector<pair<int, string>> v) {
...
+ * @param minWordLength The minimum word length to accept.
+ * @return std::map<std::string, int> Map of strings and number of occurances.
  */
-map<string, int> countWords(ifstream &in, int chars) {
-  map<string, int> words;
-  string s;
+std::map<std::string, int> countWords(std::ifstream &in,
+                                      const int &minWordLength) {
+  std::map<std::string, int> words;
+  std::string s;
```
ghost: I'd call this variable `word`, since that's what you're reading in. 

Have a look through and find any single letter variable names and see if they can be replaced with something more descriptive.

alicerussell1 (changes requested):
This is a good solution to the second exercise and the level of documentation in this is great - well done 👍
I've left a few comments below, a number of them can be applied throughout the code even though I've only written them once.

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -111,9 +110,10 @@ bool sortInRev(const pair<int, string> &a, const pair<int, string> &b) {
...
  * @return vector<pair<int, string> > A sorted vector of pairs.
  */
-vector<pair<int, string>> sortMap(map<string, int> wordMap) {
-  vector<pair<int, string>> pairs;
-  pair<int, string> p;
+std::vector<std::pair<int, std::string>>
+sortMap(const std::map<std::string, int> &wordMap) {
+  std::vector<std::pair<int, std::string>> pairs;
+  std::pair<int, std::string> p;
   for (auto i = wordMap.begin(); i != wordMap.end(); ++i) {
```
ghost: We can use a range based `for` loop here (and `const`)

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -2,58 +2,53 @@
...
-    string::size_type maxLength = findMaxSize(sorted);
-    fileout << "Word" + string(maxLength - 4, ' ') + "\tCounts" << endl;
-    // Write the list of words to the file.
-    for (auto it = sorted.begin(); it != sorted.end(); ++it) {
-      fileout << it->second + string(maxLength - (it->second.size()), ' ') +
-                     "\t"
-              << it->first << endl;
+  if (fileIn.is_open()) {
+    std::map<std::string, int> counts = countWords(fileIn, MIN_WORD_LENGTH);
+    std::vector<std::pair<int, std::string>> sortedByWordCount{sortMap(counts)};
```
ghost: You could `auto`this to avoid writing out the long templated type name

exercises/conor_finn/ex01_basics/src/main.cpp
```diff
@@ -2,58 +2,53 @@
...
+              << std::endl;
+      // Write the list of words to the file.
+      for (auto &wordCountPair : sortedByWordCount) {
+        fileOut << wordCountPair.second +
+                       std::string(
+                           maxWordLength - (wordCountPair.second.size()), ' ') +
+                       "\t"
+                << wordCountPair.first << std::endl;
+      }
+      fileOut.close();
```
ghost: This block of code that's writing to the results file could be refactored into its own function. This has the advantage of self documenting what the block of code does via the (hopefully descriptive) function name and also keeps the `wordCounterInterface` cleaner and easier to read.

Splitting larger functions into smaller ones also makes testing the functions much easier

ghost (commented):
This is looking pretty good. I've left a few more comments below :+1:

exercises/conor_finn/ex02_oo_basics/src/circle.h
```diff
@@ -0,0 +1,16 @@
...
+private:
+  const int radius;
+
+public:
+  Circle(int radius);
+  double perimiter() const override;
+  double area() const override;
+};
+
+#endif
```
alicerussell1: It is good practice to say which if is ending:  `#endif // CIRCLE_H`

exercises/conor_finn/ex02_oo_basics/src/circle.h
```diff
@@ -0,0 +1,16 @@
+#ifndef CIRCLE_H_
+#define CIRCLE_H_
+
+#include "shape.h"
+
+class Circle : public Shape {
+private:
+  const int radius;
```
alicerussell1: It is in the mantid standards to use an m_ prefix for non-static member variables i.e. `m_radius`

alicerussell1 (changes requested):
This is looking good to me, I've just added a couple of comments below

alicerussell1:
The solutions to exercise 2 are now looking good 👍 It is failing the windows test as it cannot compile, I think this is just because you haven't got `#include <string>` in exercises 1.

ghost (approved):
Exercise 1 looks good to me now. Good stuff :+1:

alicerussell1 (approved):
The tests are passing and the code looks good to me 👍

PR #47, opened 2019-07-26
Harriet brown exercises

Harrietbrown:
Exersize 1: code to count instances of words in an ascii file.
file to test: [Holmes.txt](https://raw.githubusercontent.com/martyngigg/cpp-examples/master/Holmes.txt)

Exersize 2: code to create, list, and sort shapes.


exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+void remove_punctuation_in_string(std::string &data)
+{
+	std::vector<std::string> puntuation{ ".", ",", "?", "!", "\'", "\"", "(", ")", ":", "--", ";"};
+	for (auto x : puntuation) {
+		while (data.find(x) != std::string::npos) {
+			data.replace(data.find(x), 1, " ");
+		}
+	}
+	std::transform(data.begin(), data.end(), data.begin(), ::tolower);
+	return;
```
alicerussell1, line 35: This isn't quite working correctly for `"--"`, it is only replacing one of the dashes so the word is preceded by a dash. It might work better if this is treated as a special case in this function i.e. have an extra while loop for it.

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <fstream>
+#include <map>
+#include <sstream>
+#include <string>
+#include <vector>
+#include <algorithm>
+
+void remove_punctuation_in_string(std::string &data)
+{
+	std::vector<std::string> puntuation{ ".", ",", "?", "!", "\'", "\"", "(", ")", ":", "--", ";"};
```
alicerussell1: This can just be written as a string rather than a vector of strings so like:
 `std::string puntunation{".,?'\"!():;"}` 
which is much more readable.

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+	std::ofstream result_file;
+	result_file.open("word_count.txt");
+	std::cout << std::setw(18) << "Word" << std::setw(12) << "Count" << std::endl;
+	result_file << std::setw(18) << "Word" << std::setw(12) << "Count" << std::endl;
+	std::string max_key;
+	int max_val;
+	while (map_to_print.size() > 0) {
+		max_val = 0;
+		for (auto it = map_to_print.begin(); it != map_to_print.end(); ++it) {
```
alicerussell1: This can be written as a range based loop: `for (const auto &it : map_to_print) {`

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+			if (it->second > max_val) {
+				max_val = it->second;
+				max_key = it->first;
+			}
+		}
+		std::cout << std::setw(18) << max_key << std::setw(12) << max_val << std::endl;
+		result_file << std::setw(18) << max_key << std::setw(12) << max_val << std::endl;
+		map_to_print.erase(max_key);
+		max_val = 0;
```
alicerussell1: `max_val` is already set to 0 at the start of the loop so doesn't need to be set at the end as well

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+void print_map_by_value(std::map<std::string, int> map_to_print)
+{
+	std::ofstream result_file;
+	result_file.open("word_count.txt");
+	std::cout << std::setw(18) << "Word" << std::setw(12) << "Count" << std::endl;
+	result_file << std::setw(18) << "Word" << std::setw(12) << "Count" << std::endl;
+	std::string max_key;
+	int max_val;
```
alicerussell1: Since `max_val` and `max_key` are only used inside the while loop it is better to declare it in the loop so it isn't around for any longer than it needs to be

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+	std::ifstream ascii_file;
+	std::string line;
+	std::stringstream str_strm;
+	std::string tmp;
+	std::vector<std::string> words;
+	std::map<std::string, int> word_count_map;
+	// Check the number of parameters
+	if (argc < 2) {
+		std::cerr << argv[0] << " needs to be given a file directory as an argument" << std::endl;
+		return 0;
```
alicerussell1: If a function or program exits having failed then it is convention to return some non-zero error code

alicerussell1: It would also be good if you could check that the number of arguments isn't greater than 2

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+	int max_val;
+	while (map_to_print.size() > 0) {
+		max_val = 0;
+		for (auto it = map_to_print.begin(); it != map_to_print.end(); ++it) {
+			if (it->second > max_val) {
+				max_val = it->second;
+				max_key = it->first;
+			}
+		}
+		std::cout << std::setw(18) << max_key << std::setw(12) << max_val << std::endl;
```
alicerussell1: It is probably not necessary to output this to the command line as well as the file.

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+	std::transform(data.begin(), data.end(), data.begin(), ::tolower);
+	return;
+}
+
+void print_map_by_value(std::map<std::string, int> map_to_print)
+{
+	std::ofstream result_file;
+	result_file.open("word_count.txt");
```
alicerussell1, line 41: It would be good to check if this has actually opened.

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+	while (std::getline(ascii_file, line)) {
+		remove_punctuation_in_string(line);
+		str_strm << line << " ";
+	}
+	ascii_file.close();
+	while (str_strm >> tmp) {
+		words.push_back(tmp);
+	}
+	
+	for (unsigned int i = 0; i < words.size(); i++) {
```
alicerussell1: This could also be a range base loop.

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+		if (words[i].size() > 4) {
+			if (word_count_map.find(words[i]) != word_count_map.end()) {
+				word_count_map[words[i]]++;
+			}
+			else {
+				word_count_map[words[i]] = 1;
+			}
```
alicerussell1: This if else statement isn't actually needed, you only need `word_count_map[words[i]]++;`

alicerussell1 (changes requested):
This is a review of the first exercise. It is a good attempt and it works as intended 👍 

I've added some comments below to improve the way it is written to improve performance and readability. It would also be good practice to include more documentation so it is easier to understand what it all does.

Let me know if you have any questions

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+	std::transform(data.begin(), data.end(), data.begin(), ::tolower);
+	return;
+}
+
+void print_map_by_value(std::map<std::string, int> map_to_print)
+{
+	std::ofstream result_file;
+	result_file.open("word_count.txt");
+	if (!result_file) {
+		std::cerr << "Unable to open results file word_count.txt";
```
martyngigg: In mantid we would use [Exceptions](https://www.tutorialspoint.com/cplusplus/cpp_exceptions_handling) for this and not a hard `exit` as it allows us to possibly handle the error rather than just stopping.

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+int main(int argc, char* argv[])
+{
+	std::ifstream ascii_file;
+	std::string line;
+	std::stringstream str_strm;
+	std::string tmp;
+	std::vector<std::string> words;
+	std::map<std::string, int> word_count_map;
+	// Check the number of parameters
+	if (argc < 2) {
```
martyngigg: Good sanity checking of arguments. :+1:

exercises/harriet_brown/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+		std::cerr << argv[0] << " needs to be given a file directory as an argument" << std::endl;
+		exit(1);   // call system to stop
```
martyngigg: Again we would generally use exceptions to indicate any type of error.

exercises/harriet_brown/ex02_oo_basics/inc/Circle.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
martyngigg: This is okay for small projects and unfortunately the default on visual studio but in Mantid we still use [include guards](https://www.learncpp.com/cpp-tutorial/header-guards/) for maximum portability.

exercises/harriet_brown/ex02_oo_basics/inc/Shape.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+#include <string>
+#include <math.h>
+#include <iomanip>
+#include <iostream>
+
+class Shape
+{
+	private:
+		std::string shape;
```
martyngigg: We try to follow a set of standards in Mantid about how various types of variables are [named](http://developer.mantidproject.org/Standards/CPPStandards.html#naming-conventions). This helps with understanding the scope of variables when looking at their usage.

exercises/harriet_brown/ex02_oo_basics/inc/Shape.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+#include <string>
```
martyngigg, line 4: One of the most important things for us in C++ is to keep all unnecessary includes out of a file as these cause big increases in compile time as other files include headers. This is the "include what you use" rule. 

In particular `<iostream>` should never be used in a header as it's very heavy.

exercises/harriet_brown/ex02_oo_basics/src/getInput.cpp
```diff
@@ -0,0 +1,62 @@
+#include "../inc/getInput.h"
+
+
+void getInput(std::string question_text, int &ans)
```
martyngigg: Good use of functions to break this up. In most cases I would opt for returning a value rather than using an output parameter so something like `int getInput(const std::string & question_text)`. It makes reading the code easier and doesn't require you to define a variable first to capture it when calling the function.

exercises/harriet_brown/ex02_oo_basics/src/getInput.cpp
```diff
@@ -0,0 +1,62 @@
...
+			std::cout << "You have entered wrong input" << std::endl;
+			std::cin >> ans;
+		}
+		if (!std::cin.fail())
+			break;
+	}
+	return;
+}
+
+void getInput(std::string question_text, double &ans)
```
martyngigg: This could be done with [templates](https://www.geeksforgeeks.org/templates-cpp/) to avoid duplicating the code:

```
template<typename ResultType>
ResultType getInput(const std::string & question_text) {
       ResultType ans;
       std::cout << question_text;
       std::cin >> ans;
       std::cout << std::endl;
	while (1)
	{
		if (std::cin.fail())
		{
			std::cin.clear();
			std::cin.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
			std::cout << "You have entered wrong input" << std::endl;
			std::cin >> ans;
		}
		if (!std::cin.fail())
			break;
	}
	return ans;
}

File: exercises/harriet_brown/ex02_oo_basics/src/main.cpp [hunk 19]
```diff
@@ -0,0 +1,69 @@
...
+
+void setup_new_shape(std::vector<Shape*> &shapes)
+{
+	std::string shape_type;
+	double side1;
+	double side2;
+	getInput("Shape: ", shape_type);
+	if (shape_type == "square") {
+		getInput("Side size: ", side1);
+		shapes.push_back(new Square(side1));
```
martyngigg: These days we would try to use a [`std::unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr) to avoid having to call an explicit `new` as this avoids a memory leak. So here we would have `std::vector<std::unique_ptr<Shape>>` and then `shapes.push_back(std::make_unique<Square>(side1))` for example.

exercises/harriet_brown/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,69 @@
...
+		getInput("Height size: ", side2);
+		shapes.push_back(new Triangle(side1, side2));
+	}
+	else if (shape_type == "circle") {
+		getInput("Radius size: ", side1);
+		shapes.push_back(new Circle(side1));
+	}
+}
+
+void print_option_text()
```
martyngigg, line 37: :+1:

martyngigg (approved):
Alice is away so I've added a few more comments.

This generally looks good and most things I've commented out are things that the intro books don't go over but are things that are important to consider in mantid.

martyngigg (approved):

martyngigg:
I'm going to close this now (I'll leave the branch around) as I assume they are no longer being worked on.

PR #48, opened 2019-07-31
Exercises

ciaranightingale:
Solution to exercise 1. 
run WordCounter and enter file.txt as provided below when prompted. should produce a file called text.txt which contains the solution.
[file.txt](https://github.com/mantidproject/newstarter/files/3450630/file.txt)

Solution to exercise 2
run shapes.exe and it should print the ShapeSorter functionality to the terminal

### DIFF
--- a/exercises/ciara_nightingale/ex01_basics/src/main.cpp
+++ b/exercises/ciara_nightingale/ex01_basics/src/main.cpp
@@ -19,7 +19,7 @@ bool replaceHyphens(std::string &word) {
 }
 
 /// Determines if a character is punctuation
-bool isPunc(char &const c) {
+bool isPunc(char const &c) {
   static const std::string PUNC = ".,?\'\"!();:";
   for (std::string::size_type i = 0; i != PUNC.size(); ++i) {
     if (c == PUNC[i]) {
--- a/exercises/ciara_nightingale/ex01_basics/src/main.cpp
+++ b/exercises/ciara_nightingale/ex01_basics/src/main.cpp
@@ -8,10 +8,10 @@
 #include <vector>
 #include <sstream>
 
-// replaces the hypens with whitespace
+// replaces the hypens with whitespace - used in clean_word
 bool replace_hyphens(std::string& word) {
-  // std::cout << path[6] << std::endl;
   for (std::string::size_type i = 0; i != word.size(); ++i) {
+		//check through each character in the word
     if (word[i] == '-') {
       word[i] = ' ';
       return true;
@@ -20,7 +20,7 @@ bool replace_hyphens(std::string& word) {
   return false;
 }
 
-// determine if a character is punctuation
+// determine if a character is punctuation - used in clean_word
 bool is_punc(char const c) {
   static const std::string punc = ".,?\'\"!();";
   for (std::string::size_type i = 0; i != punc.size(); ++i) {
@@ -32,7 +32,7 @@ bool is_punc(char const c) {
   return false;
 }
 
-// deal with punctuation and case sensitivity
+// deal with punctuation and case sensitivity - used in process_word
 void clean_word(std::string& word) {
   // remove characters if they are puctuation
   word.erase(remove_if(word.begin(), word.end(), is_punc), word.end());
@@ -40,22 +40,32 @@ void clean_word(std::string& word) {
   transform(word.begin(), word.end(), word.begin(), tolower);
 }
 
-// convert a map to a vector of pairs
-std::vector<std::pair<std::string, double>>
-mapToVector(const std::map<std::string, double> &map) {
-  return std::vector<std::pair<std::string, double>>(map.begin(), map.end());
+//clean the words, count them and put them into the map
+void processWord(std::string& word, std::map<std::string, int>& counter, std::string::size_type& maxlen, std::string::size_type const min_word_length) {
+		clean_word(word);
+		maxlen = std::max(maxlen, word.size()); //find the longest word.
+		if (word.size() > min_word_length) {
+				++counter[word];
+				// add the word to the map and the no. of appearances in the text.
+		}
 }
 
-// determine which pair is larger
-bool compare(const std::pair<std::string, double> pair_1,
-             const std::pair<std::string, double> pair_2) {
+// convert a map to a vector of pairs
+std::vector<std::pair<std::string, int>>
+mapToVector(const std::map<std::string, int> &map) {
+  return std::vector<std::pair<std::string, int>>(map.begin(), map.end());
+}
+
+// determine which pair is greater
+bool compare(const std::pair<std::string, int> pair_1,
+             const std::pair<std::string, int> pair_2) {
   return pair_1.second > pair_2.second;
 }
 
-void Vector_to_File(std::vector<std::pair<std::string, double>> counter,
+void Vector_to_File(std::vector<std::pair<std::string, int>> counter,
                     std::string::size_type pad) {
   std::ofstream Results("text.txt");
-  // writing the titles of the columns to the file
+  // write the titles of the columns to the file
   Results << "Word" << std::string(pad - 3, ' ') << "Usage"
           << "\n\n";
   // write the results to the file, padded to make room for the longest word
@@ -66,47 +76,41 @@ void Vector_to_File(std::vector<std::pair<std::string, double>> counter,
   }
 }
 
-// std::ifstream Load_file()
-
-void process_word(std::string& word, std::map<std::string, double>& counter, std::string::size_type& maxlen, std::string::size_type const word_length) {
-		clean_word(word);
-		maxlen = std::max(maxlen, word.size());
-		if (word.size() > word_length) {
-				++counter[word];
-		}
-}
-
-void Unique(std::string::size_type const word_length) {
-  std::map<std::string, double> counter;
+//counts the number of unique words no. of instances
+void uniqueWordCounter(std::string::size_type const min_word_length) {
+  std::map<std::string, int> counter;
   std::string path;
-  std::cout << "enter file name: " << std::endl;
-  std::cin >> path;
-  std::ifstream in(path);
+	std::string::size_type maxlen = 0;
+	std::string word;
+	std::cout << "enter file name: ";
+	std::cin >> path;
+	std::ifstream in(path);
 	if (!in) {
-		std::cerr << "failed to load file! Incorrect file name specified";
+			std::cerr << "failed to load file! Incorrect file name specified";
 	}
-  std::string::size_type maxlen = 0;
-  std::string word;
+  
   while (in >> word) {
     if (replace_hyphens(word)) {
+				//if the word contains a hyphen, remove the hyphen, split into two words.
 				std::stringstream stream_word(word);
 				std::string word_1, word_2;
 				stream_word >> word_1;
 				stream_word >> word_2;
-				process_word(word_1, counter, maxlen, word_length);
-				process_word(word_2, counter, maxlen, word_length);
+				processWord(word_1, counter, maxlen, min_word_length);
+				processWord(word_2, counter, maxlen, min_word_length);
     }
 		else {
-				process_word(word, counter, maxlen, word_length);
+				processWord(word, counter, maxlen, min_word_length);
 		}
   }
-  std::vector<std::pair<std::string, double>> count;
+	//vector containing the word and usage pairs from the map.
+  std::vector<std::pair<std::string, int>> count;
   count = mapToVector(counter);
-  std::sort(count.begin(), count.end(), compare);
-  Vector_to_File(count, maxlen);
+  std::sort(count.begin(), count.end(), compare); // reorder the vector in descending order
+  Vector_to_File(count, maxlen); // write the vector to a file
 }
 
 int main() {
-  Unique(4);
+  uniqueWordCounter(4);
   return 0;
 }
--- a/exercises/ciara_nightingale/ex01_basics/src/main.cpp
+++ b/exercises/ciara_nightingale/ex01_basics/src/main.cpp
@@ -81,13 +81,14 @@ void vectortoFile(std::vector<std::pair<std::string, int>> const counter,
 }
 
 /// Caluclates the number of instances of unique words and prints it into a file
-void uniqueWordCounter(std::string::size_type const minWordLength) {
+int uniqueWordCounter(std::string::size_type const minWordLength) {
   std::string path;
   std::cout << "enter file name: ";
   std::cin >> path;
   std::ifstream in(path);
   if (!in) {
     std::cerr << "failed to load file! Incorrect file name specified";
+		return 1;
   }
   std::string word;
   std::map<std::string, int> counter;
@@ -110,6 +111,7 @@ void uniqueWordCounter(std::string::size_type const minWordLength) {
   std::sort(count.begin(), count.end(), pairValueIsGreater);
   // Reorder the vector in descending order
   vectortoFile(count, maxlen); // Write the vector to a file
+	return 0;
 }
 
 int main() {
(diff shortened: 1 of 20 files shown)


exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include <sstream>
+
+// replaces the hypens with whitespace - used in clean_word
+bool replace_hyphens(std::string& word) {
+	bool test = false; //so the whole word is checked rather than stopping at the first instance of a hyphen
+	for (std::string::size_type i = 0; i != word.size(); ++i) {
+		//check through each character in the word
+		if (word[i] == '-') {
+				word[i] = ' ';
+				test = true;
```
ghost: These indents look a bit odd on here, I think this is because they are tabs and not spaces. 

I'm not sure which IDE you're using, but there should be a setting somewhere for it to automatically convert tabs to spaces. This makes sure the formatting is the same in whatever program you're looking at the code in (annoyingly, some programs interpret tabs as different lengths than others).

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include <iostream>
+#include <map>
+#include <ostream>
+#include <string>
+#include <vector>
+#include <sstream>
+
+// replaces the hypens with whitespace - used in clean_word
+bool replace_hyphens(std::string& word) {
+	bool test = false; //so the whole word is checked rather than stopping at the first instance of a hyphen
```
ghost: `test` is a slightly vague variable name. Perhaps a more descriptive name would be something like `hyphenReplaced`, this makes the purpose of the variable much clearer

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+	for (std::string::size_type i = 0; i != word.size(); ++i) {
+		//check through each character in the word
+		if (word[i] == '-') {
+				word[i] = ' ';
+				test = true;
+		}
+  }
+	if (test == true) {
```
ghost: This could be shortened to `if(test)`, or even better, remove the `if` statement and just return `test`

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+			return true;
+	}
+  return false;
+}
+
+// determine if a character is punctuation - used in clean_word
+bool is_punc(char const c) {
```
ghost: You can pass by reference here, especially since you're passing it as `const`. As a rule, if you're not intending to make a separate copy of whatever you're passing, pass it by reference

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+  static const std::string punc = ".,?\'\"!();:";
```
ghost: Nice use of `static` 👍

Since `punc` is a static constant, I'd make it all capitals

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+  for (std::string::size_type i = 0; i != punc.size(); ++i) {
+    if (c == punc[i]) {
+      return true;
+      break;
```
ghost, line 27: You don't need a `break` here; you've already returned out of the function so line 33 will never be reached

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+void uniqueWordCounter(std::string::size_type const min_word_length) {
+  std::map<std::string, int> counter;
+  std::string path;
+	std::string::size_type maxlen = 0;
+	std::string word;
+	std::cout << "enter file name: ";
+	std::cin >> path;
+	std::ifstream in(path);
+	if (!in) {
+			std::cerr << "failed to load file! Incorrect file name specified";
```
ghost: You're printing an error here if you can't open the file, which is good, but you're not aborting the function. This means the program is still trying to read from the `ifstream` object on line 95, which gives me a nasty pop-up message:



If your program exits with an error, it's conventional to return a non-zero `int` error code. Personally, I would have `uniqueWordCounter` return a `bool` or `int`, so `main` can know if the function was successful or not and can return a corresponding error code.

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+						processWord(more_word, counter, maxlen, min_word_length);
+				}
+				stream_word.clear();
+    }
+		else {
+  		processWord(word, counter, maxlen, min_word_length);
+		}
+  }
+	//vector containing the word and usage pairs from the map.
+  std::vector<std::pair<std::string, int>> count = mapToVector(counter);
```
ghost: As the compiler knows what type `mapToVector` returns (since you told it in your function definition), you can write 

```c++
auto count = mapToVector(counter);
``` 
instead of writing out the long templated type name again and the compiler will know what you mean.

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include <ctype.h>
+#include <fstream>
+#include <iostream>
+#include <map>
+#include <ostream>
+#include <string>
+#include <vector>
+#include <sstream>
+
+// replaces the hypens with whitespace - used in clean_word
```
ghost: I don't think this function is used in `clean_word`

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// convert a map to a vector of pairs
+std::vector<std::pair<std::string, int>> const
+mapToVector(const std::map<std::string, int> &map) {
+  return std::vector<std::pair<std::string, int>>(map.begin(), map.end());
+}
+
+// determine which pair is greater
+bool compare(std::pair<std::string, int> const pair_1,
```
ghost: This function name could be more descriptive, `compare` could be doing any number of things. Something more specific would improve readability when the function is called somewhere you can't see its docstring. 

Naming can be hard, and this is a tricky example, but maybe something along the lines of `isGreater` or `pairValueIsGreater`

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+  // remove characters if they are puctuation
+  word.erase(remove_if(word.begin(), word.end(), is_punc), word.end());
+  // convert characters to lowercase if they are uppercase
+  transform(word.begin(), word.end(), word.begin(), tolower);
+}
+
+//clean the words, count them and put them into the map
+void processWord(std::string& word, std::map<std::string, int>& counter, std::string::size_type& maxlen, std::string::size_type const min_word_length) {
+		clean_word(word);
+		maxlen = std::max(maxlen, word.size()); //find the longest word.
```
ghost: To minimise the number of operations we're doing, this line could be moved inside the `if` block below. We don't care about the length of words we're not adding to our counter

exercises/ciara_nightingale/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+    Results << counter[i].first
+            << std::string(pad + 1 - counter[i].first.size(), ' ')
+            << counter[i].second << "\n";
+  }
+}
+
+//counts the number of unique words no. of instances
+void uniqueWordCounter(std::string::size_type const min_word_length) {
+  std::map<std::string, int> counter;
+  std::string path;
```
ghost, line 85: I would put variable declarations as close to the first usage of that variable as possible (but don't declare it inside a loop). This saves people scrolling up and down to find out what type a variable is, but more importantly, it stops you declaring variables that may not be used later.

For example, you've declared `counter` and `maxlen` here, but if the file reading fails, you'll never use them and the program has reserved the memory space for nothing.

ghost (changes requested):
This works pretty well, nice job 👍 

The naming style used here is a bit inconsistent; you've used `snake_case` in some places and `camelCase` in others. Have a look at our [C++ coding standards](http://developer.mantidproject.org/Standards/CPPStandards.html#naming-conventions) for the naming conventions we use at Mantid. 

At Mantid, we use Doxygen to auto-generate documentation from our doc-strings. This means our doc-strings need to follow a specific format, see the [Function Headers](http://developer.mantidproject.org/Standards/CPPStandards.html#function-headers) section of our coding standards for an example.

I've left some more specific comments below

exercises/ciara_nightingale/ex01_basics/src/main.cpp
ciaranightingale: Correct me if I am wrong but doesn't the clean word function change the length of the string and so moving it inside the if statement would mean that three letter long words with punctuation would accidentally get through.

ghost: You're right, you definitely want the `clean_word` function outside the `if` statement. I was referring to line 50 where you're finding the max word length.

ciaranightingale: That makes more sense! Thank you.

exercises/ciara_nightingale/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,18 @@
...
+
+class Circle : public Shape {
+public:
+	///constructor if no arguments are given
+  Circle();
+	///constructor to create a circle of the given size
+  Circle(double radius);
+	///functions to calculate and return the perimeter and areaof the circle as a value of type double
+  virtual double perimeter() const override;
+  virtual double area() const override;
```
alicerussell1: These methods are not virtual in this class, instead they are overriding a virtual function so you do not need `virtual` before it.

exercises/ciara_nightingale/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,18 @@
...
+
+private:
+  double m_radius;
```
alicerussell1: Since the radius shouldn't change once constructed this can be made `const`. Similarly for the other shapes

exercises/ciara_nightingale/ex02_oo_basics/src/Shape.cpp
```diff
@@ -0,0 +1,25 @@
+#include "Shape.h"
+
+// Construct a shape if no arguments are given with type Nul and zero sides
+Shape::Shape() : m_type("Nul"), m_sides(0) {}
```
alicerussell1: Having a shape with no type or number of sides doesn't really make sense and it isn't used so this can probably be removed

exercises/ciara_nightingale/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,22 @@
...
+
+class Shape {
+public:
+  ///default constructor if no arguments are provided 
+  Shape();
+	///constructor to give the shape a type and the number of sides
+  Shape(std::string type, int sides);
+	///initialise the perimeter and area functions 
+  virtual double perimeter() const = 0;
+  virtual double area() const = 0;
```
alicerussell1, line 11: Good use of purely virtual functions and well done for making it `const`

exercises/ciara_nightingale/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,22 @@
...
+	///function to provide read only acess to the type and number of sides variables
+  std::string getType() const;
+  int getSides() const;
+
+private:
+  std::string m_type;
+  int m_sides;
```
alicerussell1: These can also be `const` as the should not be changed once the shape has been constructed

exercises/ciara_nightingale/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,109 @@
...
+/*Prints the types of all of the shapes
+*
+* @param type Type of shape the user wants to be printed
+*
+*/
+void ShapeSorter::printType(std::string type) const {
+  std::cout << "TYPES"
+            << "\n";
+	// Loop through all the shapes in the vector
+  for (auto const &shape : m_shapes) {
```
alicerussell1, line 31: Good use of a range-base for loop

exercises/ciara_nightingale/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,29 @@
+#ifndef SHAPESORTER_H
+#define SHAPESORTER_H
+#include "Circle.h"
+#include "Rectangle.h"
+#include "Shape.h"
+#include "Triangle.h"
```
alicerussell1: Since the class is only using `shape` you don't need to include `Circle.h`, `Rectangle.h` or `Triangle.h`. It is best practice to only include what you use.

exercises/ciara_nightingale/ex02_oo_basics/src/Rectangle.h
```diff
@@ -0,0 +1,21 @@
+#ifndef RECTANGLE_H
+#define RECTANGLE_H
+#include "Shape.h"
+
+class Rectangle : public Shape {
+public:
+  ///default constructor to handle situations when no arguments are given
+  Rectangle();
+	///constructor to handle squares (one side length given)
+  Rectangle(double width);
```
alicerussell1: I can see you are treating squares as just a rectangle with one side length but it would be better to treat it as a shape of it's own.

exercises/ciara_nightingale/ex02_oo_basics/src/Square.h
```diff
@@ -0,0 +1,5 @@
+#include "Shape.h"
+
+class Square:Shape {
+
```
alicerussell1, line 13: If a file isn't used it doesn't need to be included (however it would be best to deal with squares as either own shape - see comment above)

exercises/ciara_nightingale/ex02_oo_basics/src/Circle.h
alicerussell1: Well done for including `override` here

exercises/ciara_nightingale/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,29 @@
+#include "Circle.h"
+#include "Shape.h"
```
alicerussell1: You do not need to include `Shape.h` here as it is included in the header file already.

exercises/ciara_nightingale/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <math.h>
+
+// Constructor to initialise the circle if no arguments are given
+Circle::Circle() 
+		: Shape("Circle", 1), m_radius(0) {}
```
alicerussell1: You have included a constructor here with no arguments. This doesn't really make sense as every circle should have a radius so we should not allow the construction it without this information. It may also cause mistakes if someone else was using this code. Since it is not used I would recommend removing it. Similarly for the other shapes

alicerussell1 (changes requested):
This is a review for exercises 2. This is a really good solution, well done! The level of documentation in this is great. I've added some specific comments below, let me know if anything doesn't make sense.

alicerussell1 (approved):
Exercise 2 is now looking good 👍

martyngigg:
I'm going to close the PR (I'll leave the branch around) as I assume work is finished on this now.

PR #49, opened 2019-07-31
Giovanni Di Siena Exercise Solutions

giovannidisiena:
Completed functionality for ex_01 [word counter](https://www.mantidproject.org/New_Starter_C++_introduction#Exercise).

Test using [Holmes.txt](https://github.com/mantidproject/newstarter/files/3450907/Holmes.txt) (or any .txt).


exercises/giovanni_disiena/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+	//declare file read boolean and unsigned max length
+	bool read_bool;
+	std::size_t max_len = 0;
+
+	// read file
+	try {
+		read_bool = readFile(file_name, wordCounts, max_len);
+	} catch (const char* msg) {
+		std::cerr << msg << std::endl;
+		read_bool = false;
```
ghost: If reading the file has failed, then we probably want to return or throw here, rather than run the rest of the code.

exercises/giovanni_disiena/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+	}
+	else {
+		// read contents to map
+		std::string word;
+		while (inFile >> word) {
+			// clean words
+			std::vector<std::string> cleaned_words;
+			std::string lowered = str_tolower(word);
+			cleanWord(lowered, cleaned_words);
+			for (auto it = cleaned_words.begin(); it != cleaned_words.end(); ++it) {
```
ghost: This would be a good place to use a range-based `for` loop, e.g.:
```c++
for (auto &cleanWord : cleaned_words) { ... }
```
This gives you the value of `word` directly and means you don't have to keep dereferencing `it`. 

It's the equivalent of the below in Python
```python
for clean_word in cleaned_words:
```

exercises/giovanni_disiena/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+ *	storing all unique words and their numbers of occurrence
+ *
+ * @param filename: name of the input file to read
+ * @param inputmap: map which will store all unique words and instances
+ * @param char_min: default variable to limit the length of words
+ * @return bool: check for read success
+ *
+ */
+
+bool readFile(const std::string& filename, MapType& inputmap, std::size_t& max_len, const std::size_t char_min) {
```
ghost: The name `char_min` could be more descriptive, maybe `minWordLength`? You can also pass by reference, especially since it's a `const`

exercises/giovanni_disiena/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+	else {
+		// read contents to map
+		std::string word;
+		while (inFile >> word) {
+			// clean words
+			std::vector<std::string> cleaned_words;
+			std::string lowered = str_tolower(word);
+			cleanWord(lowered, cleaned_words);
+			for (auto it = cleaned_words.begin(); it != cleaned_words.end(); ++it) {
+				if ((*it).size() > char_min) {
```
ghost: Should this be `>=`? At the moment you're not adding any length 4 strings

exercises/giovanni_disiena/ex01_basics/src/main.h
```diff
@@ -0,0 +1,25 @@
...
+#include <regex>
+#include <string>
+
+// declare map/lamba/pair/set types which will store values
+// in non-increasing order unsing comparison logic
+typedef std::map<std::string, int> MapType;
+typedef std::pair<std::string, int> Pair;
+typedef const std::function<bool(Pair, Pair)>& Comparator;
+
+std::string str_tolower(std::string&);
```
ghost: Always include variable names in the function declarations. The (hopefully descriptive) variable names give lots of information about their purpose and hence how the function works. This is useful when skimming through header files to get a quick idea of how classes/functions are work

exercises/giovanni_disiena/ex01_basics/src/main.h
```diff
@@ -0,0 +1,25 @@
...
+#include <functional>
+#include <iostream>
+#include <map>
+#include <regex>
+#include <string>
+
+// declare map/lamba/pair/set types which will store values
+// in non-increasing order unsing comparison logic
+typedef std::map<std::string, int> MapType;
+typedef std::pair<std::string, int> Pair;
```
ghost: I like the use of `typedef` here, but the names could be clearer. Try to retain the information about what the container's holding. Maybe `StrIntMap` and `StrIntPair` or `WordCountPair/Map`?

exercises/giovanni_disiena/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+ * @return void
+ *
+ */
+
+void cleanWord(std::string& word, std::vector<std::string>& cleaned_words) {
+	static const std::regex string_regex("(\\B'\b|\b'\\B)|([\\.,:;!?()\"-]+)");
+	auto first = std::sregex_token_iterator(word.begin(), word.end(), string_regex, -1);
+	std::copy(first, std::sregex_token_iterator(), std::back_inserter(cleaned_words));
+}
+
```
ghost, line 71: Great doc-strings! I'd remove these blank lines between them and their functions though

exercises/giovanni_disiena/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+ *	appearing after words
+ *
+ * @param word: string to be cleaned
+ * @param cleaned_words: empty vector to store cleaned strings
+ * @return void
+ *
+ */
+
+void cleanWord(std::string& word, std::vector<std::string>& cleaned_words) {
+	static const std::regex string_regex("(\\B'\b|\b'\\B)|([\\.,:;!?()\"-]+)");
```
ghost: Maybe I'm missing something, but do you need the `(\\B'\b|\b'\\B)`? Also, it may be more efficient to run this regex on the entire contents of the file rather than on each word (don't worry too much about this if you think it will take too long, feel free to spend the time carrying on with exercise 2). Regex can be quite inefficient on smaller strings. 

Also, at Mantid, we write constant variables in all uppercase with underscores separating words

exercises/giovanni_disiena/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+ */
+
+const void writeFile(const std::vector<Pair>& orderedPair, const std::size_t& max_len, bool& read_bool) {
+	if (!read_bool) {
+		throw "Unable to write file";
+	}
+	else {
+		std::ofstream outfile("word_counts.txt");
+		std::string header_1 = "Word", header_2 = "Usage";
+		outfile << header_1 << formatWhitespace(header_1, max_len) << header_2 << "\n\n";
```
ghost: You should probably check that the file has been opened successfully before trying to write to it. Similar to what you did in `readFile`

exercises/giovanni_disiena/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+/** Takes a string from which to clean specified characters from
+ *	appearing after words
+ *
+ * @param word: string to be cleaned
+ * @param cleaned_words: empty vector to store cleaned strings
+ * @return void
+ *
+ */
+
+void cleanWord(std::string& word, std::vector<std::string>& cleaned_words) {
```
ghost: You could `const` `word` here

exercises/giovanni_disiena/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+ * @return bool: check for read success
+ *
+ */
+
+bool readFile(const std::string& filename, MapType& inputmap, std::size_t& max_len, const std::size_t char_min) {
+	// attempt to open file
+	std::ifstream inFile;
+	inFile.open(filename + ".txt");
+	if (!inFile) {
+		throw "Unable to open requested file";
```
ghost: It's unusual to throw a string in this way, usually you'd throw some form of exception e.g. `std::runtime_error`

exercises/giovanni_disiena/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+ * @param outputVector: vector of pairs to stored ordered words
+ * @param cmp: lambda function comparator
+ * @return void
+ *
+ */
+
+void orderWords(const MapType& cleanedMap, std::vector<Pair>& outputVector, const Comparator& cmp) {
+	outputVector.reserve(cleanedMap.size());
+	std::copy(cleanedMap.begin(), cleanedMap.end(), std::back_inserter(outputVector));
+	std::sort(outputVector.begin(), outputVector.end(), cmp);
```
ghost: Nice 👍

ghost (changes requested):
This is nice neat code with great doc-strings, good job 👍.

Just a quick word on formatting. At Mantid, for C++, we indent 2 characters using spaces. The reason for using spaces is that different programs can interpret tab characters as different sizes, for example reviewing your code on Github the tab character is the same as 8 spaces, whereas on Visual Studio its 2. Using spaces instead of tabs ensures the code looks the same everywhere. There should be an option on whatever IDE you're using to convert all tabs to spaces on save.

The naming style you've used is a bit inconsistent, you've got `camelCase` in some places and `snake_case` in others (we use `camelCase` for C++). See our [C++ Naming Conventions](http://developer.mantidproject.org/Standards/CPPStandards.html#naming-conventions) for a guide on what we do at Mantid.

I've left a few more comments below

exercises/giovanni_disiena/ex01_basics/src/main.cpp
giovannidisiena: I think the task asks for strings longer than 4 characters, however it isn't always clear!

task1

exercises/giovanni_disiena/ex01_basics/src/main.cpp
giovannidisiena: `(\\B'\b|\b'\\B)` removes apostrophes from the beginning and/or end of a word, however I don't know if there is a more succinct way? I will have a look at a quick fix for the running of the regex 👍

exercises/giovanni_disiena/ex01_basics/src/main.cpp
ghost: Ah, my mistake

exercises/giovanni_disiena/ex01_basics/src/main.cpp
giovannidisiena: Too much work to justify, current implementation is okay - moving on to induction tasks

giovannidisiena:
`make_unique` still seems to be causing RHEL build fail - any idea how to resolve this?

ghost:
It's building against C++11 on RHEL and `std::make_unique` was only added in C++14. If you go into `CMake/CommonSetup.cmake` and, on line 16, change `-std=c++11` to `-std=c++14`, then it should work. My fault for not changing it on master yet!

ghost (changes requested):
It looks like you've committed your build tree. Could we undo this? In general build/binary files are things you don't want in your repository. 

The `.gitignore` file tells git what files/folders to ignore. It's set to ignore the `builds` folder, but not `build`. You'll want to either rename your build folder to `builds`, or add `build/*` to your `.gitignore` file. You may need to drop the commit where you added the build files. Let me know if you need any help with this.

exercises/giovanni_disiena/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,18 @@
...
+class Circle : public ShapeBase {
+public:
+  // constructs Circle object
+  Circle(double);
+  // overriden functions
+  double getPerimeter() const override;
+  double getArea() const override;
+  double getRadius() const;
+private:
+  const double m_radius;
```
alicerussell1, line 16: Good use of `const` and good naming convention 👍

exercises/giovanni_disiena/ex02_oo_basics/src/Rectangle.cpp
```diff
@@ -0,0 +1,48 @@
...
+Rectangle::Rectangle(double length, double width)
+    : ShapeBase(4, "Rectangle"), m_length(length), m_width(width){};
+
+/** Constructs Rectangle (Square) object, setting width=length
+ *
+ * @param length: rectangle length
+ *
+ */
+Rectangle::Rectangle(double length)
+    : ShapeBase(4, "Square"), m_length(length), m_width(length) {}
```
alicerussell1: It would be better to treat square as a separate shape rather than as a rectangle with one length.

exercises/giovanni_disiena/ex02_oo_basics/src/Rectangle.h
```diff
@@ -0,0 +1,21 @@
+#ifndef RECTANGLE_H
+#define RECTANGLE_H
+
+#include "ShapeBase.h"
+
+class Rectangle : public ShapeBase {
+public:
+  // constructs Rectangle object
+  Rectangle(double, double);
```
alicerussell1: It is good practice to include the name of the variables, it makes it much easier to read and use.

exercises/giovanni_disiena/ex02_oo_basics/src/ShapeBase.h
```diff
@@ -0,0 +1,21 @@
...
+public:
+  // constructs and initialises ShapeBase object
+  ShapeBase(unsigned int, std::string);
+  // functions to get member variables
+  virtual double getPerimeter() const = 0;
+  virtual double getArea() const = 0;
+  std::string getName() const;
+  unsigned int getSides() const;
+private:
+  unsigned int m_sides;
```
alicerussell1: It would be clearer if this was called m_numberOfSides or something similar as if someone else was using it they might think this was the length of the sides or something

exercises/giovanni_disiena/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,17 @@
...
+public:
+  Shape(std::shared_ptr<ShapeBase>);
+  Shape(const Shape &);
+  Shape &operator=(const Shape &);
+  std::shared_ptr<ShapeBase> getBase() const;
+private:
+  std::shared_ptr<ShapeBase> m_shapeBasePointer;
+};
+
+#endif // SHAPE_H
```
alicerussell1: I am slightly confused as to the purpose of this class, it doesn't seem to be used anywhere.

alicerussell1: I have seen it has been removed from the `CMakeLists.txt` so I assume you were removing it? If so you need to remove the file as well.

exercises/giovanni_disiena/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,102 @@
...
+ * @param vecShape: vector of unique pointers to ShapeBase objects
+ * @param type: type of shape
+ *
+ */
+void ShapeSorter::printType(vecPtr & vecShape, std::string type) {
+  unsigned int matchCounter = 0;
+  std::cout << std::string(80, '*') << "\n";
+
+  // loop through vector of pointers to ShapeBase objects
+  for (std::vector<std::unique_ptr<ShapeBase>>::const_iterator it = vecShape.begin();
```
alicerussell1: You can use `auto` here which would make it much easier to read. A range based for loop could also be used so this would become:
`for (const auto &it : vecShape)`

exercises/giovanni_disiena/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,102 @@
...
+ * @param vecShape: vector of unique pointers to ShapeBase objects
+ * @param sides: number of sides
+ *
+ */
+void ShapeSorter::printSides(vecPtr & vecShape, unsigned int sides) {
+  unsigned int matchCounter = 0;
+  std::cout << std::string(80, '*') << "\n";
+
+  // loop through vector of pointers to ShapeBase objects
+  for (std::vector<std::unique_ptr<ShapeBase>>::const_iterator it = vecShape.begin();
```
alicerussell1: This can also be a range-base for loop, see above

exercises/giovanni_disiena/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,102 @@
+#include "ShapeSorter.h"
+
+/** Prints shapes of given type
+ *
+ * @param vecShape: vector of unique pointers to ShapeBase objects
+ * @param type: type of shape
+ *
+ */
+void ShapeSorter::printType(vecPtr & vecShape, std::string type) {
```
alicerussell1: `type` can be passed by reference here

exercises/giovanni_disiena/ex02_oo_basics/src/Use.h
```diff
@@ -0,0 +1,16 @@
...
+  Use(const Use &u);
+  ~Use();
+  Use &operator=(const Use &);
+  operator size_t() const;
+
+private:
+  unsigned *m_refCount;
+};
+
+#endif // USE_H
```
alicerussell1: This is also not in the `CMakeLists` and not used anywhere, you need to remove the file if you are not including it.

exercises/giovanni_disiena/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,24 @@
+#include "main.h"
+
+int main() {
+  // populate vector with unique pointers to ShapeBase objects
+  std::vector<std::unique_ptr<ShapeBase>> vecShape;
```
alicerussell1, line 5: Good use of smart pointers well done 👍

exercises/giovanni_disiena/ex02_oo_basics/src/Rectangle.h
alicerussell1: Similarly for all the shapes

exercises/giovanni_disiena/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,24 @@
...
+#define SHAPESORTER_H
+
+#include <algorithm>
+#include <iostream>
+#include <memory>
+#include <string>
+#include <vector>
+#include "ShapeBase.h"
+
+typedef std::vector<std::unique_ptr<ShapeBase>> vecPtr;
```
alicerussell1, line 11: Good use of `typedef`

exercises/giovanni_disiena/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,24 @@
...
+
+class ShapeSorter {
+public:
+  // member functions to print data
+  static void printType(vecPtr &, std::string);
+  static void printSides(vecPtr &, unsigned int);
+  static void printAreaDescending(vecPtr &);
+  static void printPerimeterDescending(vecPtr &);
+
+private:
```
alicerussell1: You do not need to include `private` if there is nothing private

alicerussell1 (changes requested):
Review for exercises 2:
This is a very good solution with good documentation, well done. I have left some comments below.

exercises/giovanni_disiena/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,96 @@
...
+void ShapeSorter::printAreaDescending(vecPtr &vecShape) {
+  std::cout << std::string(80, '*') << "\n";
+  // sort pointed-to vector objects into descending order of area
+  std::sort(vecShape.begin(), vecShape.end(),
+            [](const std::unique_ptr<ShapeBase> &a,
+               const std::unique_ptr<ShapeBase> &b) {
+              return a->getArea() > b->getArea();
+            });
+  unsigned int counter = 0;
+  for (std::vector<std::unique_ptr<ShapeBase>>::iterator it = vecShape.begin();
```
alicerussell1: This can be a range-based for loop, similar to the one you've done in the method above

exercises/giovanni_disiena/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,96 @@
...
+void ShapeSorter::printPerimeterDescending(vecPtr &vecShape) {
+  std::cout << std::string(80, '*') << "\n";
+  // sort pointed-to vector objects into descending order of area
+  std::sort(vecShape.begin(), vecShape.end(),
+            [](const std::unique_ptr<ShapeBase> &a,
+               const std::unique_ptr<ShapeBase> &b) {
+              return a->getPerimeter() > b->getPerimeter();
+            });
+  unsigned int counter = 0;
+  for (std::vector<std::unique_ptr<ShapeBase>>::iterator it = vecShape.begin();
```
alicerussell1: And here

alicerussell1 (changes requested):
Exercise 2 is now looking good 👍 Just got a one more suggestion below

giovannidisiena:
Completed training exercises.

PR #50, opened 2019-08-07
Move the C++ standard to C++14

ghost:
In CMake, set compiler flags to target at C++14.

### DIFF
--- a/CMake/CommonSetup.cmake
+++ b/CMake/CommonSetup.cmake
@@ -12,12 +12,7 @@ set ( CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin )
 # This allows us to group targets logically in Visual Studio
 set_property ( GLOBAL PROPERTY USE_FOLDERS ON )
 
-if ( CMAKE_COMPILER_IS_GNUCXX )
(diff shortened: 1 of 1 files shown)


CMake/CommonSetup.cmake
```diff
@@ -13,7 +13,11 @@ set ( CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin )
 set_property ( GLOBAL PROPERTY USE_FOLDERS ON )
 
 if ( CMAKE_COMPILER_IS_GNUCXX )
```
martyngigg: Thanks for do this.
Can we swap these `if`s out for the more modern way of doing this like `set(CMAKE_CXX_STANDARD 14)` in [`CommonSetup.cmake`](https://github.com/mantidproject/mantid/blob/3b516d98b0ee6bd06e98ca98cb8b0a5efa08b07e/buildconfig/CMake/CommonSetup.cmake#L241)

martyngigg (changes requested):

martyngigg (approved):
Looks good :+1:

martyngigg:
Build failures are because there are no projects defined and are expected.

PR #52, opened 2019-10-15
Add files for C++ exercises 1 and 2

DannyHindson:
Fixes #xxxx. 

Two command line programs for exercise 1 (Word Count) and exercise 2 (Shape Sorter) C++ new starter exercises:

https://www.mantidproject.org/New_Starter_C%2B%2B_introduction

To test:

1. Exercise 1: Save Holmes.txt file to D:\newstarter\builds\WordCountDataFiles\ and specify "Holmes.txt" as command line parameter
2. Exercise 2: Run Shapes.exe and it will print out the required shape details

**Reviewer**
Please comment on the following (full description):
**Code Review**

- Is the code of an acceptable quality?
- Does the code conform to the coding standards?
- Are the unit tests small and test the class in isolation?
- If there are changes in the release notes then do they describe the changes appropriately?

**Functional Tests**

- Do changes function as described? Add comments below that describe the tests performed?
- Do the changes handle unexpected situations, e.g. bad input?
- Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as Approve. A member of @mantidproject/gatekeepers will take care of it.

### DIFF
--- a/exercises/danny_hindson/ex01_basics/src/main.cpp
+++ b/exercises/danny_hindson/ex01_basics/src/main.cpp
@@ -1,0 +1,88 @@
+// Mantid Repository : https://github.com/mantidproject/mantid
+//
+// Copyright &copy; 2019 ISIS Rutherford Appleton Laboratory UKRI,
+//     NScD Oak Ridge National Laboratory, European Spallation Source
+//     & Institut Laue - Langevin
+// SPDX - License - Identifier: GPL - 3.0 +
+
+#include <algorithm>
+#include <cctype>
+#include <fstream>
+#include <iostream>
+#include <map>
+#include <sstream>
+#include <string>
+#include <vector>
+
+bool compare(const std::pair<std::string,int>& x, const std::pair<std::string,int>& y) {
+  return x.second > y.second;
+}
+
+int main(int argc, char ** argv)
+{
+  std::string fileName;
+  if (argc == 2) {
+    fileName = argv[1];
+
+    std::string folderName ="D:\\newstarter\\builds\\WordCountDataFiles\\";
+    std::ifstream inFile(folderName + fileName);
+    
+    if (inFile) {
+      std::string s_withhyphens,s;
+      std::map<std::string, int> counters;
+
+      // read file in one word at a time
+      while (inFile >> s_withhyphens) {
+        std::istringstream ss(s_withhyphens);
+        // split any words containing hyphens into two
+        while (std::getline(ss, s, '-')) {
+          for (int i = 0; i < s.length(); i++) {
+            s[i] = tolower(s[i]);
+          }
+          char charsToRemove[] = ".,?'\"!():";
+          for (int i = 0; i < strlen(charsToRemove); i++) {
+            s.erase(std::remove(s.begin(), s.end(), charsToRemove[i]), s.end());
+          }
+          if (s.length() > 4) {
+            counters[s]++;
+          }
+        }
+      }
+
+      std::vector<std::pair<std::string, int>> vec;
+
+      //sort(counters.begin(),counters.end(),Compare); // not possible on a map because doesn't have random access
+
+      // copy to vector of pairs so can sort. Could have copied to another map (multimap) instead
+      std::copy(counters.begin(), counters.end(), std::back_inserter<std::vector<std::pair<std::string, int>>>(vec));
+      sort(vec.begin(), vec.end(), compare);
+
+      std::ofstream outFile(folderName + "results.txt");
+      if (outFile) {
+        outFile << "Word" << "\t" << "Usage" << std::endl;
+        for (std::vector<std::pair<std::string, int>>::const_iterator it = vec.begin(); it != vec.end(); ++it) {
+          outFile << it->first << "\t" << it->second << std::endl;
+        }
+      }
+      else {
+        std::cout << "Cannot open specified output file" << std::endl;
+      }
+      inFile.close();
+      outFile.close();
+
+      std::cout << "Word counting succeeded!" << std::endl;
+    }
+    else {
+      std::cout << "Cannot open specified input file" << std::endl;
+      return 1;
+    }
+  }
+  else if (argc == 1) {
+    std::cout << "No filename supplied" << std::endl;
+    return 1;
+  }
+  else {
+    std::cout << "Unexpected number of arguments supplied" << std::endl;
+    return 1;
+  }
+}
--- a/exercises/danny_hindson/ex01_basics/src/main.cpp
+++ b/exercises/danny_hindson/ex01_basics/src/main.cpp
@@ -12,6 +12,7 @@
 #include <map>
 #include <sstream>
 #include <string>
+#include <string.h>
 #include <vector>
 
 bool compare(const std::pair<std::string,int>& x, const std::pair<std::string,int>& y) {
--- a/exercises/danny_hindson/ex01_basics/src/main.cpp
+++ b/exercises/danny_hindson/ex01_basics/src/main.cpp
@@ -15,7 +15,21 @@
 #include <string.h>
 #include <vector>
 
-bool compare(const std::pair<std::string,int>& x, const std::pair<std::string,int>& y) {
+using WordCount = std::pair<std::string, int>;
+using WordCounts = std::map<std::string, int>;
+
+namespace {
+  constexpr auto FOLDER_NAME = "D:\\newstarter\\builds\\WordCountDataFiles\\";
+  constexpr auto CHARS_TO_REMOVE = ".,?'\"!():";
+}
+
+/**
+ * Returns the sort order of the Word Counts - sort descending
+ * @param x First word count
+ * @param y Second word count
+ * @return boolean which is true if the first word count is bigger than the second
+ */
+bool compare(const WordCount& x, const WordCount& y) {
   return x.second > y.second;
 }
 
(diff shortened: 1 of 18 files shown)


exercises/danny_hindson/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include <cctype>
+#include <fstream>
+#include <iostream>
+#include <map>
+#include <sstream>
+#include <string>
+#include <string.h>
+#include <vector>
+
+bool compare(const std::pair<std::string,int>& x, const std::pair<std::string,int>& y) {
```
martyngigg: In the main code base we generally try to document functions/methods/classes using Doxygen as in an example [DataService](https://github.com/mantidproject/mantid/blob/master/Framework/Kernel/inc/MantidKernel/DataService.h#L420).

I would also consider defining an alias for `std::pair<std::string,int>` as it's used quite frequently over the code. In modern C++ this can be done as 

```
using WordCount = std::pair<std::string, int>;
```
This would go above the function definition. Giving it a name like this also serves to improve readability.

exercises/danny_hindson/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,89 @@
...
+  return x.second > y.second;
+}
+
+int main(int argc, char ** argv)
+{
+  std::string fileName;
+  if (argc == 2) {
+    fileName = argv[1];
+
+    std::string folderName ="D:\\newstarter\\builds\\WordCountDataFiles\\";
```
martyngigg: We generally try to define constants used in a single file in an anonymous namespace in the file using uppercase naming conventions,

e.g. a string like this could be written in C++14 as
```
namespace {
    constexpr auto FOLDER_NAME = "D:\\newstarter\\builds\\WordCountDataFiles\\";
}
```
This would go just under the `include` lines and the anonymous namespace serves to keep it defined only in this file.

exercises/danny_hindson/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,89 @@
...
+    std::ifstream inFile(folderName + fileName);
+    
+    if (inFile) {
+      std::string s_withhyphens,s;
+      std::map<std::string, int> counters;
```
martyngigg: I would also consider putting 
```
using WordCounts = std::map<std::string, int>;
```
along with the pair definition above to improve readability.

exercises/danny_hindson/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+      // read file in one word at a time
+      while (inFile >> s_withhyphens) {
+        std::istringstream ss(s_withhyphens);
+        // split any words containing hyphens into two
+        while (std::getline(ss, s, '-')) {
```
martyngigg: Can we define `s` closer to where it is used? Keeping the scope of variables small helps readability and also avoids making mistake when accidentally using a variable when it shouldn't have been used.

Could we also use variable names like `word` instead of `s`?

exercises/danny_hindson/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,89 @@
...
+          for (int i = 0; i < s.length(); i++) {
+            s[i] = tolower(s[i]);
+          }
+          char charsToRemove[] = ".,?'\"!():";
```
martyngigg: Good name 👍 Can we also define this alongside the FOLDER_NAME as a constant?

```
namespace {
  constexpr auto CHARS_TO_REMOVE[] = ".,?'\"!():";
}

File: exercises/danny_hindson/ex01_basics/src/main.cpp [hunk 6]
```diff
@@ -0,0 +1,89 @@
...
+          for (int i = 0; i < strlen(charsToRemove); i++) {
+            s.erase(std::remove(s.begin(), s.end(), charsToRemove[i]), s.end());
+          }
+          if (s.length() > 4) {
+            counters[s]++;
+          }
+        }
+      }
+
+      std::vector<std::pair<std::string, int>> vec;
```
martyngigg: Can we avoid names like `vec`? They are a little generic and don't offer much meaning.

exercises/danny_hindson/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,89 @@
...
+      //sort(counters.begin(),counters.end(),Compare); // not possible on a map because doesn't have random access
+
+      // copy to vector of pairs so can sort. Could have copied to another map (multimap) instead
+      std::copy(counters.begin(), counters.end(), std::back_inserter<std::vector<std::pair<std::string, int>>>(vec));
+      sort(vec.begin(), vec.end(), compare);
+
+      std::ofstream outFile(folderName + "results.txt");
+      if (outFile) {
+        outFile << "Word" << "\t" << "Usage" << std::endl;
+        for (std::vector<std::pair<std::string, int>>::const_iterator it = vec.begin(); it != vec.end(); ++it) {
```
martyngigg: This is perfectly correct but in C++11 and beyond it has thankfully been made possible to write this without having to type out the iterator stuff with range-based for loops:
```
for(const auto & wordAndCount : vec) {
   outFile << wordAndCount.first << "\t" << wordAndCount.second << std::endl;
}

File: exercises/danny_hindson/ex02_oo_basics/src/Shape.h [hunk 8]
```diff
@@ -0,0 +1,27 @@
...
+
+#ifndef SHAPE_H_
+#define SHAPE_H_
+
+#include <string>
+
+class Shape {
+public:
+  Shape(std::string ShapeType, int nsides) : ShapeType(ShapeType), nsides(nsides) {};
+  virtual ~Shape() {};
```
martyngigg: Again this is perfectly valid but C++11 has introduced the `default` qualifier to avoid having to rewrite what the compiler would generate when we are forced to include things like destructors in this manner:
```c++
virtual ~Shape() = default;
```

exercises/danny_hindson/ex02_oo_basics/src/Shape.cpp
```diff
@@ -0,0 +1,32 @@
+// Mantid Repository : https://github.com/mantidproject/mantid
+//
+// Copyright &copy; 2019 ISIS Rutherford Appleton Laboratory UKRI,
+//     NScD Oak Ridge National Laboratory, European Spallation Source
+//     & Institut Laue - Langevin
+// SPDX - License - Identifier: GPL - 3.0 +
+
+#include "Shape.h"
+
+/**
```
martyngigg, line 10: Good documentation on all of the members.

exercises/danny_hindson/ex02_oo_basics/src/Rectangle.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include "Shape.h"
+
+class Rectangle : public Shape {
+public:
+  double perimeter() const override;
+  double area() const override;
+  Rectangle(double side1, double side2) : Shape("Rectangle", 4), side1(side1), side2(side2) {};
+private:
+  double side1;
```
martyngigg: Just a naming convention for the main code. We prefix all member variables with `m_` so that when reading the code it's clearer that this has scope outside of a method.

exercises/danny_hindson/ex02_oo_basics/src/Rectangle.h
```diff
@@ -0,0 +1,23 @@
...
+#ifndef RECTANGLE_H_
+#define RECTANGLE_H_
+
+#include "Shape.h"
+
+class Rectangle : public Shape {
+public:
+  double perimeter() const override;
+  double area() const override;
+  Rectangle(double side1, double side2) : Shape("Rectangle", 4), side1(side1), side2(side2) {};
```
martyngigg: We generally define constructors, copy constructors etc at the top and then other public methods.

exercises/danny_hindson/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+  Square sq(2.5);
+  Rectangle r(3, 7);
+  Circle c(4.1);
+  Triangle t(1, 10);
+
+  // create vector of pointers to Shape rather than actual Shapes so the polymorphism works
+  std::vector<Shape*> shapeVector;
+
+  shapeVector.push_back(&sq);
```
martyngigg, line 28: Good! People seem to often forget that you can take the address of a stack variable and you don't have to use `new`!

exercises/danny_hindson/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,26 @@
...
+
+#ifndef SHAPESORTER_H_
+#define SHAPESORTER_H_
+
+#include "Shape.h"
+#include <vector>
+
+class ShapeSorter {
+public:
+  ShapeSorter(std::vector<Shape*> v) : shapeVec(v) {};
```
martyngigg: This is perfectly valid code but causes a copy of the vector that can be avoided using `std::move` from C++11.
```c++
ShapeSorter(std::vector<Shape*> v) : shapeVec(std::move(v)) {};
```
I'd suggest a quick read over something about [move semantics](https://mbevin.wordpress.com/2012/11/20/move-semantics/)

exercises/danny_hindson/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,26 @@
...
+  virtual ~ShapeSorter() {};
+  void printShapes(std::string ShapeType) const;
```
martyngigg: Good use of `const` on the member definition.

By default C++ copies its arguments. Most of the time this is not what we want and methods like this can be defined as
```c++
  void printShapes(const std::string &ShapeType) const;
```
so they are passed as a reference and will not copy the input the string.

martyngigg (changes requested):
This is generally a good set of solutions and it's good that care has been taken to ensure the CI passes.

My points all relate to either how we do things in Mantid or some more modern practices that perhaps are not yet known about.

martyngigg:
I also meant to add that I found https://github.com/AnthonyCalandra/modern-cpp-features quite helpful for an over view of modern C++ features. We are currently using C++14 in mantid.

exercises/danny_hindson/ex02_oo_basics/src/ShapeSorter.h
DannyHindson: OK thanks. Think I've got that. That was new to me

exercises/danny_hindson/ex01_basics/src/main.cpp
DannyHindson: Thanks. So are people just keying in comments of this general style by hand into the code units? Or if I install Doxygen do I get some kind of tool to support creating this kind of documentation?

martyngigg: While doxygen itself doesn't provide a tool several IDEs will create a template documentation string using the parameter names in the function definitions. QtCreator can do this out of the box and Visual Studio can do it with a [plugin](https://marketplace.visualstudio.com/items?itemName=cschlosser.doxdocgen).

martyngigg (approved):
This looks good now :+1:

martyngigg:
Of course on normal code this would be merged but we don't want anyone's answers `master` so I'll close this without merging. I'll leave the branch around for a bit though.

PR #53, opened 2019-10-15
Stephen Smith exercises

StephenSmith25:
Hello,

This is my solution to exercises 1 and 2.

The source code was compiled on Ubuntu, so to test it you may need to build again:
(1) For the first exercise there is an "input.txt" file, which you'll need to copy and paste into the folder with the executable
(2) Should be able to run the resulting executable as a standalone command line program


exercises/Stephen_Smith/.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,16 @@
+{
```
martyngigg: The `.vscode` files are okay in this repo but in the main repo we don't include personal settings. These can be ignored on a per-repository basis using [.git/info/exclude](https://help.github.com/en/articles/ignoring-files#explicit-repository-excludes)

exercises/Stephen_Smith/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+ * Skeleton main routine
+ */
+
+
+#include <iostream>
+#include <fstream>
+#include <map>
+#include <iomanip>
+#include "../include/splitLineIntoWords.h"
+
```
martyngigg, line 6: Try to avoid extraneous whitespace as it tends to be jarring to the eye to read. A single blank line between the includes an namespace is fine.

exercises/Stephen_Smith/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+
+
+
+namespace{
```
martyngigg, line 8: Good use of an anonymous namespace :+1:

exercises/Stephen_Smith/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+    const int MIN_WORD_LENGTH=5;
```
martyngigg: Good naming of constant. In modern C++ we can now mark this `constexpr`:

```
constexpr int MIN_WORD_LENGTH{5};
```

exercises/Stephen_Smith/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// Remove punctuation from the input word
```
martyngigg: Good for providing a comment. In the main code base we use the doxygen tool to provide this type of documentation and also document the parameters, e.g [DataService](https://github.com/mantidproject/mantid/blob/master/Framework/Kernel/inc/MantidKernel/DataService.h#L420)

exercises/Stephen_Smith/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+int removePunctuation(std::string &iString)
+{
+
```
martyngigg: Similarly, please try and avoid too many extraenous new lines and keep them for separating blocks.

exercises/Stephen_Smith/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+    std::string::iterator itr = iString.begin() ;
+
+    static const std::string punctuation(".!,\"?!");
```
martyngigg: This could also be in your anonymous namespace as 
```c++
namespace {
constexpr auto PUNCTUATION = ".!,\"?!";
}
```

exercises/Stephen_Smith/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+    }
+
+    return 0;
+
+}
+
+// Function to compare counts of two words eached stored in std::pair with the associated count
+// Input paramter = std::pair of <word,count>
+bool compareWordCounts(std::pair<std::string,int> a,std::pair<std::string,int> b)
```
martyngigg: As `std::pair<std::string, int>` is used in a few places then maybe consider defining an alias above the anonymous namespace as

using WordCount = std::pair<std::string, int>;

exercises/Stephen_Smith/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+                // remove punctuation
+                removePunctuation(word);
+                // count number of letters in word
+                std::string::size_type word_length = word.length();
+                
+                // add word of length (>= MIN_WORD_LENGTH) to map container ;
+                if ( word_length >= MIN_WORD_LENGTH)
+                {
+                    // Add word and count to map
+                    auto retPair = wordCounts.insert(std::pair<std::string,int>(word,1));
```
martyngigg: This could use the `std::make_pair(word, 1);` utility in C++11 that avoids having to specify the template types.

exercises/Stephen_Smith/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+
+
+    file.close();
+
+    // Extract each <word,count> pair into a vector container
+    std::vector<std::pair<std::string,int>> sortedWordCount;
+    
+    for (auto word : wordCounts)
```
martyngigg: In this case you want to avoid copying each element as you loop over it so you would do:
```c++
for (const auto& word : wordCounts)
```

Even better would be to consider an STL algorithm:
```c++
    std::vector<std::pair<std::string, int>> sorted(wordCounts.size());
    std::copy(std::begin(wordCounts), std::end(wordCounts), std::begin(sorted));
```

`copy` has the nice advantage that it names what you are doing and avoids having to inspect what a loop is doing.

exercises/Stephen_Smith/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+
+    // Output words into a file
+    std::string output_filename = "sorted_" + input_filename ;
+    std::ofstream outputfile(output_filename, std::ofstream::out);
+    
+    if (outputfile.is_open())
+    {
+        outputfile << std::setw(20) << "Word" << std::setw(20) << "count" << std::endl;
+        for (auto wordAndCount : sortedWordCount)
```
martyngigg: Similar with `const auto &` here.

exercises/Stephen_Smith/ex01_basics/src/splitLineIntoWords.cpp
```diff
@@ -0,0 +1,49 @@
...
+{
+    return !(( c == '-') || (isspace(c)));
+
+}
+// Splits a line into words using spaces and hyphens as delimiters between the words 
+std::vector<std::string> splitLineIntoWords(const std::string& line)
+{
+
+    // iterator over the string
+    std::string::const_iterator wordStart = line.begin();
```
martyngigg: It's good to use `auto` as much as possible to avoid having to write these long-winded iterator types.

exercises/Stephen_Smith/ex01_basics/src/splitLineIntoWords.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+    // output container
+    std::vector<std::string> lineWords;
+
+
+    while(wordStart != line.end())
+    {
+
+        // Find first character of word
+        wordStart = std::find_if(wordStart,line.end(),notSpaceOrHyphen);
```
martyngigg, line 27: Good use of STL.

exercises/Stephen_Smith/ex02_oo_basics/include/Shapes.h
```diff
@@ -0,0 +1,218 @@
...
+}
+
+
+class BaseShape{
+
+public:
+    
+    // Constructor
+    BaseShape();
+    virtual ~BaseShape() = 0 ;
```
martyngigg: You are correct that we need to define this as virtual but it doesn't need to be abstract. We can define it as `virtual ~BaseShape() = default`

exercises/Stephen_Smith/ex02_oo_basics/include/Shapes.h
```diff
@@ -0,0 +1,218 @@
...
+        void setPerimeter(double perimeter) {this->m_perimeter = perimeter;}
+
+    private:
+        double m_area;
+        double m_perimeter;
+
+};
+
+// Circle class
+class Circle : public BaseShape
```
martyngigg: In is good practice to put each class in its own file. This avoid a client having to include more code than necessary if they just want a particular shape and it is also easier to find a particular class when looking for it as a developer.

exercises/Stephen_Smith/ex02_oo_basics/include/Shapes.h
```diff
@@ -0,0 +1,218 @@
...
+
+#include <string>
+#include <vector>
+#include <memory>
+#include <iostream>
+#include <math.h>  
+
+namespace{
+
+    const double PI=3.141592653589793238463;
```
martyngigg: This can be `constexpr`. In the main codebase we would use the `M_PI` constant defined in `<cmath>`.

exercises/Stephen_Smith/ex02_oo_basics/include/Shapes.h
```diff
@@ -0,0 +1,218 @@
+#ifndef SHAPES_H_
+#define SHAPES_H_
+
+#include <string>
+#include <vector>
+#include <memory>
+#include <iostream>
```
martyngigg: In the mantid code we would never include `<iostream>` in a header as it increases compile times a lot and having it in a header means it could spread quite easily.

Anything using it should be definined in a `.cpp` file.

exercises/Stephen_Smith/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+// *************************************************//
+//Shape sorter
+//Constructors
+ShapeSorter::ShapeSorter()
+{
+
+};
+// Constructor with input list of shapes
+ShapeSorter::ShapeSorter(const std::vector<std::shared_ptr<BaseShape>>& shapeList)
```
martyngigg: In C++11 with move semantics we could do this as

```c++
ShapeSorter::ShapeSorter(std::vector<std::shared_ptr<BaseShape>> shapeList) : m_shapeList(std::move(shapeList)) {}
```
to avoid having to copy the shapes in.

exercises/Stephen_Smith/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,119 @@
...
+{
+    this->m_shapeList = shapeList;
+}
+
+// Print out the names of the shapes stored in the ShapeSorter object
+void ShapeSorter::PrintShapeNames() const{
+
+    if (!this->m_shapeList.empty())
+    {
+        for (auto shape : this->m_shapeList)
```
martyngigg:
```suggestion
        for (const auto& shape : this->m_shapeList)
```

exercises/Stephen_Smith/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,119 @@
...
+// Function to sort shapes by area
+void ShapeSorter::printShapesByArea() {
+
+    if (! this->m_shapeList.empty())
+    {
+
+        // Sort shapes by area
+       std::cout << "Sorting by area" << std::endl;
+       std::sort(m_shapeList.begin(),m_shapeList.end(),
+       [] (const std::shared_ptr<BaseShape>& a, const  std::shared_ptr<BaseShape>& b){
```
martyngigg: Good use of a lambda!

exercises/Stephen_Smith/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "../include/ShapeSorter.h"
+
+
+using namespace std;
+int main(int, char **)
+{
+
+    // Create shapes
+    auto sCircle_1 = make_shared<Circle>(1.5);
```
martyngigg: It is fine to use `shared_ptr`, it's good that the memory is managed by something. A first point of call however should be `std::unique_ptr`. This is a simpler wrapper that makes it much clearer who owns a particular type as there is no sharing possible and is just a basic wrapper around `new`/`delete`.

martyngigg (changes requested):
This is a good set of solutions. :+1:  A few comments on style and a few more modern concepts

exercises/Stephen_Smith/ex01_basics/src/main.cpp
StephenSmith25: Thank you for the example, I will change accordingly. Should we document the function in both the header and the source file, or just one of them?

martyngigg: I tend to just document them in the source file. If anything needs updating it then doesn't cause a lengthy rebuild if it's deep in the code.

exercises/Stephen_Smith/ex02_oo_basics/src/main.cpp
StephenSmith25: Okay I'll keep that in mind. For now I've switched to using raw pointers referencing the stack generated objects. I need to do a bit more reading around these smart pointers.

exercises/Stephen_Smith/ex02_oo_basics/include/ShapeSorter.h
```diff
@@ -0,0 +1,34 @@
+#ifndef SHAPESORTER_H_
+#define SHAPESORTER_H_
+
+#include "Shapes.h"
```
martyngigg: One thing to be careful of in general is only including what you use so that we can make sure compile times don't build up. In this file only `vector`, `string` would be required. For `BaseShape` we actually do something called a [forward declaration](https://www.geeksforgeeks.org/what-are-forward-declarations-in-c/) so we don't need to include it in the header at all - it would just need to be included in the .cpp file.

StephenSmith25: Okay useful to know, I'll keep this in mind, and I have adjusted the file ShapeSorter.h accordingly. Am I right in saying you cannot forward declare the base class within the other shapes, e.g Triangle?

martyngigg: That's correct. You can only forward declare something when you don't need to know the size of the actual class. In `ShapeSorter` it only uses a pointer and they are always the same size. When inheriting the compiler needs to know the layout of the base class so the include will be needed.

PR #56, opened 2019-12-09
Mathieu tillet exercises

MathieuTillet:
Hello,

This is my implementation of the exercises 1 (words counting) and 2 (geometric shape classes)  of the new starter page.

The main.cpp of exercise 2 contains a few examples of the shapes and the methods on them.


exercises/Mathieu_Tillet/CMakeLists.txt.user
```diff
@@ -0,0 +1,428 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gvardany, line 1: I'm not sure if this file should be in git.

exercises/Mathieu_Tillet/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,125 @@
...
+ */
+
+#include <iostream>
+#include <fstream>
+#include <map>
+#include <vector>
+#include <algorithm>
+#include <tuple>
+
+using namespace std;
```
gvardany, line 12: While it's fine for the exercise, out in the wild it is not a good idea to do this. In the environment where there are many libraries, this might cause conflicts. For example `shared_ptr` exists both in `std` and `boost`. So it is advised to not to use the whole namespace, but to specify explicitly `std` in front of each data structure or algorithm you are using.

exercises/Mathieu_Tillet/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,125 @@
+/**
+ * Skeleton main routine
+ */
+
+#include <iostream>
```
gvardany, line 5: It is a good idea to sort these includes alphabetically.

exercises/Mathieu_Tillet/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,125 @@
...
+    char *file_name = argv[1];
+
+    string out_name;
+    // Name of the file in which the output will be written
+    if (argc > 2){
+        // catch the second argument if one is provided
+        out_name = string(argv[2]);
+    } else {
+        // else create a new file
+        out_name = "./out.txt";
```
gvardany, line 33: In the wild, one should avoid hardcoded paths, especially containing `/` since it breaks the cross-platform compatibility; Windows uses `\`. In reality we would use a library to abstract out that aspect, such as `Poco::File` or `std::filesystem`.

exercises/Mathieu_Tillet/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,125 @@
...
+        return -1;
+    }
+
+    string block;
+    size_t max_length = 4; // size of "Word", which we systematically write
+    size_t block_size;
+    bool should_split;
+
+    // the delimiters of words
+    string delimiters = ".,?'\"!():";
```
gvardany, line 50: This can be constant.

exercises/Mathieu_Tillet/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,125 @@
...
+    ifstream in;
+    in.open(file_name);
+
+    if (!in){
+        cerr << "Error when reading the file" << endl;
+        return -1;
+    }
+
+    string block;
+    size_t max_length = 4; // size of "Word", which we systematically write
```
gvardany, line 45: This can be constant.

exercises/Mathieu_Tillet/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,125 @@
...
+    size_t block_size;
+    bool should_split;
+
+    // the delimiters of words
+    string delimiters = ".,?'\"!():";
+
+    while (in >> block){
+        // iterate on each space or endofline separated block in the text
```
gvardany, line 53: Good to know how to do is manually, but in mantid we have boost available, so I suggest we try to use `boost::split` to tokenise the strings.

exercises/Mathieu_Tillet/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,125 @@
...
+                    max_length = block_size - start;
+                }
+            }
+        }
+    }
+    in.close();
+
+    // we store the elements of word_counts in a vector so they can be sorted
+    vector<tuple<int, string>> vec_count;
+    for (map<string, int>::iterator it = word_count.begin(); it != word_count.end(); ++it){
```
gvardany, line 100: You should be able to `std::back_inserter` here.

exercises/Mathieu_Tillet/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,125 @@
...
+        vec_count.push_back(tuple<int, string>(it->second, it->first));
+    }
+
+    // sort the vector's elements in increasing order, by the first value of the tuple (ie the number of occurences)
+    sort(vec_count.begin(), vec_count.end());
```
gvardany, line 105: You can sort here descending straight using a lambda function.

exercises/Mathieu_Tillet/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    ofstream out;
+    size_t spaces;
+    out.open(out_name);
+    out << "Word     Usage" << endl;
+    // we go through the elements of the vector in reverse order so that they are sorted decreasingly in the output
+    for (vector<tuple<int, string>>::iterator it = vec_count.end(); it != vec_count.begin(); --it){
```
gvardany, line 112: You should be able to use a range based loop here `for (const auto& item : vector_count)`

exercises/Mathieu_Tillet/ex02_oo_basics/src/circle.h
```diff
@@ -0,0 +1,25 @@
+/* Circle shape */
+
+#ifndef CIRCLE_H
+#define CIRCLE_H
+
+#include<cmath>
+
+#include "shape.h"
```
gvardany, line 8: Normally your includes should be first, then the system includes.

exercises/Mathieu_Tillet/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,40 @@
+/**
+ * Skeleton main routine
+ */
+
+#include <iostream>
+
+#include "square.h"
```
gvardany, line 7: For the classes we use CamelCase convention.

exercises/Mathieu_Tillet/ex02_oo_basics/src/shape.cpp
```diff
@@ -0,0 +1,5 @@
+#include "shape.h"
+
+ostream& operator<<(ostream &os, Shape &sh){
```
gvardany, line 3: Here it could be `const Shape &sh`

exercises/Mathieu_Tillet/ex02_oo_basics/src/shapesorter.cpp
```diff
@@ -0,0 +1,47 @@
...
+void ShapeSorter::print_by_type(vector<Shape*> &vec, string type, ostream &os){
+    for (vector<Shape*>::iterator it = vec.begin(); it != vec.end(); ++it){
+        if ((*it)->shape_type == type){
+            os << *(*it) << endl;
+        }
+    }
+}
+
+
+bool perimeter_comp(Shape *s1, Shape *s2){
```
gvardany, line 24: Those can be private.

exercises/Mathieu_Tillet/ex02_oo_basics/src/shapesorter.cpp
```diff
@@ -0,0 +1,47 @@
+#include "shapesorter.h"
+
+ShapeSorter::ShapeSorter(){}
```
gvardany, line 3: General remarks for the methods here: for the names we use lower case camel case, so `printBySides`, then for the arguments that are const, state those explicitly in the signature (for example the `type` in print by type), for printing you can use the range based loop, and normally we also require doxygen comments as developer documentation.

exercises/Mathieu_Tillet/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,26 @@
...
+    int sides;
+
+    Shape(){}
+    virtual ~Shape(){}
+    virtual double perimeter()=0;
+    virtual double area()=0;
+    virtual ostream& display(ostream &os)=0;
+};
+
+ostream& operator<<(ostream &os, Shape &sh);
```
gvardany, line 24: 👍

gvardany (commented):
Looks good overall, few remarks on how to improve:

PR #57, opened 2020-01-15
Adam washington exercises

rprospero:


exercises/template/ex01_basics/src/main.cpp
```diff
@@ -2,6 +2,120 @@
...
+
+using namespace std;
+
+bool fileMissing(const string &f) {
+  // Determine if a file is missing
+  ifstream test(f);
+  return !test.good();
+}
+
+bool valid_char(const char x) {
```
LamarMoore: Non-standard naming of function names. Please take a look at the Coding Standards [here](http://developer.mantidproject.org/Standards/index.html)

LamarMoore: There is already a function designed for this in the standard http://www.cplusplus.com/reference/cctype/ispunct/

exercises/template/ex01_basics/src/main.cpp
```diff
@@ -2,6 +2,120 @@
...
+  if (a.first < b.first) {
+    return true;
+  }
+  return false;
+}
+
+string string_strip(const string &word) {
+  // Remove invalid characters from a string
+  string result;
+  copy_if(word.begin(), word.end(), back_inserter(result), valid_char);
```
LamarMoore: As I said above, there is a function for this. However, we prefer the use of lambdas as opposed to free functions.

exercises/template/ex01_basics/src/main.cpp
```diff
@@ -2,6 +2,120 @@
...
+  return result;
+}
+
+map<string, int> add_word(map<string, int> acc, const string &base_word) {
```
LamarMoore: take a non-const reference to `acc` instead of making so many copies.

exercises/template/ex01_basics/src/main.cpp
```diff
@@ -2,6 +2,120 @@
...
+#include <algorithm>
+#include <fstream>
+#include <iostream>
+#include <iterator>
+#include <map>
+#include <numeric>
+#include <sstream>
+#include <vector>
+
+using namespace std;
```
LamarMoore: typically we do not do this. We use the `std` prefix e.g `std::accumulate` so that we can identify the use of the standard versus other libraries like `boost` or `tbb` for example.

exercises/template/ex01_basics/src/main.cpp
```diff
@@ -2,6 +2,120 @@
...
+  case '`':
+  case ':':
+  case '_':
+    return false;
+  default:
+    return true;
+  }
+}
+
+bool pair_compare(const pair<string, int> a, const pair<string, int> b) {
```
LamarMoore: use a lambda

exercises/template/ex01_basics/src/main.cpp
```diff
@@ -2,6 +2,120 @@
...
+  }
+
+  // Read all the files into a map
+  map<string, int> combined;
+  combined =
+      accumulate(arguments.begin(), arguments.end(), combined, load_file);
+
+  // Sort the map by occurance and word
+  auto result = vector<pair<string, int>>(combined.begin(), combined.end());
+  sort(result.begin(), result.end(), pair_compare);
```
LamarMoore: use a lambda instead of a free function. Also free functions should be in an anonymous namespace (just for future reference).

exercises/template/ex01_basics/src/main.cpp
```diff
@@ -2,6 +2,120 @@
...
+
+  // Display that results
+  for_each(result.begin(), result.end(), print_map_line);
```
LamarMoore: lambda

exercises/template/ex01_basics/src/main.cpp
```diff
@@ -2,6 +2,120 @@
...
+  if (a.second < b.second) {
+    return false;
+  }
+  if (a.first < b.first) {
+    return true;
+  }
+  return false;
+}
+
+string string_strip(const string &word) {
```
LamarMoore: can `word` be modified in place?

exercises/template/ex01_basics/src/main.cpp
LamarMoore: Also you should capture by reference instead of copying.

LamarMoore (commented):
Reviewed exercise 1. If you haven't already would be a good idea to familiarise yourself with coding standards and other developer documentation found [here](http://developer.mantidproject.org/)

exercises/template/ex02_oo_basics/src/shape.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+public:
+  explicit Square(double s) : side(s) {}
+  int sides() const override;
+  double perimeter() const override;
+  double area() const override;
+  string type() const override;
+};
+
+class Rectangle : public Shape {
```
LamarMoore, line 46: In general we prefer separate `*.h` `*.cpp` files for classes but as this is an exercise don't bother changing it.

exercises/template/ex02_oo_basics/src/shape.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include <iostream>
+#include <math.h>
+#include <memory>
+#include <vector>
+
+using namespace std;
+
+class Shape {
+public:
+  virtual int sides() const = 0;
```
LamarMoore: unsigned int or short

exercises/template/ex02_oo_basics/src/shape.hpp
```diff
@@ -0,0 +1,85 @@
...
+  virtual double perimeter() const = 0;
+  virtual double area() const = 0;
+  virtual string type() const = 0;
+};
+
+ostream &operator<<(ostream &os, const Shape *shape);
+
+class Circle : public Shape {
+private:
+  double radius;
```
LamarMoore: prefix `m_` for class member variables.

exercises/template/ex02_oo_basics/src/shape.cpp
```diff
@@ -0,0 +1,76 @@
...
+  // print shapes with given number of sides
+  for_each(contents.begin(), contents.end(),
+	   [side_count](const unique_ptr<Shape> &value) {
+	     if (value->sides() == side_count)
+	       cout << value << endl;
+	   });
+}
+void ShapeSorter::typed(const string &type) const {
+  // print shapes of a given type
+  for_each(contents.begin(), contents.end(), [type](const unique_ptr<Shape> &value) {
```
LamarMoore: capture `type` by reference.

exercises/template/ex02_oo_basics/src/shape.cpp
```diff
@@ -0,0 +1,76 @@
...
+ostream &operator<<(ostream &os, const unique_ptr<Shape> &shape) {
+  // Print the important parts of a shape
+  os << "A " << shape->type();
+  os << " with " << shape->sides() << " sides";
+  os << ", a permieter of " << shape->perimeter();
+  os << ", and and area of " << shape->area();
+  return os;
+};
+
+int Circle::sides() const { return 1; }
```
LamarMoore: prefer constants to hard-coded values

exercises/template/ex02_oo_basics/src/shape.hpp
```diff
@@ -0,0 +1,85 @@
...
+  explicit Triangle(double b, double h) :base(b), height(h) {}
+  int sides() const override;
+  double perimeter() const override;
+  double area() const override;
+  string type() const override;
+};
+
+class ShapeSorter {
+private:
+  vector<unique_ptr<Shape>> contents;
```
LamarMoore: `contents` isn't a good name. Maybe something like `m_shapes`

exercises/template/ex02_oo_basics/src/shape.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+public:
+  explicit ShapeSorter(vector<unique_ptr<Shape>> &c) : contents(move(c)){};
```
LamarMoore: use `vector<unique_ptr<Shape>> &&c` also, do we need to use pointers?

exercises/template/ex02_oo_basics/src/shape.cpp
```diff
@@ -0,0 +1,76 @@
+#include "shape.hpp"
+
+#include <algorithm>
+#include <iostream>
+
+ostream &operator<<(ostream &os, const unique_ptr<Shape> &shape) {
```
LamarMoore, line 10: You could implement streaming for each shape so we could see the radius for circles or base and height for triangles for example. However I am not too bothered.

exercises/template/ex02_oo_basics/src/main.cpp
```diff
@@ -2,6 +2,29 @@
...
+#include <vector>
+
+#include "shape.hpp"
+
+
+using namespace std;
+
+int main(int, char **) {
+  vector<unique_ptr<Shape>> shapes;
+  shapes.push_back(unique_ptr<Shape>(new Circle(5)));
```
LamarMoore: `make_unique`? Again, are pointers necessary here?

exercises/template/ex02_oo_basics/src/main.cpp
```diff
@@ -2,6 +2,29 @@
...
+  shapes.push_back(unique_ptr<Shape>(new Square(5)));
+  shapes.push_back(unique_ptr<Shape>(new Rectangle(1, 11)));
+  shapes.push_back(unique_ptr<Shape>(new Triangle(5, 6)));
+
+  ShapeSorter sorter(shapes);
```
LamarMoore, line 21: You will need to call `move` here is you use move semantics in the constructor in the example I provided below.

LamarMoore (commented):
Overall a good solution for exercise 2. Just a few minor comments.

exercises/template/ex02_oo_basics/src/shape.hpp
rprospero: I've looked into trying to avoid using a pointer here, but I'm not sure how to do that while still allowing the vector to be heterogeneous for all of the subclasses of Shape.  The best alternate solutions that I came up with were doing either a union or a variant.  The variant had the obvious downside of requiring C++17.  Both solutions had the ugliness of needing to add in each child class as it is added.

I'm likely missing something, but it looks like we do need pointers here if multiple different types will go into the same vector.

exercises/template/ex01_basics/src/main.cpp
```diff
@@ -2,6 +2,97 @@
...
+  }
+
+  void stringStrip(std::string &word) {
+    // Remove invalid characters from a string
+    word.erase(std::remove_if(word.begin(), word.end(),
+			      [](char letter){return ispunct(letter);}),
+	       word.end());
+  }
+
+  std::map<std::string, int> addWord(std::map<std::string, int> acc, std::string word) {
```
LamarMoore, line 30: if you take a const reference you will not need to copy or return `acc`

LamarMoore: `word` should be a const ref

LamarMoore (commented):
Minor points but I am happy over-all.

PR #58, opened 2020-03-12
Guillaume communie exercises

gui-co:


exercises/guillaume_communie/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <cctype>
+#include <iomanip>
+#include <set>
+
+const std::string RESULTS_FILENAME("word_counts.data");
+
+/**
+ * Print the program usage.
+ */
+void usage(void) {
```
gvardany: I don't think you have to specify void for the argument list, you can as well leave it blank.

exercises/guillaume_communie/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,93 @@
...
+  std::cout << "Usage: WordCounter <file_name>" << std::endl;
+  exit(1);
+}
+
+/**
+ * Check if a characters is an 'interword' character.
+ * @param[in] c The character to test
+ * @return      True if the character is an interword character
+ */
+bool isInterWord(int c) {
```
gvardany: You can store these symbols in a string, then check if you can find the char in it, this might be nicer.

exercises/guillaume_communie/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,93 @@
...
+    // try to open the file
+    std::ifstream f(argv[1]);
+    if (!f.good()) {
+        std::cout << "Error: please provide a valid file path." << std::endl;
+        usage();
+    }
+
+    // read the words
+    std::map<std::string, int> dictionnary;
+    int character;
```
gvardany: Can you not have char?

exercises/guillaume_communie/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,93 @@
...
+    std::string word;
+    do {
+        character = f.get();
+        if (isalpha(character)) {
+            word.push_back((char) tolower(character));
```
gvardany: It's a good idea to specify the namespace for the algorithms so that there is no ambiguity.

exercises/guillaume_communie/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,93 @@
...
+          } else if (isInterWord(character)) {
+              if ((!word.empty()) && (word.size() > 4)) {
+                  auto it = dictionnary.find(word);
```
gvardany: Not sure if you have to really find this. You can try just `dicitionary[word]++`?

exercises/guillaume_communie/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,93 @@
...
+                           return e1.first < e2.first;
+                       else
+                           return e1.second > e2.second;
+                   });
+
+    // print the results
+    std::ofstream out_file(RESULTS_FILENAME);
+    out_file << std::setw(15) << std::left;
+    out_file << "Words" << "Usage" << std::endl;
+    for (auto it = sorted_dic.begin() ; it != sorted_dic.end() ; it++) {
```
gvardany: You should be able to use the range based loop. If not, since you are not modifying the set you could do `cbegin` and `cend`. Also it is better to do pre-increment `++it`.

exercises/guillaume_communie/ex02_oo_basics/inc/Circle.h
```diff
@@ -0,0 +1,43 @@
+#ifndef CIRCLE_H
```
gvardany: For information, we just moved to `#pragma once` instead of header guards.

exercises/guillaume_communie/ex02_oo_basics/inc/Circle.h
```diff
@@ -0,0 +1,43 @@
...
+     * Create a circle by providing its radius
+     * @param[in] r The circle radius
+     */
+    Circle(double r);
+
+    /**
+     * Get the perimeter
+     * @return The circle perimeter
+     */
+    double getPerimeter(void);
```
gvardany: Again, you don't have to specify void.

exercises/guillaume_communie/ex02_oo_basics/inc/Circle.h
```diff
@@ -0,0 +1,43 @@
...
+     * Print the circle
+     * @return A string that contains the circle data
+     */
+    std::string print(void);
+
+private:
+    /**
+     * Store the circle radius
+     */
+    double radius;
```
gvardany: For member variables we use the prefix `m_` to make it explicit.

exercises/guillaume_communie/ex02_oo_basics/inc/Shape.h
```diff
@@ -0,0 +1,66 @@
...
+     * Create a shape by a given type name and a number of sides
+     * @param[in] t The type name
+     * @param[in] n The number of sides
+     */
+    Shape(std::string t, int n);
+
+    /**
+     * Delete a shape
+     */
+    //~Shape(void);
```
gvardany: You should have a virtual distructor, since it is an abstract class.

exercises/guillaume_communie/ex02_oo_basics/inc/Shape.h
```diff
@@ -0,0 +1,66 @@
...
+     * Get the number of sides
+     * @return The shape number of sides
+     */
+    int getSides(void);
+
+    /**
+     * Get the perimeter
+     * @return The shape perimeter
+     */
+    virtual double getPerimeter(void) = 0;
```
gvardany: For the overrides of the virtuals, you can make them explicit with `override` keyword.

exercises/guillaume_communie/ex02_oo_basics/inc/ShapeSorter.h
```diff
@@ -0,0 +1,44 @@
...
+ * A shape sorter
+ */
+class ShapeSorter {
+public:
+    /**
+     * Print shape that match a certain type
+     * @param[in] v The vector of shapes
+     * @param[in] t The type to match
+     */
+    static void printMatchType(std::vector<Shape*> v, std::string t);
```
gvardany: Composite object is better to pass by reference to avoid copy. If they are not going to be modified, const& is even more explicit.

exercises/guillaume_communie/ex02_oo_basics/inc/Shape.h
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Get the area
+     * @return The shape area
+     */
+    virtual double getArea(void) =0;
```
gvardany: the getters could be marked as const methods

exercises/guillaume_communie/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,40 @@
+#include "ShapeSorter.h"
```
gvardany, line 1: A general comment for this class. You are doing std::cout in each method, which is the same line more or less repeated. It would be better to have a method that receives a vector and prints all of it, and then the filter methods, which will filter the input vector based on the corresponding criteria and then pass over the printer method. You could also use std::find_if or similar for the filters.

exercises/guillaume_communie/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,40 @@
...
+            std::cout << (*it)->print() << std::endl;
+}
+
+void ShapeSorter::printMatchSides(std::vector<Shape*> v, int n) {
+    for (auto it = v.begin() ; it != v.end() ; it++)
+        if ((*it)->getSides() == n)
+            std::cout << (*it)->print() << std::endl;
+}
+
+void ShapeSorter::printSortedByPerimeter(std::vector<Shape*> v) {
```
gvardany, line 25: You are sorting the original vector, but should do a copy; while this should print sorted, it should not modify the original vector.

gvardany (commented):
Looks pretty good, just few improvements to keep in mind.

exercises/guillaume_communie/ex01_basics/src/main.cpp
gui-co: I used the return type of `get()`. I changed to a `char`

exercises/guillaume_communie/ex01_basics/src/main.cpp
gui-co: I was not sure that `dicitionary[word]` would assign 0 to the `int` but it seems to be the case.

exercises/guillaume_communie/ex01_basics/src/main.cpp
gui-co: I change to the range based loop.

exercises/guillaume_communie/ex02_oo_basics/inc/ShapeSorter.h
gui-co: I changed to `const &` for the first two methods. I kept the copy for the others (see below).

exercises/guillaume_communie/ex02_oo_basics/src/ShapeSorter.cpp
gui-co: I am sorting a copy of the vector. I changed the documentation of the two functions in accordance.

exercises/guillaume_communie/ex02_oo_basics/src/ShapeSorter.cpp
gui-co: * I added a method to print a vector of shapes
* for the second point, I do not really know how to do it. For the first method (`printMatchType`) I used `std::copy_if` to create a new vector and print it with the previously mentioned method. Otherwise, `std::find_if` would have to be called several times to print all the solution and not the first one (?)

exercises/guillaume_communie/ex01_basics/src/main.cpp
gui-co: Here I am using `isalpha()` and `tolower()` from `ctype.h`

exercises/guillaume_communie/ex02_oo_basics/src/ShapeSorter.cpp
gvardany: Indeed find finds only the first one.

PR #59, opened 2020-04-09
Dominik Arominski exercises

darominski:
Reading file, raising run_time exceptions on failure to read a file, making counting words case-insensitive, insensitive to punctuation and hyphenation, sorting in descending order and saving into a file

### DIFF
--- a/exercises/dominik_arominski/ex01_basics/src/main.cpp
+++ b/exercises/dominik_arominski/ex01_basics/src/main.cpp
@@ -1,0 +1,82 @@
+#include <fstream>
+#include <string>
+#include <map>
+#include <vector>
+#include <algorithm>
+
+std::string RemovePunctuation(std::string word) {
+  const char punctuation[] = {'.', ',', '?', '!', '"', '\'', ')', '('};
+  for (auto punc : punctuation) {
+    if(word.find(punc) < 100) {
+      word = word.substr(0, word.find(punc)) +  word.substr(word.find(punc)+1, word.size());
+    }
+  }
+  return word;
+}
+
+std::vector<std::string> RemoveHyphenation(std::string const word) {
+  std::string const word1 = word.substr(0, word.find("-"));
+  std::string const word2 = word.substr(word.find("-")+1, word.size());
+  return {word1, word2}; 
+}
+
+int main(int argn, char **argc)
+{
+  if (argn < 2) {
+    throw std::runtime_error("File name not provided.");
+  }
+  else if (argn > 2) {
+    throw std::runtime_error("More than one input file has been provided. Please provide only one file name.");
+  }
+  
+  std::fstream inputFile;
+  inputFile.open(argc[1], std::fstream::in);
+  if (!inputFile) {
+    throw std::runtime_error("Could not open input file.");
+  }
+
+  std::map<std::string, int> wordMap;
+  std::string word;
+  while(inputFile >> word) {
+    if (word.size() < 4 ) {
+      continue;
+    }
+    word = RemovePunctuation(word);
+    if (word.find("-") < 100) {
+      auto newWords = RemoveHyphenation(word);
+      for (auto newWord : newWords) {
+	if (newWord.size() < 4)
+	  continue;
+	std::transform(newWord.begin(), newWord.end(), newWord.begin(), [](const unsigned char c){ return std::tolower(c); });
+	wordMap[newWord] += 1;
+      }
+    }
+    else {
+      std::transform(word.begin(), word.end(), word.begin(), [](const unsigned char c){ return std::tolower(c); });
+      wordMap[word] += 1;
+    }
+  }
+    
+  std::vector<std::pair<std::string, int>> wordVector;
+  for (auto itr = wordMap.begin(); itr != wordMap.end(); ++itr)
+    wordVector.push_back(*itr);
+  
+  sort(wordVector.begin(), wordVector.end(), [=](std::pair<std::string, int>& a, std::pair<std::string, int>& b) {
+	 return a.second > b.second;
+    });
+
+  std::fstream outputFile;
+  outputFile.open("results.dat", std::fstream::out);
+  if (!outputFile) {
+    throw std::runtime_error("Could not open output file.");
+  }
+  outputFile << "Word\tUsage" << std::endl;
+  
+  for (auto elem : wordVector) {
+    outputFile << elem.first << "\t" << elem.second << std::endl;
+  }
+  
+  inputFile.close();
+  outputFile.close();
+  return 0;
+}
--- a/exercises/dominik_arominski/ex01_basics/src/main.cpp~
+++ b/exercises/dominik_arominski/ex01_basics/src/main.cpp~
@@ -1,0 +1,64 @@
+/**
+ * Skeleton main routine
+ */
+
+#include <iostream>
+#include <fstream>
+#include <string>
+#include <iostream>
+#include <map>
+#include <set>
+#include <algorithm>
+#include <functional>
+
+std::string RemovePunctuation(std::string word) {
+  const char punctuation[] = {'.', ',', '?', '!', '"', '\'', ')', '('};
+  for (auto punc : punctuation) {
+    if(word.find(punc) < 100) {
+      word = word.substr(0, word.find(punc)) +  word.substr(word.find(punc)+1, word.size());
+    }
+  }
+  return word;
+}
+
+
+int main(int argn, char **argc)
+{
+  if (argn < 2) {
+    throw std::runtime_error("File name not provided.");
+  }
+  else if (argn > 2) {
+    throw std::runtime_error("More than one input file has been provided. Please provide only one file name.");
+  }
+  
+  std::fstream inputFile;
+  inputFile.open(argc[1], std::fstream::in);
+  if (!inputFile) {
+    throw std::runtime_error("Could not open file");
+  }
+
+  std::map<std::string, int> wordOccurences;
+  std::string word;
+  while(inputFile >> word) {
+    if (word.size() < 4 ) {
+      continue;
+    }
+    word = RemovePunctuation(word);
+    
+    std::transform(word.begin(), word.end(), word.begin(), [](const unsigned char c){ return std::tolower(c); });
+    wordOccurences[word] += 1;
+  }
+    
+  typedef std::function<bool(std::pair<std::string, int>, std::pair<std::string, int>)> Comparator;
+  Comparator compFunctor = [](std::pair<std::string, int> elem1, std::pair<std::string, int> elem2) {
+    return elem1.second < elem2.second;
+  };
+  
+  std::set<std::pair<std::string, int>, Comparator> orderedMap(wordOccurences.begin(), wordOccurences.end(), compFunctor);
+ 
+  for (std::pair<std::string, int> element : orderedMap)
+    std::cout << element.first << " : " << element.second << std::endl;
+ 
+
+  return 0;
+}
--- a/exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
+++ b/exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
@@ -1,7 +1,71 @@
-/**
- * Skeleton main routine
- */
+#include <iostream>
+#include <string>
+#include <cmath>
 
-int main(int, char **)
-{
+class Shape {
+
+protected:
+  std::string type;
+  int nSides;
+
+public:
+  
+  std::string GetType() { return this->type;};
+  int GetSides() { return this->nSides;};
+  
+  virtual double CalculatePerimeter() {};  
+  
+};
+
+class Triangle : public virtual Shape {
+private:
+  double height;
+  double base;
+public:
+  double CalculatePerimeter() {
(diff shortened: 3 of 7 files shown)


exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
+#include <fstream>
```
gvardany, line 1: It is a good idea to sort the includes alphabetically.

exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <string>
+#include <map>
+#include <vector>
+#include <algorithm>
+
+std::string RemovePunctuation(std::string word) {
```
gvardany, line 7: For the functions and methods, we use lower camel case, so `removePunctuation`.
For passing the arguments, anything that is composite object, including strings, should be passed by & or const & to avoid copy. If you pass by &, there is no point to return.

exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+  const char punctuation[] = {'.', ',', '?', '!', '"', '\'', ')', '('};
```
gvardany, line 8: This can nowadays be static constexpr, you could also take this out of the function body, and make a constant outside. In real code this should also be wrapped by anonymous namespace, to avoid any potential conflicts when linking.

exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+  for (auto punc : punctuation) {
+    if(word.find(punc) < 100) {
```
gvardany, line 10: Can you explain what's the reason for this check?

exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+      word = word.substr(0, word.find(punc)) +  word.substr(word.find(punc)+1, word.size());
+    }
+  }
+  return word;
+}
+
+std::vector<std::string> RemoveHyphenation(std::string const word) {
```
gvardany, line 17: Here could be const &.

exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  std::map<std::string, int> wordMap;
+  std::string word;
+  while(inputFile >> word) {
+    if (word.size() < 4 ) {
+      continue;
+    }
+    word = RemovePunctuation(word);
+    if (word.find("-") < 100) {
```
gvardany, line 45: Again the check, is this in the requirements of the problem?

exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+      auto newWords = RemoveHyphenation(word);
+      for (auto newWord : newWords) {
+	if (newWord.size() < 4)
+	  continue;
+	std::transform(newWord.begin(), newWord.end(), newWord.begin(), [](const unsigned char c){ return std::tolower(c); });
+	wordMap[newWord] += 1;
+      }
+    }
+    else {
+      std::transform(word.begin(), word.end(), word.begin(), [](const unsigned char c){ return std::tolower(c); });
```
gvardany, line 55: These 2 lines are almost the same as the lines above, it could be better organised here.

exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+      wordMap[word] += 1;
```
gvardany, line 56: `wordMap[word]++` is shorter.

exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+    
+  std::vector<std::pair<std::string, int>> wordVector;
+  for (auto itr = wordMap.begin(); itr != wordMap.end(); ++itr)
```
gvardany, line 61: `const auto`

exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+    wordVector.push_back(*itr);
+  
+  sort(wordVector.begin(), wordVector.end(), [=](std::pair<std::string, int>& a, std::pair<std::string, int>& b) {
```
gvardany, line 64: Make explicit `std::sort`, also since C++ 14 you can have generic lambdas, so you could do const auto& for the arguments.

exercises/dominik_arominski/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+    });
+
+  std::fstream outputFile;
+  outputFile.open("results.dat", std::fstream::out);
+  if (!outputFile) {
+    throw std::runtime_error("Could not open output file.");
+  }
+  outputFile << "Word\tUsage" << std::endl;
+  
+  for (auto elem : wordVector) {
```
gvardany, line 75: Again, const auto& to avoid a copy.

exercises/dominik_arominski/ex01_basics/src/main.cpp~
```diff
@@ -0,0 +1,64 @@
+/**
```
gvardany, line 1: What is this file?

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,154 @@
+#include <iostream>
+#include <string>
+#include <cmath>
+#include <vector>
+#include <algorithm>
+
+class Shape {
+
+protected:
+  std::string type;
```
gvardany, line 10: For the member variables we use `m_` prefix in Mantid.

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+  int nSides;
+
+public:
+  
+  std::string GetType() const { return type;};
```
gvardany, line 15: Again, lower camel case for the methods: `getType()`

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+  int GetSides() const { return nSides;};
+  
+  virtual double CalculatePerimeter() const { return 0;};
```
gvardany, line 18: Shouldn't these be pure virtual, i.e. = 0?

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+public:
+  
+  std::string GetType() const { return type;};
+  int GetSides() const { return nSides;};
+  
+  virtual double CalculatePerimeter() const { return 0;};  
+  virtual double CalculateArea() const { return 0;};  
+
+  friend std::ostream& operator<<(std::ostream& out, const Shape& shape) {
```
gvardany, line 21: Why friend?

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+    return out << shape.GetType() << "\t" << shape.GetSides() << "\t" << shape.CalculateArea() << "\t" << shape.CalculatePerimeter();
+  }
+  
+};
+
+class Triangle : public virtual Shape {
```
gvardany, line 27: Just `public Shape` would do.

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+  double CalculatePerimeter() const {
+    return 2.0 * M_PI * radius;
+  }
+  double CalculateArea() const {
+    return M_PI * pow(radius, 2);
+  }
+};
+
+class Rectangle : public virtual Shape {
+protected:
```
gvardany, line 64: Those were private for the other shape, I see that you intent to subclass for the Square, but it would be nice to have all the shapes consistent in this sense.

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+public:
+  Square(double fSide=0) : Rectangle(fSide) {
+    side2 = side1;
+    type = "Square";
+  }
+};
+
+class ShapeSorter {
+
+public:
```
gvardany, line 92: For these methods, it would be nicer to separate the printing from the sorting/filtering. Whether the sorting should order the original vector, or make a copy, is not specified in the exercise, so you can decide.

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+private:
+  double height;
+  double base;
+public:
+  Triangle(double fHeight, double fBase) : height(fHeight), base(fBase) {
+    type = "Triangle";
+    nSides = 3;
+  }
+  
+  double CalculatePerimeter() const {
```
gvardany, line 37: If the base method is pure virtual, you can make explicit by adding `override` keyword here.

gvardany (commented):
Good job, just few details to be aware of for the real codebase.

exercises/dominik_arominski/ex01_basics/src/main.cpp
darominski: It was meant to find whether the punctuation character was in the string. I believe it would be better if the check was: if(word.find(punc) != std::string::npos).

exercises/dominik_arominski/ex01_basics/src/main.cpp
darominski: Agreed.

exercises/dominik_arominski/ex01_basics/src/main.cpp
darominski: Agreed and will take note of that for the future.

exercises/dominik_arominski/ex01_basics/src/main.cpp
darominski: Agreed.

exercises/dominik_arominski/ex01_basics/src/main.cpp
darominski: The check should be different, as said before: if (word.find("-") != std::string::npos). The check for hyphenation was in the requirements.

exercises/dominik_arominski/ex01_basics/src/main.cpp
darominski: Agreed.

exercises/dominik_arominski/ex01_basics/src/main.cpp
darominski: Agreed, then the for would need to slightly change form to: for (const auto itr : wordMap).

exercises/dominik_arominski/ex01_basics/src/main.cpp
darominski: Agreed, will take note.

exercises/dominik_arominski/ex01_basics/src/main.cpp
darominski: Agreed.

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
darominski: Agreed,

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
darominski: You are correct, it does not have to be friend of this class as it accesses only public fields and functions of this class. I removed it from the body of the class.

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
darominski: Agreed.

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
darominski: I will take note.

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
darominski: Indeed, that would be correct and better.

exercises/dominik_arominski/ex02_oo_basics/src/main.cpp
darominski: I will take note.

PR #60, opened 2020-07-20
Add project command to remove CMake warning

DannyHindson:
CMake has started producing a warning since v3.16 if you don't specify a project command at the top of the master CMakeLists.txt file:
CMakeWarning

So thought I'd resolve this to avoid any confusion.
Also moved the minimum required version on to match the version required by the CMakeLists in the main Mantid repository

### DIFF
--- a/exercises/template/CMakeLists.txt
+++ b/exercises/template/CMakeLists.txt
@@ -1,4 +1,6 @@
-cmake_minimum_required ( VERSION 2.8.5 )
+cmake_minimum_required ( VERSION 3.14 )
+
+project(newstarter)
 
 add_subdirectory ( ex01_basics )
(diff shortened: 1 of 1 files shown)


DannyHindson:
@martyngigg - I've noticed just now there's a CMake warning has appeared when running CMake on the new starter project. So have put this PR in for the change. Not sure if I'm supposed to put a PR? The RHEL and Windows jobs failed with a "no projects" error

martyngigg:
@DannyHindson Thanks for spotting this. A PR is fine. Although as this is not mission critical I'm happy with minor changes like this directly.

The builds expect a directory to have been created that is a copy of the template so the failure is expected but it's not a problem in this case. I'll merge this.

PR #61, opened 2020-07-21
Excercise 1 completed

tolu28-coder:


exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,184 @@
+
+#include <iostream>
+#include <vector>
+#include <fstream>
+#include <string>
+#include <iomanip>
+using namespace std;
```
DannyHindson: Please can you remove this "using" statement.
We don't use "using namespace std" to avoid name clashes between functions in the Mantid code and those in the standard library. It means the code is a bit more verbose because you need to add the std:: prefix in various places but it's a bit safer.

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+}
+
+
+void removePunctuation(string &stringName) {
+	/* Function accepts a string a modifies string by
+	removing any occurence of ".,?'"!():" within word 
+	and returns nothing
+	*/
+	for (int i = 0; i < stringName.size();) {
```
DannyHindson: This works fine but could be made a bit more concise by using the std::erase and std::remove functions from the standard library eg
          for (int i = 0; i < strlen(CHARS_TO_REMOVE); i++) {
            stringName.erase(std::remove(stringName.begin(), stringName.end(), CHARS_TO_REMOVE[i]), stringName.end());

You should be able to find some examples\explanation of this online. The std::remove call doesn't actually remove the chars - it just moves them to the back of the string ready to be deleted by the std::erase

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,184 @@
...
+		stream.get(InputCharacter);
+		if (InputCharacter == '-' || InputCharacter == ' ' || InputCharacter == '\n'||stream.eof())
+			break;
+		OutputString += InputCharacter;
+	}
+	return OutputString;
+}
+
+
+int findString(const string &SearchString, const vector <string> &SearchList) {
```
DannyHindson: There's a "find" function in the standard library that does this for you so I think you could remove the findString function eg
auto it = std::find(VectorOfWords.begin(),VectorOfWords.end(),HolderString)
if (HolderString.size() > 4 && it==VectorOfWords.end()){
}

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,184 @@
...
+		else {
+			if (HolderString.size() > 4 && location != -1)
+			{
+				VectorOfCounts.at(location) += 1;
+		}
+		}
+	}
+	FileToRead.close();
+
+	for (int i = 0; i < VectorOfWords.size(); i++) {
```
DannyHindson: I quite like what you've done here in many ways in that it's quite a novel way of generating the sorted list of counts. However I'm going to ask you to change it for 2 reasons:
a) the double loop is quite slow because the two vectors you're looping over are ~1500 elements each
b) there are already standard functions in the standard library for sorting vectors that mean you don't have to expend as much thought on this part
Have a looking at the standard library "sort" function. I think to use it here you will perhaps need to reformulate the VectorOfWords and VectorOfCounts into a single vector containing pairs

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,184 @@
...
+#include <iomanip>
+using namespace std;
+
+
+void toLower(string &stringName) {
+	/* Function accepts a string and modifies string
+	by making it lowercase and returns nothing
+	*/
+	string PlaceHolderString{  };
+	for (char Letter : stringName)
```
DannyHindson: Two rather pedantic points relating to this range based for loop:
- we tend to use auto in these ie for (auto Letter : stringName). If you've not come across the auto keyword let's discuss
- where the iterator isn't being changed we normally use & ie for (auto& Letter : stringName)

In this example both changes will have minimal functional impact but in a different case where the iterator is a bigger class then it would improve performance.

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,184 @@
...
+#include <string>
+#include <iomanip>
+using namespace std;
+
+
+void toLower(string &stringName) {
+	/* Function accepts a string and modifies string
+	by making it lowercase and returns nothing
+	*/
+	string PlaceHolderString{  };
```
DannyHindson: Please can you use camelCase for variable names ie placeHolderString. I'll just highlight this one but please can you change other variable names in this unit as well

DannyHindson (changes requested):
Hi Tolu, a good first attempt. I think a few of my comments perhaps originate from the fact that you didn't have access to the same C++ book that the others had. That book had some chapters which describes some of the standard library functions I've referred to in my comments. Aside from the detail level comments here are a couple of general ones:
1) please can you change your editor to insert 2 spaces whenever you hit the tab key. This is part of our coding standards. So the main.cpp unit shouldn't have any tab characters in it
2) many people when answering this question using the std::map class because it provides a convenient way of storing a key (in this case the word) and a value (in this case the count). Take some time to have a look at that class. It does have one downside which is that you can't sort it.  I'm fine if you leave this using the vector class instead but worth looking up this class

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,137 @@
...
+      {
+        vectorOfCounts[std::distance(vectorOfWords.begin(), location)] += 1;
+      }
+    }
+  }
+  fileToRead.close();
+  int maxCount{ 0 };
+  /*Next Block of code implements a variation of counting
+  sort  on vectorOfCounts to produce filled vectorOfOrderToRead*/
+  for (auto& count : vectorOfCounts) {
```
DannyHindson: Thanks for making the change here - now much faster.

When most people do the sorting at this stage of the exercise they use the std::sort function in the standard library. I can see you've used a counting sort method here instead and I may be persuaded on this point - please could you let me know why this might be better than using the std::sort function?

Even if we stick with your version it would be good practice to make it easy to swap in the std::sort function at some point in the future. Plus this would also fulfil part of the point of the exercise which is to demonstrate you know how to call the std::sort function. So with this in mind please could you do these changes:

a) pull the counting sort code out into a separate function called countingSort or similar
b) the std::sort function would require a single vector containing the words\counts. So please could you replace these two vectors with a vector of pairs:
```  
  std::vector <std::string> vectorOfWords{};
  std::vector <int> vectorOfCounts{};
```
The "pairs" vector would be:
```
  std::vector <std::pair<std::string,int>> vectorOfWords{};
```
I think your counting sort code can be adapted to work with this arrangement

c) please can you write a second sorting function called stdSort or similar that uses the std::sort function. This function can be left unused once I've checked on your reasoning re counting sort

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+
+
+std::string getWord(std::ifstream& stream) {
+  /* Function accept an input stream object and return a string from
+  input stream by reading stream 1 character at a time  and formats it  
+  by making it lowercase and removing punctuation marks then 
+  concatenates it to a string until a space ,newline or '-'  character.
+  */
+  std::string outputString{};
```
DannyHindson: This comment doesn't relate to anything new you've changed in this 2nd round of commits but just thought of it now.
In general in the Mantid codebase we don't tend to put the empty braces around the declaration of a string because "std::string outputString" has the same effect ie default constructor is called and you end up with an empty string. This isn't explicity documented in our coding standard but perhaps comes under the general approach of matching the style of the code that's already there. 
So please can you remove the braces from this and any other string declarations?

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <vector>
+#include <fstream>
+#include <string>
+#include <iomanip>
+#include <algorithm>
+
+
+
+std::string getWord(std::ifstream& stream) {
+  /* Function accept an input stream object and return a string from
```
DannyHindson: Please can you move this comment above the function signature and add a couple of lines to describe the parameter and return value. This will make it follow the Doxygen convention and is described here in the Mantid coding standards:

FunctionHeaders

DannyHindson (changes requested):
Hi Tolu, thanks for making the changes for my first round of comments. I've added a few more there but think we're getting there. The main comment is around the sorting and the relative merits of std::sort vs the counting sort algorithm you've implemented

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -35,99 +68,83 @@ int main() {
...
   if (!fileToRead) {
     /* This if block checks if file was opened correctly
-    and if it was not program is terminated
+    and if not program is terminated
     */
     std::cout << "File could not be opened" << std::endl;
     std::cout << "Ensure that file path is valid" << std::endl;
   }
+  std::string holderString;
+  auto myPredictor = [&holderString](std::pair<int, std::string>& pairOfCountAndWord)
```
DannyHindson: Should this be myPredicator rather than myPredictor?

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -35,99 +68,83 @@ int main() {
...
     */
-    std::string holderString{ };
+    
     holderString = getWord(fileToRead);
-    auto location = std::find(vectorOfWords.begin(), vectorOfWords.end(), holderString);
-    if (holderString.size() > 4 && location == vectorOfWords.end()) {
-      vectorOfWords.push_back(holderString);
-      vectorOfCounts.push_back(1);
+    auto location = std::find_if(vectorOfCountsAndWords.begin(),vectorOfCountsAndWords.end(), myPredictor);
+    if (holderString.size() > 4 && location == vectorOfCountsAndWords.end()) {
```
DannyHindson: Please can you pull the if size>4 check out into a surrounding if statement so you don't need to repeat it here and L107. I think the check on the location can then be turned into an if...else statement without any conditions on the else part.

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -35,99 +68,83 @@ int main() {
...
-    if (count > maxCount) {
-      maxCount = count;
-    }
-  }
-  std::vector <std::vector <int>> vectorForSort;
-  for (int i = 0; i < maxCount; i++) {
-    std::vector <int> emptyVector;
-    vectorForSort.push_back(emptyVector);
-  }
+  //countingSort(vectorOfCountsAndWords);
```
DannyHindson: Please can you add a comment just above this line to say that alternative counting sort method has been implemented that is faster than most implementations of the std::sort algorithm. This will indicate why you've left it commented out
I say "most implementations" because the C++ standard library doesn't mandate how std::sort has to be implemented so in theory the Microsoft version (for example) could use counting sort - but I know in reality this isn't the case.

DannyHindson (changes requested):
Thanks for the last round of changes. Looks good - I've made ~3 v minor further comments and then think we're done

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+    VectorOfWords until the end of file has been reached. This block
+    has an if , else if block that determines whether word is more than
+    4 letters and already exists in VectorOfWords. if it does it increments
+    the count in VectorOfCounts otherwise it is added to VectorOfWords
+    */
+    
+    holderString = getWord(fileToRead);
+    auto location = std::find_if(vectorOfCountsAndWords.begin(),vectorOfCountsAndWords.end(), myPredictor);
+    if (holderString.size() > 4 && location == vectorOfCountsAndWords.end()) {
+      vectorOfCountsAndWords.push_back(std::make_pair(1,holderString));
```
DannyHindson: Just thought of one other thing that we're trying to encourage. We are in general trying to use the function emplace_back instead of push_back when adding items to vectors. The reason is that it has better performance when temporary objects are added as in this case.
To use emplace back to construct an object in place you can just pass the arguments to the constructor of the pair:
vectorOfCountsAndWords.emplace_back(1,holderString);
If you've not pushed your latest set of changes yet please could you make this change also?

DannyHindson (changes requested):
One more comment that came up in another review that I'll repeat here

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -99,20 +99,23 @@ int main() {
     */
     
     holderString = getWord(fileToRead);
-    auto location = std::find_if(vectorOfCountsAndWords.begin(),vectorOfCountsAndWords.end(), myPredictor);
-    if (holderString.size() > 4 && location == vectorOfCountsAndWords.end()) {
-      vectorOfCountsAndWords.push_back(std::make_pair(1,holderString));
+    auto location = std::find_if(vectorOfCountsAndWords.begin(),vectorOfCountsAndWords.end(), myPredicator);
+    if (holderString.size() > 4){
```
DannyHindson, line 103: Please can you move the find_if line inside the if statement where you check the size of holderString. Don't think you need to do the call to find_if for words < 4

exercises/Toluwalase_Agoro/ex01_basics/src/main.cpp
```diff
@@ -99,20 +99,23 @@ int main() {
...
+    if (location == vectorOfCountsAndWords.end()) {
+      vectorOfCountsAndWords.emplace_back(1,holderString);
     }
     else {
-      if (holderString.size() > 4 && location != vectorOfCountsAndWords.end())
-      {
+
```
DannyHindson: Please can you remove this blank line

DannyHindson (changes requested):

DannyHindson (approved):
All looks good now, well done.
One minor thing that would stop it passing the Jenkins jobs on the real Mantid repository is that the indentation is a bit irregular around lines L105-110. This is best corrected by installing\running a clang-format tool either as a git command or depending on which development tool you're using you may be able to set up a tool in there.

martyngigg:

PR #62, opened 2020-07-21
Matthew Cumber Exercise 1 Completed

Matt-Cumber:
Here is my solution to exercise 1.


exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include <iomanip>
+#include <algorithm>
+#include <fstream>
+#include <string>
+#include <vector>
+#include <ctype.h>
+#include "UniqueWord.h"
+
+// Default path for output file reading files from
+std::string const PATH = "/Users/Matt/STFC Placement/C++ Introduction/newstarter/exercises/matthew_cumber/ex01_basics/";
```
DannyHindson: I like the fact you've pulled this out as a constant. Couple of comments on it:
1) we normally declare string constants like this using constexpr which tells the compiler that the value is known at compile time
2) we also normally declare them in an anonymous namespace which means the constant's scope is limited to this code unit
So something like:
namespace {
  constexpr auto PATH= "/Users/Matt/STFC Placement/etc";
}
I've used auto here instead of the std::string because C++ doesn't let you use string with constexpr so it allows me to cover up a slightly more nasty type - which I think would be "constexpr char PATH[]" or something. I believe you can use std::string with constexpr with C++20...

Also please can you change this so your user home path isn't included. Not sure which OS you're on (Mac?) but maybe you can use ~ or similar to refer to your home area. This is partly to avoid it being specific to your user and partly to make it work cross platform

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+/** Tests if a word has already been found the increments it's count
+ * @param uniqueWords The vector of existing unique words found
+ * @param word The word that is being tested
+ * @return A boolean, true if a word already exists, false otherwise
+ */
+bool inUniqueWords(std::vector<UniqueWord> &uniqueWords,std::string word)
```
DannyHindson: Since you're not changing the 2nd "word" parameter you can pass this by reference as well for better performed ie const std::string& word

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  return false;
+}
+
+/** Writes the unique words to an output file "UniqueWords.txt"
+ * @param uniqueWords The vector of unique words to be written to the file
+ * @return no return value
+ */
+void writeUniqueWordsToFile(std::vector<UniqueWord> const uniqueWords)
```
DannyHindson: You should pass the parameter by reference for better performance. Also as a matter of style\consistency I suggest moving the const should be before the std::vector (but I think that doesn't have any functional difference):

ie const std::vector<UniqueWord>& uniqueWords

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+int main(int argc, char const **argv)
+{
+  std::vector<UniqueWord> uniqueWords; // Stores words and counts as objects in a vector
+
+	// Check a file path has been provided as an argument on cl
+	if(argc < 2)
+	{
+		std::cout << "Please provide a path to the file to be read as a command line argument." << std::endl;
```
DannyHindson: Something funny has happened to the indenting here. Next time we get together it's worth discussing how to set up your development environment to apply clang code formatting that will do this for you. Generally speaking you can either apply it in the IDE or you can run a custom git command to apply the formatting before doing a commit

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,199 @@
...
+    for(int i = 0; i < word.length(); ++i)
+    {
+      // Erase letter if punctuation
+      while(isPunctuation(word[i]))
+      {
+        word.erase(i,1);
+        if(i >= word.length()) break; // Check not at end of word
+      }
+
+      if(i >= word.length()) break; // Check not at end of word
```
DannyHindson: Not sure this line is needed since if i was >= word.length() the code would have already done a "break" on L128?

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,199 @@
...
+    if(word.length() < 5) continue;
+
+    // Remove Punctuation, Convert all letters to lower case, and Remove hyphens
+    for(int i = 0; i < word.length(); ++i)
+    {
+      // Erase letter if punctuation
+      while(isPunctuation(word[i]))
+      {
+        word.erase(i,1);
+        if(i >= word.length()) break; // Check not at end of word
```
DannyHindson: This comment is a bit subjective - but it feels like there's a bit too much attempted in this loop through the characters of "word". I think it's functionally OK but it looks quite complicated and could be difficult to maintain.

Please can you have a go at rearranging this a bit so that you don't need this kind of "check not at end of word" line. Also would be nice if the logic to check the word length is >4 and then add the word into the list could be done in one place - currently it's repeated at L149 and L166

Perhaps you could manage the hyphens in their own loop rather than trying to do it in the same pass as removing punctuation\converting to lower case. You can split the word variable returned from "file>>word" into further subwords using the std::getline function. Something like:

        std::istringstream ss(word);
        std::string wordWithoutHyphens;
        while (std::getline(ss, wordWithoutHyphens, '-')) {
        }

DannyHindson (changes requested):
Hi Matt, this is a good first attempt. You've set up the class nicely. As discussed earlier today a lot of people use the std::map class instead. While we normally lean towards library functions where possible, you can't sort the map it's not a 100% solution either. So happy to go with your class based solution.
I've added a few detail-level comments next to the individual lines

DannyHindson:
One other thing. I just noticed that you've marked this pull request as draft. Normally if a pull request is ready for review you would remove the "draft" label. This is usually used while people are still working on the code and maybe they want to see if it passes the Jenkins tests for example.

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <algorithm>
+#include <fstream>
+#include <string>
+#include <sstream>
+#include <vector>
+#include <ctype.h>
+#include <regex>
+#include "UniqueWord.h"
+
+std::string PATH = ""; // Will store the path of the file to be read
```
DannyHindson, line 12: I can see you had a bit of "fun" trying to get the user home area to work in a cross platform way. When I made the comment I'd forgotten this wasn't that straightforward so apologies for sending you off on a bit of a tricky investigation there. For future reference we have a couple of ways of doing this in Mantid in these two units that you should be able to see in GitHub:
Framework\API\src\FileProperty.cpp (see function getHomePath)
Framework\Kernel\src\ConfigService.cpp (there's a function getAppDataDir that uses a 3rd party Poco library which provides a function Poco::Path::home)

Changing this to take a user input is fine and is what some of the others have done. One small thing - I think that since this variable is no longer a constant or a global variable it can be made a local variable in main rather than sitting up here

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+/** Tests if a word has already been found the increments it's count
```
DannyHindson, line 14: I salute your spelling corrections that I see you made elsewhere (I hardly dare search for occurence in the Mantid code base to check it we've all spotted this!). On a similar vein, please can you change to "...then increments its count"

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,145 @@
+#include <iostream>
+#include <iomanip>
+#include <algorithm>
+#include <fstream>
+#include <string>
+#include <sstream>
+#include <vector>
+#include <ctype.h>
+#include <regex>
+#include "UniqueWord.h"
```
DannyHindson, line 10: Please can you move the #include for your class above the system includes. This ordering is part of the Mantid coding standards.
Once you've installed a clang-format tool (either as git custom command or in your IDE) this kind of thing will get fixed automatically.

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+/** Tests if a word has already been found the increments it's count
+ * @param uniqueWords The vector of existing unique words found
+ * @param word The word that is being tested
+ * @return A boolean, true if a word already exists, false otherwise
+ */
+bool inUniqueWords(std::vector<UniqueWord> &uniqueWords, const std::string &word)
+{
+  // Loop unique words, if word exists increment it's count
+  for(auto &w : uniqueWords)
```
DannyHindson, line 22: Please can you change this to use the std::find function:

auto it = std::find(uniqueWords.begin(), uniqueWords.end(), word)
if (it!=uniqueWords.end()){
}

This is a similar quantity of code but it means it's easier to swap the type of uniqueWords over to one of the other STL container classes (list, map, queue etc) which may not support random access.
The potential for doing this in this specific situation isn't huge but as a piece of general practice this change may come in useful.

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,145 @@
...
+    // Now read all words hyphen separated
+    std::istringstream ss(word);
+    std::string wordWithoutHyphens;
+    while(std::getline(ss, wordWithoutHyphens, '-'))
+    {
+      // Skip words less than 5 characters
+      if(wordWithoutHyphens.length() < 5) continue;
+
+      // Else if not in found words already, create new word
+      if(!inUniqueWords(uniqueWords,wordWithoutHyphens))
```
DannyHindson, line 117: It's not v obvious that InUniqueWords also increments the count. Please could you rename the function to incrementWordCounts or similar and put the logic to handle a new word inside it?

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,145 @@
...
+  
+  std::cout << "Reading file \"" + PATH + "\"" << std::endl;
+
+  std::string word;
+
+  // Loop file word by word, reads characters separated by spaces
+  while(file >> word)
+  {
+    // Loop punctuation characters and using regex replace 1 or more occurrences with ""
+    for(const char &p : PUNCTUATION)
```
DannyHindson, line 99: nice rearrangement of this loop thanks. Much cleaner now 👍

exercises/matthew_cumber/ex01_basics/src/UniqueWord.cpp
```diff
@@ -0,0 +1,65 @@
+#include <iostream>
+#include <string>
+#include "UniqueWord.h"
```
DannyHindson, line 3: Please can you put the system includes beneath this UniqueWord.h include. That's one of the coding standards. This will be fixed automatically by the clang-format tool if you install that.

DannyHindson (changes requested):
Thanks for making the changes. I've added a few more minor comments but think we're pretty close now.

exercises/matthew_cumber/ex01_basics/src/main.cpp
Matt-Cumber: Yes I was struggling with passing the windows check as I guess I started to try and use a Unix based method of getting a users home directory. I will have a look at the unit tests you've mentioned, thanks.

DannyHindson (commented):
Just ran the code to double check it still works and spotted one more thing.

You currently use the std::regex_replace function which is fine functionally but seems a bit slow (on my laptop, your code takes a few secs to run rather than completing almost straight away and think it's because of this).
For replacing text we normally use a combination of the std::string::erase and std::remove functions like this:

my_str.erase(std::remove(my_str.begin(), my_str.end(), 'A'), my_str.end())

Worth replacing your regex_replace call with this and seeing if it's faster. The above is a fairly common technique in C++. It's slightly obscure because the remove function doesn't actually remove the characters - it just moves them to the end of the string and then erase gets rid of them.

PR #63, opened 2020-07-21
Initial commit for exercise 1

clayton-tom:


exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+#include <iostream>
+#include <sstream>
+#include <fstream>
+#include <algorithm>
+#include <cctype>
+#include <string.h>
+#include <vector>
+#include <map>
+using namespace std;
```
DannyHindson: Not sure if this is documented but we don't use "using namespace std". Although it's makes the code more verbose by having the std:: in a lot of places, it means there's not any danger of name clashes between names in the Mantid code and the standard library

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,158 @@
...
+  while (std::getline(tokenStream, token, delim)) {
+    if(token.size() < 4) {
+      continue;
+    }
+
+    std::transform(token.begin(), token.end(), token.begin(),
+    [](unsigned char c){ return std::tolower(c);});
+    
+    tokens.push_back(token);
+    //cout << token << endl;
```
DannyHindson: Please can you delete this. It's not unusual to have commented out stuff like this while developing but we try to remove this kind of thing before merging the code into master

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,158 @@
...
+  std::string token;
+  std::istringstream tokenStream(s);
+  
+  while (std::getline(tokenStream, token, delim)) {
+    if(token.size() < 4) {
+      continue;
+    }
+
+    std::transform(token.begin(), token.end(), token.begin(),
+    [](unsigned char c){ return std::tolower(c);});
```
DannyHindson, line 45: Nice use of the transform function

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/**
+ * Function to replace puncuation with whitespace in a string
+ */
+
+void depunctuate(std::string& line) {
+
+  std::replace_if(line.begin() , line.end(), 
+  [] (const char& c) { return std::ispunct(c) ;},' ');
```
DannyHindson: I'd not come across ispunct before.  
The list of characters that ispunct considers to be punctuation seems to be longer than the list specified in the exercise so please could you change this to use the shorter list in the exercise: .,?'"!():
Also the behaviour here for apostrophes isn't quite what I was expecting eg Engineer's should be converted to Engineers however this code ends up counting it as Engineer. Maybe the wording of the exercise isn't 100% clear but that's what the intention was...
Regardless of that - the approach of converting hyphens to spaces is nice so that you can just identifying words later on by using a space delimiter.

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,158 @@
...
+ */
+
+std::vector<std::string> splitByDelim(const std::string& s, char delim) {
+  
+  std::vector<std::string> tokens;
+  std::string token;
+  std::istringstream tokenStream(s);
+  
+  while (std::getline(tokenStream, token, delim)) {
+    if(token.size() < 4) {
```
DannyHindson: I think this should be <=4. The code should only count words that are longer than 4 characters

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/**
+ * Function to populate a map with key=word, value=no of
+ * occurrences for an input vector of single word strings
+ */ 
+
+void fillMap(std::map<std::string, int>& wordCountMap, std::vector<std::string>& words) {
+
+  typedef vector<std::string>::size_type vec_sz;
```
DannyHindson: Please can you use PascalCase for any typedefs ie Vec_Sz or better VecSz

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,158 @@
...
+  inFile.close();
+  return words;
+}
+
+/**
+ * Function to populate a map with key=word, value=no of
+ * occurrences for an input vector of single word strings
+ */ 
+
+void fillMap(std::map<std::string, int>& wordCountMap, std::vector<std::string>& words) {
```
DannyHindson: Please can you add const to the 2nd parameter because I don't think it's modified. Helps readability of what the function is doing

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,158 @@
...
+  std::cin >> fin;
+  return fin;
+}
+
+/**
+ * Function to split up a string and return a vector of words with
+ * len <4, and converted to lower case
+ */
+
+std::vector<std::string> splitByDelim(const std::string& s, char delim) {
```
DannyHindson: Please can you add const to the 2nd parameter

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,158 @@
...
+    outFile << freq_vect[i].second << endl;
+  }
+  outFile.close();
+}
+
+/**
+ * Main function to operate and test the program
+ */
+
+int main(int, char **)
```
DannyHindson, line 172: Nice small\clean main function

DannyHindson (changes requested):
Hi Tom - good first exercise, well done. Nice use of the standard library in various places and the code is easy to read and clean.
There seem to be a few files that are part of your commit than shouldn't be there eg the .json files. Probably best not to include the various .txt files either.
One way of excluding file types is to edit the .gitignore file. Might not be appropriate here though because there could be some circumstances where a .json file could be part of a commit.
I've included a few detail level comments against specific lines

exercises/tom_clayton/ex01_basics/src/main.cpp
clayton-tom: Thanks, removed and made sure std:: is inserted where needed

exercises/tom_clayton/ex01_basics/src/main.cpp
clayton-tom: Done

exercises/tom_clayton/ex01_basics/src/main.cpp
clayton-tom: Replaced this with a new function that determines whether a given character is to be removed, replaced with whitespace or left. specPunct(char c) in new commit

exercises/tom_clayton/ex01_basics/src/main.cpp
clayton-tom: Done

exercises/tom_clayton/ex01_basics/src/main.cpp
clayton-tom: Done

exercises/tom_clayton/ex01_basics/src/main.cpp
clayton-tom: Done

exercises/tom_clayton/ex01_basics/src/main.cpp
clayton-tom: Done

exercises/tom_clayton/ex01_basics/src/CMakeLists.txt
```diff
@@ -0,0 +1,1151 @@
+THE ADVENTURES OF SHERLOCK HOLMES
```
DannyHindson: This CMakeLists.txt seems to contain the content from Holmes.txt.
I don't think you need a CMakeLists.txt file at this level so this one can be deleted (not just deleted from Git but you can delete locally as well)

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,19 @@
+{
```
DannyHindson: For some reason the change to the .gitignore file doesn't seem to have removed these files as part of the pushes you've done. I think if you run the following git command for each file that needs removing from GitHub it should sort it out:

git rm -r --cached <filename>

Perhaps changes to .gitignore always require this manual patch up if you've already committed the files - can't remember

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include <cctype>
+#include <string.h>
+#include <vector>
+#include <map>
+
+/**
+ * Function to return a string filepath from user input
+ */
+
+std::string getInputFn() {
```
DannyHindson, line 18: Please can you change the format of this comment to match the format you used for specPunct. For that one you've used the Doxygen format required by the coding standards. There aren't any parameters here but I think you still need to describe the return parameter
Similar comment for splitByDelim, depunctuate functions also

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+  std::cout << "Please enter the path of the file to be analysed" << std::endl;
+  std::string fin;
+  std::cin >> fin;
+  return fin;
+}
+
+/**
+ * Function to split up a string and return a vector of words with
+ * len <4, and converted to lower case
```
DannyHindson: I'm getting pedantic now - but please can you update this to <= as well

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+ *  replaced with whitespace.
+ *  Specified punctuation to be removed: .,?'"!()
+ *  Specified punctuation to be replaced with whitespace: -
+ * 
+ *  @param c - Reference to the character in question
+ *  @return int 0 if non-punct, 1 if punct to remove, 2 if punct to
+ *  replace
+ */
+
+int specPunct(char c) {
```
DannyHindson: Rather than returning an int where the values have particular meanings it would be better to return an enum value. However I think that you can perhaps save a few lines of code by merging this function into the depunctuate function (see comments against that function)

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * Function to remove most punctuation, and replace hyphens with
+ * whitespace
+ */
+
+void depunctuate(std::string& line) {
+
+  std::replace_if(line.begin() , line.end(), 
+  [] (const char& c) { return specPunct(c)==2;},' ');
```
DannyHindson: Rather than calling specPunct here could you just replace this with:
```
  std::replace_if(line.begin(), line.end(),
    [](const char& c) { return c == '-'; }, ' ');
```

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+  line.erase(
+    std::remove_if(line.begin() , line.end(), 
+    [] (const char& c) { return specPunct(c)==1;}),
+  line.end());
```
DannyHindson, line 76: Similarly here I wonder if you could you just put this in a for loop that loops around the charRemove array and then perhaps you don't need the specPunct function at all?

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+  typedef std::vector<std::string>::size_type VecSz;
+  VecSz words_size = words.size();
+  if(words_size == 0){
+    throw std::domain_error("Empty words vector");
+  }
+
+  for(VecSz i = 0; i < words_size; i++) {
+    std::string currentWord = words[i];
+    wordCountMap[currentWord] ++;
+  }
```
DannyHindson, line 120: You could shrink this a litte by writing:
```
  for (auto& currentWord : words) {
    wordCountMap[currentWord] ++;
  }
```
Might avoid you needing to declare the typedef if you just checked if (words.size()==0) higher up

DannyHindson (changes requested):
Thanks for making the changes. I've come back with a few more comments but looking good and probably my final ones.

.vscode/settings.json
```diff
@@ -0,0 +1,44 @@
+{
```
DannyHindson: I see that you managed to remove one of the files in the .vscode folder in one of your recent commits. Not the end of the world but if you could also remove this one that would be good - wasn't v clear in my earlier comments I meant both

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+std::string getInputFn() {
+  std::cout << "Please enter the path of the file to be analysed" << std::endl;
+  std::string fin;
+  std::cin >> fin;
+  return fin;
+}
+
+/** Split up a string and return a vector of words with
+ *  len <=4, and converted to lower case
```
DannyHindson: In my previous comment I asked you to change this from len <4 to len <=4
I now see that it should read "...return a vector of words with len >4". Please can you update

exercises/tom_clayton/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+/** Returns different integer values to classify an input char as
+ *  either non-punctuation, puntuation to be removed or to be
+ *  replaced with whitespace.
+ *  Specified punctuation to be removed: .,?'"!()
+ *  Specified punctuation to be replaced with whitespace: -
+ * 
+ *  @param c - Reference to the character in question
+ *  @return int 0 if non-punct, 1 if punct to remove, 2 if punct to
+ *  replace
```
DannyHindson: Now that you've merged the original specPunt function into depunctutate, I don't think you need the comment on lines 53-61 any more?

DannyHindson (changes requested):
Thanks for the last round of changes. Have added ~3 v minor comments (one file removal and two adjustments to comments) and then think we're done.

clayton-tom:
That should be the lot! Also removed the .txt files as I'd thought that including them in the .gitignore would remove them on the next commit, didn't realise they needed removing manually

DannyHindson (approved):
All looks good now thanks, nice work!

martyngigg:

PR #64, opened 2020-07-21
Finished first exercise

joseph-torsney:
Finished the first exercise to count the usages of words in an input text file


exercises/joseph_torsney/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+    c == '('
+  );
+}
+
+/** Reads the contents of a .txt file at filepath
+ * 
+ * @param filepath the path to the file
+ * @return std::string of the file contents
+ */
+std::string readFile(const std::string filepath)
```
DannyHindson: You should pass this string parameter by reference since you're not changing (better performance) ie const std::string& filepath

exercises/joseph_torsney/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  return(ret);
+}
+
+/** Counts the number of times a word appears (usage) in a given string vector.
+ * 
+ * @param words the string vector to count
+ * @return std::map<std::string, int> maps each word to the number of usages.
+ */
+std::map<std::string, int> countWords(std::vector<std::string> words) {
```
DannyHindson: Please can you pass words by const ref for performance ie const std::vector<std::string>& words

exercises/joseph_torsney/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+      return(a.second > b.second);
+    }
+  );
+
+  return(v);
+}
+
+int main(int argc, const char** argv)
+{ 
+  std::string text = readFile(argv[1]);
```
DannyHindson: Please can you check argc is 2 (or at least 2) before trying to access argv[1]

exercises/joseph_torsney/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+ * 
+ * @param words the string vector to count
+ * @return std::map<std::string, int> maps each word to the number of usages.
+ */
+std::map<std::string, int> countWords(std::vector<std::string> words) {
+  std::map<std::string, int> usage;
+  std::map<std::string, int>::iterator it;
+
+  for (int i = 0; i < words.size(); i++) {
+    it = usage.find(words[i]);
```
DannyHindson: I think you can replace the contents of this loop by simply:
usage[words[i]]++
I think the map manages whether the entry exists or not for you

DannyHindson (changes requested):
Hi Joe, have run out of time a bit today but here are a few comments to be going on with. Looks like a good effort overall, well done. I'll come back to this on Thurs when I'm back

exercises/joseph_torsney/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/** Removes punctuation from a string
+ * 
+ * @param str the string to process
+ * @return std::string with no punctuation
+ */
+std::string removePunct(std::string str)
+{
+  for (int i = 0, len = (int) str.size(); i < len; i++) {
```
DannyHindson: I think this for loop could be made a bit more simple because C++ evaluates the condition before each iteration. So no need to reset len below on L68 and you can just write:
```
for (int i = 0; i < str.size(); i++) {
```
However rather than doing that change you can do something even more concise by using the remove_if function which means you don't need a for loop at all:
```
str.erase(std::remove_if(str.begin(), str.end(), &myIsPunct), str.end());
```

exercises/joseph_torsney/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,185 @@
...
+  // process characters of s
+  while (i != s.size()) {
+
+    // ignore whitespace
+    while(i != s.size() && (isspace(s[i]) || s[i] == '-')) {
+      i++;
+    }
+
+    // otherwise, we are at the start of a word.
+    string_size j = i;
```
DannyHindson: This loop performs as required functionally but in many cases it's possible to avoid relying on any complicated checks on indices in loops by using the c++ library a bit more.
In this case if you choose to read in the filestream word by word rather than doing it all at once in readFile you get a marker for free in the filestream which is equivalent to "j"
For example if you just open the filestream in readFile then there are a couple of approaches you can follow to simplify this function:
a) the >> operator respects the same logic as isspace and gives you the content of the file word by word. To manage the hyphens you then need to have an inner loop that uses the getline function. So something like:
```
while (filestr>>word){
  std::istringstream ss(wordWithoutHyphens);
  getline(filestr,wordWithoutHyphens,'-'){
    \\build up vector of words in here...
  }
}
```
b) if you first replace all hypens with a space then you don't need to bother with the getline inner loop

exercises/joseph_torsney/ex01_basics/src/main.cpp
DannyHindson: Don't think you've made this change yet? Just in case I wasn't clear I think you can just do:
```
  for (int i = 0; i < words.size(); i++) {
    usage[words[i]]++;
  }
```

DannyHindson (changes requested):
Thanks for making the changes based on my review on Tues. Here are a few more now I've had a bit more time to look through but I think we're pretty close

DannyHindson (changes requested):
Just noticed one more thing - don't think you've implemented the requirement to only count words that are more than 4 characters. Should be just a one line change somewhere near the counting

exercises/joseph_torsney/ex01_basics/src/main.cpp
joseph-torsney: Thought I had but I was messing around with git so that's probably why!

exercises/joseph_torsney/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,129 @@
...
+    c == '('
+  );
+}
+
+/** Transforms a string to lower case.
+ * 
+ * @param str the string to transform
+ * @return std::string the string in lower case
+ */
+std::string toLowerStr(std::string str) {
```
DannyHindson, line 33: Please can you pass the parameter as const ref ie const std::string&

exercises/joseph_torsney/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+  return str;
+}
+
+/** Removes punctuation from a string
+ * 
+ * @param str the string to process
+ * @return std::string with no punctuation
+ */
+std::string removePunct(std::string str)
```
DannyHindson, line 46: Please can you pass the parameter as const ref ie const std::string&

exercises/joseph_torsney/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,129 @@
...
+ * 
+ * @param map the map to sort
+ * @return std::vector<std::pair<std::string, int>> sorted key, value pairs from map.
+ */
+std::vector<std::pair<std::string, int>> sortByValue(std::map<std::string, int> map) 
+{
+  // declare and create a vector of pairs
+  std::vector<std::pair<std::string, int>> v;
+  for (auto& it : map) {
+    v.push_back(it);
```
DannyHindson, line 93: In general on Mantid we're trying to use the emplace_back function rather than push_back when adding entries into vectors. This is because it has better performance in certain situations (when inserting a temporary object). This isn't one of those situations but we're taking the view that we'll use emplace_back anyway just so we don't have to think on each and every case

exercises/joseph_torsney/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,129 @@
...
+  // count and sort the usage map into vector of pairs
+  auto usageVector = sortByValue(usage);
+
+  std::ofstream outdata;
+  outdata.open(argv[2]);
+
+  outdata << "Word\tUsage\n" << std::endl;
+  // print everything out using the iterator :)
+  for (auto it : usageVector) {
+    outdata << it.first << "\t" << it.second << std::endl;
```
DannyHindson, line 127: Minor thing but you're output isn't nicely aligned for some of the longer words. Perhaps you could use the std::setw(n) to set the width of the first column?
image

DannyHindson (changes requested):
Looking good! In the home straight now I think. Just 3-4 v minor things that I've commented on but most of heavy lifting done thanks

exercises/joseph_torsney/ex01_basics/src/main.cpp
joseph-torsney: How do I do that here, since I'm calling .erase on str?

DannyHindson: I think what I meant to say was that it would be more efficient to change the function signature to:

void removePunct(std::string& str)

Same adjustment needed to my other comment on toLowerStr.

If you do these two changes you won't be able to nest the calls to these 2 functions any more where you manipulate wordWithoutHyphens so you gain a line of code. But you won't have any string copies happening which means the code will be faster so I think that's worth it.

At this point you could do away with the two functions entirely and just put the contents of removePunct\toLowerStr in wordUsageFromFile. However the std::transform and std::erase syntax is quite dense so having the functions makes it easy to read.

martyngigg:

PR #65, opened 2020-07-22
Matthew cumber exercise 2

Matt-Cumber:


exercises/matthew_cumber/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,22 @@
...
+public:
+  /// Constructor of circle
+  Circle(double _radius);
+
+  /// implementation of area and perimeter for a circle (virtual in Shape class)
+  double area();
+  double perimeter();
+
+private:
+  double radius;
```
alicerussell1: For non-static member variables we use the prefix `m_` so in this case it would be `m_radius`.
http://developer.mantidproject.org/Standards/CPPStandards.html#naming-conventions

exercises/matthew_cumber/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,22 @@
...
+ * radius : Stores the radius of the circle
+ */
+class Circle : public Shape
+{
+public:
+  /// Constructor of circle
+  Circle(double _radius);
+
+  /// implementation of area and perimeter for a circle (virtual in Shape class)
+  double area();
```
alicerussell1: When overriding a virtual function you can use the [override](https://en.cppreference.com/w/cpp/language/override) keyword. This means the compiler can check if any parameters are incorrect and doesn't override as intended.

exercises/matthew_cumber/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,24 @@
...
+ * type : The type of shape determined by derived class
+ * numSides : The number of sides the shape has determined by derived class
+ */
+class Shape
+{
+public:
+  virtual ~Shape() = default;
+
+  /// methods to be implemented in derived classes
+  virtual double area() = 0;
```
alicerussell1: Since this  method is just getting the area and shouldn't be changing the shape we can make the method `const`. So this would become `virtual double area() const = 0;`

exercises/matthew_cumber/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,93 @@
+#pragma once
+
+/** A function to compare the area of two shapes
+ * @param s1 The first shape
+ * @param s2 The second shape
+ * @return A boolean true if the area of s1 < area of s2, false otherwise
+ */
```
alicerussell1, line 7: Great job with the documentation!

exercises/matthew_cumber/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,30 @@
+#include <iostream>
+#include <string>
```
alicerussell1: We should avoid including things that are not required in the file. It's not noticeable here but for big projects like Mantid this can slow down compile time

exercises/matthew_cumber/ex02_oo_basics/src/Triangle.cpp
```diff
@@ -0,0 +1,34 @@
...
+  return (height*base)/2.0;
+}
+
+/** Calculates the perimeter of a triangle
+ * @return Value of perimeter
+ */
+double Triangle::perimeter()
+{
+  double baseSquaredOverFour = pow(base,2)/4.0;
+  double root = sqrt(pow(height,2) + baseSquaredOverFour);
```
alicerussell1: These can be `const`. I would generally make all variable const and then remove it when needed

alicerussell1 (changes requested):
This is a good solution to exercise 2 - well done 👍 
I've left some comments below, note that some of them apply to all the shapes however I have only written them once

exercises/matthew_cumber/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,27 @@
...
+
+/*
+ * Abstract class for all shapes for common members and methods
+ * m_type : The type of shape determined by derived class
+ * m_numSides : The number of sides the shape has determined by derived class
+ */
+class Shape
+{
+public:
+  virtual ~Shape() = default;
```
alicerussell1, line 13: We don't need to include the default destructor here as it should already be set to default

exercises/matthew_cumber/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,93 @@
...
+
+template <class T> class ShapeSorter
+{
+public:
+
+  /** A method to print only shapes of only a given type
+   * @param shapes A vector of pointers of shape objects
+   * @param type A string which is the type of shapes to be printed
+   */
+  void findType(const std::vector<std::unique_ptr<T>> &shapes, const std::string type)
```
alicerussell1: It is best to pass strings by reference as it is more efficient

exercises/matthew_cumber/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,93 @@
...
+
+  /** A method to print the area of the shapes in a vector in descending order
+   * @param shapes A vector of pointers of shape objects
+   * @return No return value
+   */
+  void printOrderArea(std::vector<std::unique_ptr<T>> &shapes)
+  {
+    // Sort shapes by area then reverse order to get descending order
+    std::sort(shapes.begin(), shapes.end(), compareArea);
+    std::reverse(shapes.begin(), shapes.end());
```
alicerussell1: Rather than have `reverse` here to get it into descending order you could just switch the sign in compareArea to be greater than. (similarly for perimeter)

alicerussell1 (changes requested):
This is looking good, just a couple of comments below that I didn't notice last time

exercises/matthew_cumber/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    return;
+  }
+
+  /** A method to print only shapes with a given number of sides
+   * @param shapes A vector of pointers of shape objects
+   * @param numSides The number of sides a shape must have to be printed
+   */
+  void findSides(const std::vector<std::unique_ptr<T>> &shapes, const int &numSides)
```
alicerussell1, line 47: Just a quick comment here, int doesn't need to be passed by reference in this case. The main reason for passing by reference is so that the value of the variable doesn't need to copied. So in this case an int isn't going to cost much to copy whereas strings can be large.
There's no need to change this, just so you are aware

alicerussell1 (approved):
I'm happy with this now, it's a good solution is the exercise 👍

I've written one comment below but there's no need to change it

PR #67, opened 2020-07-23
Finished exercise 2

joseph-torsney:
Finished exercise 2 and created a new branch for this pull request


exercises/joseph_torsney/ex02_oo_basics/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# Common setup
+include ( CommonSetup )
+
+###############################################################################
+# Executable
+###############################################################################
+
+set ( SRC_FILES
+  src/main.cpp
+  # add any additional source files here
```
alicerussell1, line 27: Make sure whenever you add any new files you add them to the `CMakeLists`. This file should include all the headers and cpp files. It looks as though we are missing circle, square and triangle

exercises/joseph_torsney/ex02_oo_basics/src/circle.h
```diff
@@ -0,0 +1,24 @@
+#ifndef CIRCLE_H
+#define CIRCLE_H
```
alicerussell1: Up until a few months ago in Mantid we used these header guards, however we have now switched to using `#pragma once` at the top of the header file instead

exercises/joseph_torsney/ex02_oo_basics/src/circle.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <math.h>
+#include "shape.h"
+
+class Circle: public Shape {
+  double radius;
```
alicerussell1: Although this is private by default, it is better to use the private assessor here.

exercises/joseph_torsney/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "circle.h"
+#include "triangle.h"
+#include "shapeSorter.h"
+
+#include <iostream>
+#include <vector>
+
+int main(int, char **)
+{
+  Rect* myRect = new Rect(10, 200); // why do these have to be pointers?
```
alicerussell1: Good question! The issue comes when creating the vector of the shapes as you can not have a vector of objects of an abstract class. This is because when it compiles the vector needs to know the size of the object which is does by creating the object - this can't be done for abstract cases. So we use a pointer which has a fixed amount of memory.

exercises/joseph_torsney/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,28 @@
...
+  Square* mySquare = new Square(10);
+  Circle* myCircle = new Circle(10);
+  Triangle* myTriangle = new Triangle(10, 200);
+
+  ShapeSorter* sorter = new ShapeSorter(
```
alicerussell1: When using raw pointers we have to be careful not to leave the pointer around and create a memory leak. To do this you will want to delete the pointer when you have finished by doing something like `delete sorter`.

However there is an safer and easier way. You can also use smart pointers which handle this for you so you don't have to worry about it. (These don't appear in the book as they were introduced in c++ 11)

exercises/joseph_torsney/ex02_oo_basics/src/rect.h
```diff
@@ -0,0 +1,25 @@
+#ifndef RECT_H
+#define RECT_H
+
+#include "shape.h"
+
+class Rect: public Shape {
+  double side1;
```
alicerussell1: The [mantid ](http://developer.mantidproject.org/Standards/CPPStandards.html#naming-conventions)standard is to use the prefix `m_` before member variables e.g. `m_side1`

exercises/joseph_torsney/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,39 @@
...
+
+    int getSides() {
+      return sides;
+    }
+
+    std::string getType() {
+      return type;
+    }
+
+    virtual double getArea() { return 0; }
```
alicerussell1: It would be better to make this a pure virtual function. This would mean that we wouldn't be able to create a subclass of shape without overriding this function.
This is done by adding an `=0` e.g.
`virtual double getArea() = 0`

exercises/joseph_torsney/ex02_oo_basics/src/shapeSorter.h
```diff
@@ -0,0 +1,76 @@
...
+  public:
+    ShapeSorter(std::vector<Shape*> v) {
+      shapes = v;
+    }
+  
+    /** Prints the shape objects that match a given type
+     * 
+     * @param type the type of shapes to print
+     */
+    void printMatching(std::string type) {
```
alicerussell1: It is best to pass strings by reference as it is more efficient e.g.
` void printMatching(const std::string &type) `

exercises/joseph_torsney/ex02_oo_basics/src/shapeSorter.h
```diff
@@ -0,0 +1,76 @@
...
+        }
+      }
+    }
+
+    /** Prints the shape objects that have a given number of sides
+     * 
+     * @param sides the number of sides of shapes to print
+     */
+    void printSides(int sides) {
+      for (auto it = shapes.begin(); it != shapes.end(); it++) {
```
alicerussell1: This can be written as a [range-based for loop](https://en.cppreference.com/w/cpp/language/range-for)

exercises/joseph_torsney/ex02_oo_basics/src/square.h
```diff
@@ -0,0 +1,14 @@
+#ifndef SQUARE_H
+#define SQUARE_H
+
+#include "rect.h"
+
+class Square: public Rect {
```
alicerussell1: It is not quite correct to have a square inherit from a rectangle.  It would be better to inherit from a shape. In the mathematical case a square is a rectangle however in a coding sense it can lead to errors. Say someone added a stretch method to rectangle which add 10 to the width. This would compile and run with a square however it would no longer be a square

exercises/joseph_torsney/ex02_oo_basics/src/shape.h
alicerussell1: These methods can also be `const`

exercises/joseph_torsney/ex02_oo_basics/src/rect.h
```diff
@@ -0,0 +1,25 @@
...
+  double side2;
+
+  public:
+    Rect(double a, double b): Shape("rect", 4) {
+      side1 = a;
+      side2 = b;
+    }
+
+    double getArea() {
```
alicerussell1: As these methods are overriding a virtual function you can add the [`override`](https://en.cppreference.com/w/cpp/language/override) keyword to the function definition which will check at compile time if any parameters are incorrect and doesn't override as intended.

exercises/joseph_torsney/ex02_oo_basics/src/shapeSorter.h
```diff
@@ -0,0 +1,76 @@
...
+      printAll();
+    }
+
+    /** Prints the shapes in descending order of perimeters
+     * 
+     */
+    void printPerimeters() {
+      std::sort(shapes.begin(), shapes.end(), 
+        [](const auto& a, const auto& b) -> bool {
+          return a->getPerimeter() > b->getPerimeter();
```
alicerussell1: Good use of a lambda and the standard library 👍

alicerussell1 (changes requested):
This is a great start, it works well. I've written a few comments below, it would also be good in include some documentation for each method.

The main thing to be aware of how we split up the header and cpp files. In the header file you would expect to find the class definitions and variables and within the cpp you would then see the implementation of the class. For example, a header may look like this
```
class MyClass {
public:
	# constructor
	MyClass(int x);
	# any public functions or variables
	int Foo(int num);
private:
	#any private functions or variables
	int myVariable;
}
```
and then in the cpp we would have:
```
#constructor
MyClass(int x) : myVariable(x) {}

#implementation of functions
int MyClass::Foo(int num){
	return num;
}
```

Let me know if you have any questions.

exercises/joseph_torsney/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "triangle.h"
+#include "shapeSorter.h"
+
+#include <iostream>
+#include <vector>
+#include <memory>
+
+int main(int, char **)
+{
+  std::shared_ptr<Rect> myRect(new Rect(10, 200) );
```
alicerussell1: In the mantid standards we try and use `make_shared` rather than `new` if possible

exercises/joseph_torsney/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,29 @@
...
+  std::shared_ptr<Square> mySquare(new Square(10));
+  std::shared_ptr<Circle> myCircle(new Circle(10));
+  std::shared_ptr<Triangle> myTriangle(new Triangle(10, 200));
+
+  ShapeSorter* sorter = new ShapeSorter(
```
alicerussell1: `ShapeSorter` is still a raw pointer, is there a reason it can't be a smart pointer or was it just missed off?

alicerussell1: Or does it even need to be a pointer?

exercises/joseph_torsney/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,22 @@
...
+
+class Shape {
+  public:
+    Shape(std::string t, int s);
+
+    int getSides();
+
+    std::string getType();
+
+    const virtual double getArea() = 0;
```
alicerussell1: This may have been a miscommunication on my part but the `const` here does not make method `const` rather it makes sure the return value is `const` (which for a build in datatypes like double it does nothing). 
To make the method `const` you will want to put `const` after the brackets
 i.e. `virtual double getArea() const=0;`
Then this will ensure that shape is not changed when calling this method.

exercises/joseph_torsney/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,22 @@
...
+
+    const virtual double getPerimeter() = 0;
+
+    std::string toString();
+
+  protected:
```
alicerussell1, line 19: Can these be private?

exercises/joseph_torsney/ex02_oo_basics/src/shapeSorter.cpp
```diff
@@ -0,0 +1,49 @@
+#include "shapeSorter.h"
+#include "shape.h"
+#include <iostream>
+#include <vector>
+#include <algorithm>
+
+ShapeSorter::ShapeSorter(std::vector<std::shared_ptr<Shape>> v) : m_shapes(v) {};
+
+void ShapeSorter::printMatching(const std::string &type) {
+  for (auto it : m_shapes) {
```
alicerussell1: Since we are not changing the shapes it is more efficient to get the value by reference e.g.
`for (const auto &it : m_shapes) `

exercises/joseph_torsney/ex02_oo_basics/src/square.cpp
```diff
@@ -0,0 +1,3 @@
+#include "square.h"
+
+Square::Square(double s) : Shape("square", 4), m_side1(s) {};
```
alicerussell1: You don't need to change this but just so you are aware if the constructor is simple like this it is okay to keep it in the header

exercises/joseph_torsney/ex02_oo_basics/src/triangle.h
```diff
@@ -0,0 +1,18 @@
...
+  public:
+    Triangle(double b, double h);
+
+    const double getArea() override;
+
+    const double getPerimeter() override;
+
+  private:
+    double m_base;
+    double m_height;
```
alicerussell1: Since these values don't change after initialization these can be `const`

alicerussell1 (changes requested):
This is looking good, I've just added couple more small comments

exercises/joseph_torsney/ex02_oo_basics/src/square.h
```diff
@@ -0,0 +1,16 @@
...
+
+  public:
+    Square(double side);
+
+    double getArea() const override;
+
+    double getPerimeter() const override;
+
+  private:
+    const int m_side;
```
alicerussell1: I think this should be a double since the variable is the length of the side. This is causing the warning on the windows build as it is trying to convert the inputted double to an int here.

alicerussell1 (approved):
I'm happy with this now - good work, well done 👍

martyngigg:

PR #68, opened 2020-07-23
Toluwalase agoro exercise2

tolu28-coder:
Exercise 2 completed


exercises/Toluwalase_Agoro/ex02_oo_basics/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+###############################################################################
+
+set ( SRC_FILES
+  src/main.cpp
+  # add any additional source files here
+)
+
+set ( HDR_FILES
+  # add any additional header files here
+)
```
alicerussell1: One thing to be aware of is within Mantid we use `CMake`. When we add any files we have to update files called `CMakeLists.txt` so CMake will look for them. So in here we will want to include all the files added for exercise 2.

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,244 @@
...
+#include <vector>
+#include <cmath>
+#include <iostream>
+#include <iomanip>
+#include <algorithm>
+
+
+class BaseShape {
+public:
+    int numberOfSides{};
```
alicerussell1: The empty braces here are not really needed, this does the same as `int numberOfSide`

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,244 @@
...
+    std::string type{};
+    float area{0.0};
+    float perimeter{ 0.0 };
```
alicerussell1: Well done for using direct initialization in this case 👍 It is probably not too noticeable in this situation as you are only printing the number however I would use a `double` rather than a `float` in most cases as it has a much higher precision

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,244 @@
...
+    virtual float calculateArea(void) { return 0.0; }
+    virtual float calculatePerimeter(void) { return 1.0; }
```
alicerussell1: I think this would be better if it was pure virtual as this would ensure that any class inheriting from it will have to override this method. This is done by adding `= 0`. It is also not necessary to pass `void` as a parameter you can just have a set of empty brackets. So overall this would become
`virtual float calculatePerimeter() = 0 `

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,244 @@
...
+    virtual float calculateArea(void) { return 0.0; }
+    virtual float calculatePerimeter(void) { return 1.0; } 
+
+    
+
+};
+
+class Square : public BaseShape {
+  public:
+    float calculateArea(void) override ;
```
alicerussell1: Good use of the `override` keyword 👍

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,244 @@
...
+
+  }
+  return perimeter;
+}
+
+class Rectangle : public BaseShape {
+public:
+  float calculateArea(void) override;
+  float calculatePerimeter(void) override;
+  float length;
```
alicerussell1: In the [mantid](http://developer.mantidproject.org/Standards/CPPStandards.html#naming-conventions) we use the prefix `m_` for member variables, e.g. `m_length`.

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,244 @@
...
+  return perimeter;
+}
+
+class ShapeSorter {
+  public: 
+    std::vector <BaseShape*> allShapes{};
+    std::vector <int> orderToReadArea;
+    std::vector <int> orderToReadPerimeter;
+    std::vector <int> orderToReadNumberOfSides;
+  public:
```
alicerussell1: There is no need to write public twice, once applies to the variables and functions

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,244 @@
...
+      
+    }
+    void displayByType(std::string shapeType);
+    void displayByArea(void);
+    void displayByPerimeter(void);
+    void displayByNumberOfSides(void);
+
+
+};
+void ShapeSorter :: displayByType(std:: string shapeType) {
```
alicerussell1: It is more efficient to pass string by constant reference i.e. `const std::string shapeType`

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,244 @@
...
+  int i{ 1 };
+  for (auto shape : allShapes) {
```
alicerussell1: Good use of a range base for loop, it can be made more efficient by passing by const reference

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Shapes.h
```diff
@@ -0,0 +1,244 @@
...
+    if ((*shape).type == shapeType) {
+      
+       (*shape).calculateArea();
```
alicerussell1: You can use arrow notation here to make it easier to read so this would be `shape->calculateArea`

exercises/Toluwalase_Agoro/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,57 @@
+#include "Shapes.h"
+
+int main(int, char **)
+{
+  Square smallSquare{4};
+  Square largeSquare{ 6 };
+  Square largerSquare{ 8 };
+  Square largestSquare{ 10 };
+ // std::cout << largeSquare.calculateArea() << "   " << largeSquare.calculatePerimeter();
```
alicerussell1: Can we avoid having commented out code, I'd recommend when creating PRs so go through and delete anything left that is no longer needed

exercises/Toluwalase_Agoro/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,57 @@
...
+  Triangle smallTriangle{ 3,4 };
+  Triangle largeTriangle{ 6,8 };
+  Triangle largestTriangle{ 10,12 };
+
+  Rectangle smallRectangle{ 2,3 };
+  Rectangle largeRectangle{ 4,5 };
+  Rectangle largerRectangle{ 6,7 };
+  Rectangle largestRectangle{ 8,10 };
+
+  std::vector < BaseShape*> allShapes;
```
alicerussell1: This is an interesting approach - technically not wrong but have you considered using smart_pointers here rather than raw pointers. If you've not come across them yet I would recommend looking into them

alicerussell1 (changes requested):
This is a good solution, well done 👍 Just a few comments below - let me know if anything doesn't make sense. Note a number of the comments apply across the different classes but I have only written them once.

The main thing to be aware of is how we arrange the code. For each class we would ideally want a separate header and cpp file. So overall you would end up with a number of files e.g. Square.h, Square.cpp, Circle.h, Circle.cpp, BaseShape.h etc.
Within the headers you want to include the class declaration, function definitions and variables and in the cpp file you will want to include the function implementations. You code is separated out correctly for this so it should just be a case of moving into separate files.

exercises/Toluwalase_Agoro/ex02_oo_basics/src/BaseShape.h
```diff
@@ -0,0 +1,20 @@
...
+#include <algorithm>
+
+
+class BaseShape {
+  public:
+    int m_numberOfSides{0};
+    std::string m_type;
+    double m_area{0.0};
+    double m_perimeter{ 0.0 };
+    virtual double calculateArea()=0;
```
alicerussell1, line 16: These function can be `const` as they should not change anything within shape, so this would become
`virtual double calculateArea() const = 0;`

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include <string>
+#include "BaseShape.h"
+
+class Circle : public BaseShape {
+public:
+  const double pi = 3.14159265358979323846;
```
alicerussell1, line 7: Constants like this can be static i.e. independent of the class and following the standards the variable name should also be in capitals. However in Mantid we would use the constant `M_PI` which is found in `cmath`.

exercises/Toluwalase_Agoro/ex02_oo_basics/src/BaseShape.h
```diff
@@ -0,0 +1,20 @@
...
+#include <iomanip>
+#include <algorithm>
+
+
+class BaseShape {
+  public:
+    int m_numberOfSides{0};
+    std::string m_type;
+    double m_area{0.0};
+    double m_perimeter{ 0.0 };
```
alicerussell1, line 15: These can be `private` (or if the need to be accessed by the subclass `protected`)

exercises/Toluwalase_Agoro/ex02_oo_basics/src/Rectangle.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include <string>
+#include "BaseShape.h"
+
+class Rectangle : public BaseShape {
+public:
+  double calculateArea() override;
+  double calculatePerimeter() override;
+  double m_length;
+  double m_width;
```
alicerussell1, line 10: These can be `private`

alicerussell1: Similarly with the member variables in the other shapes

exercises/Toluwalase_Agoro/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,123 @@
...
+void ShapeSorter::addShape(std::shared_ptr<BaseShape> newShape) {
+  newShape-> calculateArea();
+  newShape-> calculatePerimeter();
+  m_allShapes.emplace_back(newShape);
+
+}
+
+void ShapeSorter::displayByType(std::string& shapeType) {
+  int i{ 1 };
+  for (auto& shape : m_allShapes) {
```
alicerussell1, line 23:
```suggestion
  for (const auto& shape : m_allShapes) {
```

exercises/Toluwalase_Agoro/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,123 @@
...
+      std::cout << std::setw(8) << "Area: " << shape->m_area << "  ";
+      std::cout << std::setw(15) << "Perimeter: " << shape->m_perimeter << std::endl;
+      i++;
+    }
+  }
+}
+
+void ShapeSorter::displayByArea() {
+
+  if (m_orderToReadArea.size() == 0) {
```
alicerussell1, line 43: This check is not needed and could lead to bugs, for example if someone displayed the areas then added another shape and wanted to display again the new shape would not be in the list.

exercises/Toluwalase_Agoro/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,123 @@
...
+    for (int i = 0; i < m_allShapes.size(); i++) {
+      double maxArea{ 0 };
+      int index{ 0 };
+      for (int j = 0; j < m_allShapes.size(); j++) {
+        if (m_allShapes[j]->m_area > maxArea && std::find(m_orderToReadArea.begin(), m_orderToReadArea.end(), j) == m_orderToReadArea.end()) {
+          index = j;
+          maxArea = m_allShapes[j]->m_area;
+        }
```
alicerussell1, line 51: Have you come across the [sort](http://www.cplusplus.com/reference/algorithm/sort/) function in the standard library? This can be used here to do most of the work for us.

exercises/Toluwalase_Agoro/ex02_oo_basics/src/ShapeSorter.cpp
alicerussell1: Same for perimeter and number of sides

exercises/Toluwalase_Agoro/ex02_oo_basics/src/ShapeSorter.cpp
alicerussell1: Same for perimeter and number of sides

exercises/Toluwalase_Agoro/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,123 @@
...
+      std::cout << std::setw(15) << "Perimeter: " << shape->m_perimeter << std::endl;
+      i++;
+    }
+  }
+}
+
+void ShapeSorter::displayByArea() {
+
+  if (m_orderToReadArea.size() == 0) {
+    for (int i = 0; i < m_allShapes.size(); i++) {
```
alicerussell1, line 44: This could be a range-based for loop

exercises/Toluwalase_Agoro/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,25 @@
...
+#include <memory>
+#include <vector>
+#include <string>
+#include "BaseShape.h"
+class ShapeSorter {
+public:
+  std::vector < std::shared_ptr<BaseShape>> m_allShapes{};
+  std::vector <int> m_orderToReadArea;
+  std::vector <int> m_orderToReadPerimeter;
+  std::vector <int> m_orderToReadNumberOfSides;
```
alicerussell1, line 11: These can be private.

alicerussell1 (changes requested):
Thank you the changes, this is becoming a great solution, I've added a few more comments that I didn't notice last time.

martyngigg:

PR #70, opened 2020-07-23
Tom Clayton Exercise 2

clayton-tom:


exercises/tom_clayton/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,18 @@
+#include "Circle.h"
+#include <cmath>
+
+const double PI = 2 * acos(0.0);
+
+//Circle constructor
+Circle::Circle(double radius) {
+  m_radius = radius;
```
alicerussell1, line 8: well done for using the `m_` prefix 👍

exercises/tom_clayton/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,18 @@
+#include "Circle.h"
+#include <cmath>
+
+const double PI = 2 * acos(0.0);
```
alicerussell1, line 4: Good job including pi as a constant at the top and following the standard for naming. However there is a simpler way as `cmath` contains the constant `M_PI`

exercises/tom_clayton/ex02_oo_basics/src/Circle.h
```diff
@@ -0,0 +1,17 @@
+#include "Shape.h"
```
alicerussell1, line 2: At the start of each header you will want to include `#pragma once`. This is a header guard that prevents having duplicate imports

exercises/tom_clayton/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,35 @@
+#ifndef SHAPE_H_
+#define SHAPE_H_
+
+#include <string>
+
+class Shape
+{
+  protected:
```
alicerussell1, line 6: It would be better if you could make these private if possible

exercises/tom_clayton/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,35 @@
...
+    std::string m_type;
+    int m_sides;
+    
+  public:
+    Shape();
+
+    std::string shapeToString();
+
+    //Virtual functions
+    virtual double calcPerimeter(double h, double b);
```
alicerussell1: You can make these pure virtual which means that may class that inherits from this class has to override this function. If you do this then you can include the formula for calculating perimeter/area in the overridden functions rather than passing it into shape.

To override the functions you just declare the function in the subclass as you have declared `calcPerimeter` in Triangle. You will need to do it for each shape

exercises/tom_clayton/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,35 @@
...
+    //Member variable accessors and mutators
+    double getPerimeter();
+    double getArea();
+    std::string getType();
+    int getSides();
+
+    void setPerimeter(double perimeter);
+    void setArea(double area);
+    void setType(std::string type);
+    void setSides(int sides);
```
alicerussell1, line 32: These setters and getters are fine but you don't necessarily have to include one for each parameter if they are not used.

exercises/tom_clayton/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include <vector>
+
+//ShapeSorter constructor
+ShapeSorter::ShapeSorter(std::vector<Shape> shpVect) {
+  m_shpVect = shpVect;
+  m_vecLen = shpVect.size();
+}
+
+// Typedef to be used to iterate through the vector
+typedef std::vector<Shape>::size_type VecSz;
```
alicerussell1: Good use of a `typedef` 👍

exercises/tom_clayton/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+/** Print out all shapes of a given type, or an error message if no
+ *  shapes of that type are found
+ *  @param string type for type of string
+ */ 
+const void ShapeSorter::printShapesType(const std::string type) {
+  bool flag = false;
+  for(VecSz i = 0; i < getVecLen(); ++i) {
```
alicerussell1: You could use a range-based for loop (I don't think they are in the book but I would look into them if you haven't seen them before). For example this would be:
`for(const auto &shape: m_shpVect)`

exercises/tom_clayton/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+/** Print out all shapes of a given number of sides, or an error
+ *  message if no shapes of that type are found
+ *  @param int sides for no of sides
+ */ 
+const void ShapeSorter::printShapesSides(int sides) {
+  bool flag = false;
+  for(VecSz i = 0; i < getVecLen(); ++i) {
+    if(m_shpVect[i].getSides() == sides) {
+      m_shpVect[i].shapeToString();
```
alicerussell1: I think you have forgotten to actually print the string that's outputted here

exercises/tom_clayton/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,114 @@
...
+    std::cout << "No shapes found of type " << type << std::endl;
+  }
+}
+
+/** Print out all shapes of a given number of sides, or an error
+ *  message if no shapes of that type are found
+ *  @param int sides for no of sides
+ */ 
+const void ShapeSorter::printShapesSides(int sides) {
+  bool flag = false;
```
alicerussell1: `flag` isn't a very useful name as you don't know what this flag is for. Any variable should be clear what it is without having to look through all the code

exercises/tom_clayton/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,40 @@
+#ifndef SHAPESORTER_H_
+#define SHAPESORTER_H_
+
+#include "Shape.h"
+#include "Square.h"
+#include "Rectangle.h"
+#include "Circle.h"
+#include "Triangle.h"
```
alicerussell1: You should not need to include the headers for all the shapes - just the `Shape.h` class. Try to avoid including any unnecessary imports as for large projects like mantid, this can slow the compile time

exercises/tom_clayton/ex02_oo_basics/src/Triangle.h
```diff
@@ -0,0 +1,31 @@
...
+    double m_base;
+
+  public:
+    Triangle(double height, double base);
+
+    void setBase(double base);
+
+    void setHeight(double height);
+
+    double calcPerimeter(double height, double base);
```
alicerussell1: As this method is overriding a virtual function in `Shape` you can use the [override](https://en.cppreference.com/w/cpp/language/override) keyword which will check if it matches the function it is overriding when compiling

alicerussell1 (changes requested):
This is a great start, you've followed all the standards and there is great documentation 👍

I am unable to build it locally - I am getting an error with `std::sort` in the `ShapeSorter`. This will be because for sort to work the functions `areaCompare` and `periCompare` need to be static (independent of the class). 

I have added some comments below - let me know if anything doesn't make sense,

exercises/tom_clayton/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,32 @@
...
+Circle::Circle(double radius) : Shape("Circle", 0) {
+  m_radius = radius;
+  Shape::setPerimeter(calcPerimeter());
+  Shape::setArea(calcArea());
+}
+
+
+// Implementation of Shape virtual functions
+double Circle::calcPerimeter() {
+  double r = getRadius();
```
alicerussell1, line 16: Can we make this `const`? I generally try to make everything `const` and then removing it when it should change.

exercises/tom_clayton/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,32 @@
+#include "Circle.h"
+#include <cmath>
+
+const double PI = 2 * acos(0.0);
+
+//Circle constructor
+Circle::Circle(double radius) : Shape("Circle", 0) {
```
alicerussell1, line 7: I think this is probably a typo but a circle should have 1 side

exercises/tom_clayton/ex02_oo_basics/src/Rectangle.h
```diff
@@ -0,0 +1,26 @@
...
+{
+  private: 
+    double m_side1;
+    double m_side2;
+
+  public:
+    Rectangle(double side1, double side2);
+
+    // Implementation of Shape virtual functions
+    double calcPerimeter() override;
```
alicerussell1, line 14: These methods can be `const` too as they do not change the shape, so this would become:
`double calcPerimeter() const override`

exercises/tom_clayton/ex02_oo_basics/src/Shape.cpp
```diff
@@ -0,0 +1,55 @@
...
+//Mutator methods
+void Shape::setPerimeter(double perimeter) {
+  m_perimeter = perimeter;
+}
+
+void Shape::setArea(double area) {
+  m_area = area;
+}
+
+void Shape::setType(std::string type) {
```
alicerussell1, line 32: It is best to pass strings by reference as it is more efficient e.g. 
`void Shape::setType(std::string &type)`

exercises/tom_clayton/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,111 @@
...
+  m_vecLen = shpVect.size();
+}
+
+/** Print out all shapes of a given type, or an error message if no
+ *  shapes of that type are found
+ *  @param string type for type of string
+ */ 
+const void ShapeSorter::printShapesType(const std::string type) {
+  bool shapeTypeFound = false;
+  for(auto shape_ptr : m_shpVect) {
```
alicerussell1, line 38: You can use const references here as we are not changing each shape within the loop
 i.e `for(const &auto shape_ptr : m_shpVect) `

alicerussell1: Similarly for the other for loops in this file

exercises/tom_clayton/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,31 @@
...
+  private: 
+    std::vector<std::shared_ptr<Shape>> m_shpVect;
+    VecSz m_vecLen;
+    
+  public:
+    ShapeSorter(std::vector<std::shared_ptr<Shape>> shpVect);
+
+    const void printShapesType(std::string);
+
+    const void printShapesSides(int);
```
alicerussell1, line 20: For readability it is good practice to have parameter names in the header so people don't have to go to the cpp or docs to figure out what each of the parameters are. e.g. `const void printShapesSides(int sides);`

exercises/tom_clayton/ex02_oo_basics/src/Circle.cpp
alicerussell1: There are a number of instances where `const` can be added though out the solution - I won't put comments on all of them

alicerussell1 (changes requested):
This is looking good now! I've just added a few final (quite picky) comments below

martyngigg:

PR #71, opened 2020-07-24
Matthew cumber exercise 1 Branch

Matt-Cumber:
I renamed my exercise 1 branch to match my exercise 2 branch naming format and I'm not sure if I've created a new branch or not so I'm making a new pull request.


DannyHindson:
While you can rename a branch locally I think on the remote repository (ie GitHub) what happens is that you end up with a new 2nd branch. So to finish the process you need to delete the old GitHub branch. Please can you try and run this:

git push origin --delete <old_name>

This will also (I think) close the old PR down but useful to see the comments on there so adding a link to the old PR: #62

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -6,29 +8,31 @@
 #include <sstream>
 #include <vector>
 #include <ctype.h>
-#include <regex>
-#include "UniqueWord.h"
-
-std::string PATH = ""; // Will store the path of the file to be read
 
-/** Tests if a word has already been found the increments it's count
+/** Checks if a word already exists in the vector or words, then increments it's count if found,
```
DannyHindson: Please can you change "it's" to "its"

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -6,29 +8,31 @@
...
+ * @param word The word being tested
+ * @return No return value 
  */
-bool inUniqueWords(std::vector<UniqueWord> &uniqueWords, const std::string &word)
+void incrementCountOrAddNewWord(std::vector<UniqueWord> &uniqueWords, const std::string &word)
 {
-  // Loop unique words, if word exists increment it's count
-  for(auto &w : uniqueWords)
+  // Attempt to find word in the vector uniqueWords
+  auto it = std::find_if(uniqueWords.begin(), uniqueWords.end(), [&word](const UniqueWord &obj){return obj.getWord() == word;});
```
DannyHindson, line 22: Thanks for changing. Looking at this again my argument for changing on my previous review wasn't quite right.
If you had been using a "raw" for loop that accessed the items in the vector using an index eg uniqueWords[i] then not all of the STL containers support random access like this. However you weren't...
So the benefit of this change is actually about performance - the STL find function on some platforms may well be quicker than writing your own loop

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    // Now separate original word into words split by hyphens
+    std::istringstream ss(word);
+    std::string wordWithoutHyphens;
+    while(std::getline(ss, wordWithoutHyphens, '-'))
+    {
+      // Skip words less than 5 characters
+      if(wordWithoutHyphens.length() < 5) continue;
+
+      // Else check if word has been found already and increment it's count or create a new word
```
DannyHindson: Please can you change "it's" to "its"

exercises/matthew_cumber/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+  // If not at end, then word has already been found in input file and is stored in the vector, so increment count
+  if(it != uniqueWords.end())
+    it->incrementCount();
+
+  // Else the word has not been found previously, so create a new instance and add it to the vector 
+  else 
+  {
+    auto newWord = UniqueWord(word, 1);
+    uniqueWords.push_back(newWord);
```
DannyHindson: In general we're trying to move towards using the function emplace_back rather than push_back because in some circumstances it gives better performance (when you're adding a temporary to a vector).
This isn't one of those cirucmstances but to avoid too much thought we're simply defaulting to using emplace_back in all cases so please could you swap to that

DannyHindson (changes requested):
Thanks for the changes + can see you fed in a couple of changes which perhaps originated in the exercise 2 review (eg m_ names for members) 
Added a few more v minor comments but think we're fast converging.
Please could you also remove the file exercises/matthew_cumber/ex01_basics/src/.DS_Store

DannyHindson (approved):
Great! I think we're finished. No further comments from me. At this point we would normally merge the code into the master branch but we skip that step in this special "newstarter" repository

martyngigg:

PR #74, opened 2020-12-14
Daniel Murphy exercises

DanielMurphy22:
New Starter exercises: https://www.mantidproject.org/New_Starter_C%2B%2B_introduction


exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
+#include <iostream>
+#include <fstream>
+#include <map>
+#include <string>
+#include <vector>
+#include <algorithm>
+#include <functional>
+using namespace std;
```
DannyHindson: I'm not sure if this is actually written down anywhere in our coding standards but we don't use "using namespace std" in Mantid. The std:: prefix is used wherever a type is used from the standard library. Looks like you've already done this in most cases if not all so you just need to remove this using statement

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+		// do not log spaces as words
+		if (length > 4)	{	words.push_back(line.substr(((*iter)+1), length));	}
+	}
+
+	return words;
+}
+
+std::map<string, int> ReadFile::UpdateMap(std::vector<string> words)
+{
+	static std::map<string, int> Counter;
```
DannyHindson: The approach you've used for building up the map works but it's a bit obscure - in particular the use of a static variable "Counter" here that lives on between calls to this UpdateMap function. The approach might cause problems if there was more than one thread calling UpdateMap because there can only ever be a single instance of "Counter" shared between all calling code.

It would be more standard to pass the map into the function by reference and get the function to update the map declared on L111. With this approach you won't need to copy the map every time the function finishes (which I think is what's happening every time "return Counter" occurs on L69) so might be quicker as well as working better with multiple calling threads.

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+	for (string word : words)
```
DannyHindson: Nice use of range for loop. Worth putting "for (string& word: words)" to avoid copying each string from the words vector during each iteration. Or you'll see people using the auto keyword here ie "for (auto&word: words)"

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+	{
+		int length = *(iter+1) - ((*iter)+1); // word length = end - start
+		// do not log spaces as words
+		if (length > 4)	{	words.push_back(line.substr(((*iter)+1), length));	}
+	}
+
+	return words;
+}
+
+std::map<string, int> ReadFile::UpdateMap(std::vector<string> words)
```
DannyHindson: You should make the words parameter a const reference parameter here to avoid a copy and to highlight the fact you're not changing it ie "UpdateMap(const std::vector<string> &words)"

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// https://www.geeksforgeeks.org/sorting-a-map-by-value-in-c-stl/
+bool cmp(std::pair<string, int>& a, std::pair<string, int>& b)
+{
+	return a.second > b.second;
+}
+
+// Convert map into vector of pairs, sorted by value of counts
+std::vector<std::pair<string, int> > sort_by_value(std::map<string, int>& M)
```
DannyHindson: I think you can make the single parameter of this function "const" ie const std::map<string, int>& M

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+	static std::map<string, int> Counter;
+	for (string word : words)
+	{
+		auto member_check = Counter.find(word);
+		// if word already a member, increment its count by 1
+		if(member_check != Counter.end())
+
+		{	Counter[word]++ ;	} 
+		else // add word to map
+		{	Counter[word] = 1;	}
```
DannyHindson: You don't need an if else statement here strictly speaking because the [] operator creates a new entry in the map with a zero count if one doesn't exist. The ++ operator would then increment it to 1 as required. So the first part of the if works in all cases

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+		CounterMap = o.UpdateMap(words_analysed);
+	}
+
+	// Convert map into vector of pairs, sorted by value of counts
+	auto CounterVec = sort_by_value(CounterMap);
+
+	// Write out results to file
+	string filename_out = "Stats_WC_" + filename;
+	std::fstream fs_out{filename_out, std::ios::out};
+	fs_out << "\n \nWords                       Count\n#####                       #####\n \n";
```
DannyHindson: The padding of the columns can be done a bit more elegantly using the setw function. It works something like this:
```
fs_out << setw(20) << "Words" << setw(8) << "Count";
```

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
...
+	positions.push_back(static_cast<int>(line.length())); // add the end of the line
+
+	// Cut the line by the space positions into words
+	std::vector<string> words;
+	for (auto iter = positions.begin(); iter != positions.end()-1; iter++)
+	{
+		int length = *(iter+1) - ((*iter)+1); // word length = end - start
+		// do not log spaces as words
+		if (length > 4)	{	words.push_back(line.substr(((*iter)+1), length));	}
+	}
```
DannyHindson: This chunk of code between L36 and L51 works fine but it can be done more concisely using the std::istringstream class (input stream). This mimics someone keying in data to a terminal via cin. Probably quickest to explain with an example. Something like this will do the equivalent work where the >> operator does the chopping up into words for (almost) free:
```
  std::istringstream lineOfText(line);
  std::string word;
  while (lineOfText >> word) {
    if (word.size() > 4)
      words.push_back(word);
  }
```
You'll need #include \<sstream\> also

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,151 @@
+#include <iostream>
+#include <fstream>
+#include <map>
+#include <string>
+#include <vector>
+#include <algorithm>
+#include <functional>
+using namespace std;
+
+class ReadFile
```
DannyHindson: I'm not sure the class adds a lot of value here. You may as well just have a couple of standalone functions I think.

DannyHindson (changes requested):
Sorry got distracted a lot today so only just got rounding to reviewing the first exercise. I've added a few detail comments but looks pretty close and a good first effort. You've got the idea of having to copy the map into a vector to sort which is one of the normal sticking points. Also used a few nice standard library algorithms (transform, erase) along the way.

One formatting point. You need to set up your editor to insert spaces instead of tabs when indenting. The Mantid standard is to use 2 spaces for indentation. Not sure what editor you're using but in Visual Studio this is under Tools, Options and then Text Editor. You basically need to get your editor to insert two spaces whenever you hit the tab key.

I'll have a look at the second exercise now or tomorrow

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
DanielMurphy22: I agree that the use of `static` seems odd and passing by reference. Just one little thought is that in the new cpp book, when it mentions passing by reference (Page 81) it sums up that "Passing by reference is best avoided." without mentioning any cases when it could be useful.

It obviously doesn't make sense here to pass by const ref, but just out of interest what else would work here? Would using an initialisation constructor to hold the map in an overarching class work better?

DavidFair: It's not my place to reply to this, but somehow I ended up subscribed when I posted the other PR

What you could do is encapsulate this into a nice OO object, since we have multiple related methods spread around and we might want to handle edge cases (like updating the count for a single word) in the future:

```
class WordMap{
public:
int getNumberWords(const std::string &word);
void updateWordMap(const std::vector<std::string> &wordList);

private:
using WordMapT = std::map<std::string, int>;
WordMapT counter;
};
```

Then all you need to do is initialise the object type and pass that by ref instead

```
WordMap mappedWords;
mappedWords.updateWordMap(myWords);

```
We can also pass by ref which is very cheap (one pointer length)
```
void printWordMapCounts(const WordMap &allWords){
   const auto &seenWords = allWords.getAllWordsSeen();     // Assume this exists
   for (const auto& word: seenWords){
        std::cout << "I saw " << word << allWords.getNumberWords(word) << " times";
   }
}
```

DavidFair: > (Page 81) it sums up that "Passing by reference is best avoided." without mentioning any cases when it could be useful.

What I assume they are trying to say (though I don't have a copy of the book) is that it's a code smell for a few reasons: 

- People will pass in data structures by reference to free functions to modify the data inside. This is programming using C methodology in C++, tie your data and the methods that manipulate it together in a class.
- Modifying parameters forces the reader to look into the implementation to check what happens to their container. If you saw `std::map<std::string, int> getWordCount(const std::string sentence);` you'd know straight away that it's going to make and return a new container (though this gets into the computation cost too)
- People tend to fail the "and / or test" presented in Clean Code, especially having multiple out parameters. "Can you describe your function without using the word *and* or *or*, if not it's likely doing too much and needs chunking down"

DannyHindson: I'll see if I can get a word in between David's posts :wink:
Funny you noticed that sentence in the new book. I was asked by Martyn\Lamar to review the book with a view to updating the new starter reading list and that line jumped out at me also. I personally think it's a bit strong to say "passing by reference is best avoided". It wouldn't be my first choice of how to pass data into and out of a free function but to say avoid it completely is overshooting a bit in my opinion.
As David's pointed out creating a class is one way to avoid it. I guess if you've more than one function acting on the map it makes the extra lines of code to define the class and create it worthwhile. I believe the exercises are designed to mean that the "new starter" doesn't have to show off their knowledge of classes until exercise 2 so most people don't use a class when answering this question.
So in terms of what you should change to get this PR approved - I don't mind whether you just change the function to take the parameter by reference or create a class. I had commented that your ReadFile class wasn't adding much value but that was because it didn't have a data member. If you follow David's example, it would have a data member.

DanielMurphy22: Thanks to both!!!! It's really interesting having comments on my code and the further explanations by David. I'll just change the function to pass by ref for the sake of this exercise and bear in mind that using a class with a data member could be a better option in the future. 
I'm glad to hear that you think the book was a little strong on avoiding pass by reference. It can be hard for someone new to cpp to know how much the advice in the book is a reflection of the community consensus or just the author's opinions. Thankfully they are very clear at many points about avoiding raw pointers, but possibly some other things aren't stated objectively.

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
DanielMurphy22: This is quite nice. I'd neglected sstream after happily using fstream already. Out of interest, on page 204, section 37.2 on string streams, do you think they meant option a. to be `std::istringstream` rather than a. and c. being the same??

exercises/Daniel_Murphy/ex02_oo_basics/src/main.h
```diff
@@ -0,0 +1,104 @@
...
+  virtual int get_sides() = 0;
+  virtual double get_area() = 0;
+  virtual double get_perimeter() = 0;
+  virtual ~Shape() {};
+};
+
+class Triangle : public Shape
+{
+public:
+  double b, h;
```
DannyHindson: We normally prefix any data member names with `m_` so these would be called `m_b` and `m_h`
Also it's worth using more than a single letter if possible to better to call them `m_base` and `m_height`
I won't repeat this against all the members but please can you rename any other data members

exercises/Daniel_Murphy/ex02_oo_basics/src/main.h
```diff
@@ -0,0 +1,104 @@
...
+  std::string type = "Triangle";
```
DannyHindson: I think you've got this type member defined against each of the "child" classes. It would be better to define this once against the parent `Shape` class and then implement the getType method on the `Shape` class as well. You can then specify the value of the type data member in the constructor of the child classes eg
```
Square(double side1)
    : type{"Square"}, s1{side1}
```
The top level class doesn't have to be a pure abstract class ie 100% pure virtual without any data members

exercises/Daniel_Murphy/ex02_oo_basics/src/main.h
```diff
@@ -0,0 +1,104 @@
...
+  int sides = 3;
```
DannyHindson: If you've got sides defined on each child class as well then the same comment I've made about the type data member applies here also - you can move it and the corresponding "get" function up to the parent class

exercises/Daniel_Murphy/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,264 @@
...
+  return sides;
+}
+double Circle::get_perimeter()
+{
+  double p = 2 * M_PI * r;
+  return p;
+}
+double Circle::get_area()
+{
+  double a = 4 * M_PI * pow(r,2);
```
DannyHindson: The area should be just pi r squared

exercises/Daniel_Murphy/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,264 @@
...
+    std::cout << std::string(7-num_len, ' ') << n << std::string(25-word_len, ' ') << a << "\n";
+  }
+  std::cout << "\n";
+}
+
+int main()
+{
+  ShapeSorter o;
+
+  Triangle a = {1.56,7.8}; o.add_new(a);
```
DannyHindson: Please can you separate these two statements onto two lines

exercises/Daniel_Murphy/ex02_oo_basics/src/main.h
```diff
@@ -0,0 +1,104 @@
...
+  std::string get_type() override;
+  int get_sides() override;
+  double get_perimeter() override;
+  double get_area() override;
+};
+
+class ShapeSorter
+{
+public:
+  std::vector<std::unique_ptr<Shape>> shapes;
```
DannyHindson: Well done for using a vector of pointers. I think that if you just have a vector of shapes then the polymorphic behaviour doesn't kick in when you call things like get_perimeter

exercises/Daniel_Murphy/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,264 @@
...
+  }
+  std::cout << "\n";
+}
+
+void ShapeSorter::area_descending()
+{
+  std::cout << "\nShape Number      Area / sq u \n";
+  std::cout << "############      ###########\n\n";
+
+  std::map<double, int> area_map;
```
DannyHindson: The approach you've used of using a map to sort the data is perhaps a bit unusual. It would probably be a bit more obvious to the reader if you just sorted the shapes vector using std::sort. Or if you prefer to keep the master copy of the vector in it's original order, you could copy the vector and then sort. The vector is a vector of pointers so you're not copying the actual objects so this wouldn't be v costly.
Perhaps the use of the map might also cause a problem if two shapes had the same area or perimeter? Possible the entry would get overwritten

exercises/Daniel_Murphy/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,264 @@
...
+  int number = 1;
+  for (auto const&  el : shapes)
+  {
+    double shape_area = el->get_area();
+    area_map[shape_area] = number;
```
DannyHindson: The link between the shapes and the numbers you're using as an identifier feels a little bit loose. It would be nicer if you had a method on the shape called getID() or similar that returned the number. I think that would probably mean you'd have to supply the ID or make the ShapeSorter generate it when the shape is added to the ShapeSorter

exercises/Daniel_Murphy/ex02_oo_basics/src/main.h
```diff
@@ -0,0 +1,104 @@
...
+  void add_new(Triangle a);
+  void add_new(Square a);
+  void add_new(Rectangle a);
+  void add_new(Circle a);
```
DannyHindson: I think you should simple have an add method that takes a parameter of type Shape. You could then call `shapes.emplace_back` inside it. There's not much difference between emplace_back and push_back in this case because you're adding a pointer to a shape rather than creating a shape "in place" in the vector. But we do generally use emplace_back rather than push_back because it can be faster in some cases so do carry on using emplace_back.

Once you've done this, you will need to move the creation of the pointer into the calling code in main.cpp:
```
o.add(std::make_unique<Triangle>(6,4))
```
As you can see I've done away with the `b` object here because it's quicker just to create it inside the call to make_unique

exercises/Daniel_Murphy/ex02_oo_basics/src/main.cpp
DannyHindson: I've made a comment against main.h on a v similar topic, but I don't think there's much point in creating the shape `a` here and then creating a second triangle inside add_new (which is what's happening when std::make_unique is called). You may as well just create it once inside std::make_unique and move that call into this main() method

DannyHindson (changes requested):
As well as the detail comments here's one general comment. Normally there would simply be a main.cpp without a corresponding main.h. If you need to create some classes this would normally be done in a separate .cpp\.h pair of files. So I suggest you get rid of main.h and have main.cpp and then Shapes.h and Shapes.cpp.
If the classes are of a reasonable size, we would have one .h and one .cpp per class but that is perhaps overkill here. Fine to put all the shape classes in one .h\.cpp
That aside it looks pretty good. Few tweaks around the class set up but you've got the hang of most of the concepts I think.

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+std::map<std::string, int> UpdateMap(const std::vector<std::string> &words, std::map<std::string,int>& Counter)
+{
+  for (auto& word : words)
+  {
+    // if word already a member, increment its count by 1
+    Counter[word]++ ;
+  }
+  return Counter;
```
DannyHindson: Thanks for changing this. Now that you're passing Counter by reference you don't need a return type in the function signature nor a return statement here. L92 will also need adjusting if you change this

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  return words;
+}
+
+std::map<std::string, int> UpdateMap(const std::vector<std::string> &words, std::map<std::string,int>& Counter)
+{
+  for (auto& word : words)
+  {
+    // if word already a member, increment its count by 1
```
DannyHindson: Bit pedantic this but you probably don't need this comment now that the single line manages the situation where word isn't already in the map (if you were going to explain anything with a comment it should perhaps highlight that L43 manages the "not already there" situation).

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,126 @@
...
+std::vector<std::pair<std::string, int> > sort_by_value(const std::map<std::string, int>& M) 
+{
+  // Declare vector of pairs 
+  std::vector<std::pair<std::string, int> > A; 
+
+  // Copy key-value pair from Map to vector of pairs 
+  for (auto& it : M) { A.push_back(it); } 
+
+  // Sort using comparator function 
+  sort(A.begin(), A.end(), cmp);
```
DannyHindson: Although this compiles it's good practice to write std::sort rather than simply sort.
I think it compiles OK as it is because when the compiler doesn't find "sort" in the current namespace it spots that the parameters are defined in the std namespace so it goes off and looks for a sort in that namespace. If you defined your own sort though the absence of the prefix cause confusion over which one is being called

DannyHindson (changes requested):
I've added a few minor comments against exercise 1 but I think that one will be complete after these are addressed. Thanks

exercises/Daniel_Murphy/ex02_oo_basics/src/main.h
DanielMurphy22: I feel like I'm missing something basic about inheriting data members.

This short example works:
```
#include <iostream>
#include <string>

class Shape
{
public:
  std::string m_type = "Shape";
};

class Triangle : public Shape
{
public: 
//   Triangle()
//     : m_type{"Triangle"}
//   {}
  void out()
  {
    std::cout << m_type;
  }

};

int main()
{
  Triangle o;
  o.out();
}
```
But this does not ( `class 'Triangle' does not have any field named 'm_type'`):
```
#include <iostream>
#include <string>

class Shape
{
public:
  std::string m_type = "Shape";
};

class Triangle : public Shape
{
public: 
  Triangle()
    : m_type{"Triangle"}
  {}
  void out()
  {
    std::cout << m_type;
  }

};

int main()
{
  Triangle o;
  o.out();
}
```
So in the first one it seems that m_type is inherited correctly and it's not for the constructor in the second example??

DannyHindson: Sorry Daniel, it's my dodgy code I'd written straight into GitHub. In my earlier comment I should have written:
```
Square(double side1)
    : Shape{"Square"}, s1{side1}
```
You will then need to add a constructor to the Shape class that takes a single string parameter and writes this to m_type

DanielMurphy22: Ahh! That makes sense. I had thought there must be a more general way to put these data members into the parent class but I didn't know how to use the constructor in the children like that. 

Thanks for the quick response!!

DanielMurphy22:
Hmm... build failures are due to the `{}` I've added to the virtual functions in `Shapes` to prevent [this error](https://stackoverflow.com/questions/8830025/linker-error-undefined-reference-to-vtable-for-square-code-includes-virtual): 
`undefined reference to 'vtable for Shape'`
So without `{}` my local comipler won't complete and with them the build server won't pass. 

Should I be handling this differently?

DannyHindson:
> Hmm... build failures are due to the `{}` I've added to the virtual functions in `Shapes` to prevent [this error](https://stackoverflow.com/questions/8830025/linker-error-undefined-reference-to-vtable-for-square-code-includes-virtual):
> `undefined reference to 'vtable for Shape'`
> So without `{}` my local comipler won't complete and with them the build server won't pass.
> 
> Should I be handling this differently?

If you want to define a pure virtual function at the top of a class hierarchy you should just put =0 on the end:
```
virtual double get_area()=0;

**Comment by DanielMurphy22** [comment]
> If you want to define a pure virtual function at the top of a class hierarchy you should just put =0 on the end:
> 
> ```
> virtual double get_area()=0;
> ```

Ahh okay, my understanding of Abstract classes isn't quite there. On pg138 the book says that abstract classes (with at least one pure virtual function) cannot be _instantiated_. Presumably the call to `Shape` [in each child class](https://github.com/mantidproject/newstarter/pull/74/files?file-filters%5B%5D=.h#diff-2ec7deec01174086e5db37407716e25f0beb5ecaf4422162f8e3925e8a1f1c81R41) doesn't count? Does instantiation only count if it's inside `int main()`?

**Comment by DannyHindson** [comment]
> > If you want to define a pure virtual function at the top of a class hierarchy you should just put =0 on the end:
> > ```
> > virtual double get_area()=0;
> > ```
> 
> Ahh okay, my understanding of Abstract classes isn't quite there. On pg138 the book says that abstract classes (with at least one pure virtual function) cannot be _instantiated_. Presumably the call to `Shape` [in each child class](https://github.com/mantidproject/newstarter/pull/74/files?file-filters%5B%5D=.h#diff-2ec7deec01174086e5db37407716e25f0beb5ecaf4422162f8e3925e8a1f1c81R41) doesn't count? Does instantiation only count if it's inside `int main()`?

Yes. This kind of call where a child class constructor "internally" calls a parent class constructor isn't really instantiating an object of the parent class. You still end up with an object of type Triangle or Square rather than an object of type Shape in the calling code.

File: exercises/Daniel_Murphy/ex02_oo_basics/src/shape.h [hunk 22]
```diff
@@ -0,0 +1,103 @@
...
+#include <string>
+#include <cmath>
+#include <algorithm>
+#include <memory>
+#include <map>
+
+class Shape
+{
+public:
+  std::string m_type;
```
DannyHindson: In general, any data members are normally protected or private. I think m_type, m_sides, m_ID can all be private. Although m_type is used for Triangle\Square objects, it's via a method defined on the parent Shape class so private will work OK here

exercises/Daniel_Murphy/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,103 @@
...
+
+  std::unique_ptr<Shape> clone() override;
+  double get_perimeter() override;
+  double get_area() override;
+};
+
+class ShapeSorter
+{
+public:
+  std::vector<std::unique_ptr<Shape>> shapes;
```
DannyHindson: Please can you rename to m_shapes and this can also be private

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,124 @@
...
+  std::string word;
+  while (lineOfText >> word) {
+    if (word.size() > 4)
+      words.push_back(word);
+  }
+
+  return words;
+}
+
+void UpdateMap(const std::vector<std::string> &words, std::map<std::string,int>& Counter)
```
DannyHindson: I know I said I didn't have any more comments on ex 1 but just noticed that this function has got small enough now that it's not really worth having a separate function. I think you should just put the for loop into the calling code lower down

exercises/Daniel_Murphy/ex02_oo_basics/src/shape.cpp
```diff
@@ -0,0 +1,205 @@
...
+  for (const auto& el : shapes)
+    shapes_by_perimeter.emplace_back(el->clone());
+
+  struct {
+    bool operator()(std::unique_ptr<Shape>& a, std::unique_ptr<Shape>& b) const
+    {   
+        return a->get_perimeter() > b->get_perimeter();
+    }   
+
+  } descending_perimeter;
```
DannyHindson: The use of a struct like this with an overloaded () operator to define the comparison function for std::sort isn't something I'd seen before. I think it would be more in keeping with the approaches used in Mantid to define either a standalone function or use a lambda function inside std::sort. Probably best to use a lambda function - something like:
```
std::sort(shapes_by_perimeter.begin(), shapes_by_perimeter.end(), [](std::unique_ptr<Shape>& a, std::unique_ptr<Shape>& b) {return a->get_perimeter() > b->get_perimeter(); });
```

exercises/Daniel_Murphy/ex02_oo_basics/src/shape.cpp
```diff
@@ -0,0 +1,205 @@
...
+  for (const auto& el : shapes)
+    shapes_by_area.emplace_back(el->clone());
+
+  struct {
+      bool operator()(std::unique_ptr<Shape>& a, std::unique_ptr<Shape>& b) const
+      {   
+          return a->get_area() > b->get_area();
+      }   
+
+  } descending_area;
```
DannyHindson: My comment against descending_perimeter applies here as well

exercises/Daniel_Murphy/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,103 @@
...
+  int m_ID;
+
+  std::string get_type();
+  int get_sides();
+  int get_ID();
+
+
+  Shape(std::string type, int sides, int ID)
+    : m_type{type}, m_sides{sides}, m_ID{ID}
+  {}
```
DannyHindson, line 16: Please can you move the constructor to the top of the public section above any normal methods such as get_type

exercises/Daniel_Murphy/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,103 @@
+#ifndef MAIN_H
+#define MAIN_H
```
DannyHindson: Nice header guard. I was just about to point out that the name needed updating to SHAPE_H. But I remembered that we've started using "#pragma once" instead so please can you change to that. This is a preprocessor directive that achieves the same thing ie prevent the header being imported multiple times in a particular compilation.

DannyHindson (changes requested):
Looking good. I've added a few detail level comments.
One overall comment that might be a bit tedious to address - part of the Mantid coding standards for C++ is to use camelCase for variable and function names containing multiple words. You've tended to use underscores instead. Please could you update in both exercises?
In the first exercise for example, the function AnalyseLine which should be analyseLine and CounterVec should be counterVec
The one exception to this is the use of m_ for data members on classes. So m_type is fine

DanielMurphy22:
Thanks @DannyHindson for all the time you've spent giving me really in-depth comments! I think I have implemented camelCase everywhere for both exercises. Coding standards comments like that and say using pragma once are super useful!!

Fingers crossed I'm pretty much there now. Also, do you agree with my comment here?: https://github.com/mantidproject/newstarter/pull/74#discussion_r544279355

exercises/Daniel_Murphy/ex02_oo_basics/src/shape.h
```diff
@@ -0,0 +1,98 @@
...
+#include <vector>
+#include <string>
+#include <cmath>
+#include <algorithm>
+#include <memory>
+#include <map>
+
+class Shape
+{
+private:
```
DannyHindson: Are you around next week Daniel? There's one last thing I would normally pick up on which is that this private section should be below the public section (just another coding standards thing). 

If you're not around though we can just close this off now without this detail because neither of us will remember what was going on after the holidays.

Exercise one looks good now

DavidFair: @DannyHindson Just an FYI, Daniel mentioned on a different PR today was his last day before holidays (in case he doesn't look at this whilst he's enjoying his time off)

DanielMurphy22: Correct, I am off next week. That change is rather small so I've made it and I'll just hope it passes the build server. Merry Xmas guys! 🎅

exercises/Daniel_Murphy/ex01_basics/src/main.cpp
DannyHindson: Yes you're right. a) should be std::istringstream

DannyHindson (approved):
Thanks for making that last change. All looks good now, well done. Happy Christmas!

One check before you get going on changes in the main repository - have you set up clang formatting in your C++ editor\IDE? It's not checked by Jenkins in the new starter repository but it will be in the main repo. If no then you should be able to set this up in your IDE or as a git custom command

gemmaguest:
Closing seeing as this was all completed

PR #75, opened 2020-12-14
Fix compilation errors on CentOS

DavidFair:
Fixes various issues preventing the new starter exercises building, see commits for details

### DIFF
--- a/CMake/CommonSetup.cmake
+++ b/CMake/CommonSetup.cmake
@@ -14,5 +14,14 @@ set_property ( GLOBAL PROPERTY USE_FOLDERS ON )
 
 set ( CMAKE_CXX_STANDARD 17 )
 set ( CMAKE_CXX_STANDARD_REQUIRED ON )
+set ( CMAKE_CXX_EXTENSIONS OFF )
+
+# Enable warnings in all subprojects
+if(MSVC)
+  add_compile_options(/W3 /WX)  # W3 avoids std lib warnings
+else()
+  add_compile_options(-Wall -Wextra -pedantic -Werror)
+endif()
+
 
 add_definitions ( -D_USE_MATH_DEFINES )
--- a/Jenkins/buildscript.py
+++ b/Jenkins/buildscript.py
@@ -19,7 +19,7 @@ import sys
 try:
   WORKSPACE = os.environ["WORKSPACE"]
 except KeyError:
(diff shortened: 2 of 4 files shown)


DanielMurphy22 (approved):
Of course this would fail the build server. Good to have this fixed so it compiles against C++17

DavidFair:
I've updated the build servers to use Python 3 and added a small sample project so we can easily tell if it's the CI or new code causing issues (so a new starter will hopefully report broken CI earlier on)

I've also tweaked the Jenkins jobs not to delete the repos between, since we clone ~500MB each pass and already delete the build dir between runs anyway

@DanielMurphy22 could you give this a re-review when you get a chance

DanielMurphy22 (approved):
I like the new verb "F-string -inate"

DavidFair:
Build failures are expected as we don't have any projects for the CI to officially build
@martyngigg When you get a chance could you review / merge this? I don't have the merge permission on this repo

martyngigg (approved):
Thanks for looking at this. This always gets left out in the cold until someone new gets around to it.

martyngigg:
> I've also tweaked the Jenkins jobs not to delete the repos between, since we clone ~500MB each pass and already delete the build dir between runs anyway

Is this repo really that big?

DavidFair:
> > I've also tweaked the Jenkins jobs not to delete the repos between, since we clone ~500MB each pass and already delete the build dir between runs anyway
> 
> Is this repo really that big?

Huh I just cloned it again to check, now only 137MB instead of the 500MB ish - we do have a new router arriving today as I have had real problems with packet storms on our LAN / corruption etc. so I wonder if this was git trying over and over to get valid pack files :confused: 

I originally assumed people had pushed binaries in the past, but ignore me completely on that one

DanielMurphy22:
> > > I've also tweaked the Jenkins jobs not to delete the repos between, since we clone ~500MB each pass and already delete the build dir between runs anyway
> > 
> > 
> > Is this repo really that big?
> 
> Huh I just cloned it again to check, now only 137MB instead of the 500MB ish - we do have a new router arriving today as I have had real problems with packet storms on our LAN / corruption etc. so I wonder if this was git trying over and over to get valid pack files 😕
> 
> I originally assumed people had pushed binaries in the past, but ignore me completely on that one

It was also slow for me cloning to my Mac on site over VPN yesterday afternoon.

PR #76, opened 2021-02-09
Tom hampson exercises

thomashampson:
Attempt at solving the new starter C++ exercises.

### DIFF
--- a/exercises/tom_hampson/ex01_basics/src/main.cpp
+++ b/exercises/tom_hampson/ex01_basics/src/main.cpp
@@ -1,0 +1,85 @@
+/**
+ * Skeleton main routine
+ */
+#include <iostream>
+#include <string>
+#include <fstream>
+#include <sstream>
+#include <regex>
+#include <algorithm>
+#include <map>
+
+bool compareWordCount(std::pair<std::string, int> wordA, std::pair<std::string, int> wordB)
+{
+	return wordA.second > wordB.second;
+}
+
+int main(int, char **)
+{
+	//For now, just hard code the holmes file.
+	std::string inputFileName = "C:/workdir/newstarter/exercises/tom_hampson/ex01_basics/testData/Holmes.txt";
+
+  
+	std::ifstream inputFile(inputFileName);
+
+	if (!inputFile)
+	{
+		std::cout << "Unable to open file " << inputFileName << std::endl;
+		return 1;
+	}
+
+	//Line of text from input source.
+	std::string line;
+
+	std::map<std::string, int> wordCounts;
+
+	//Read file one line at a time, or until we reach a hyphen.
+	while (std::getline(inputFile, line, '-'))
+	{
+		std::istringstream wordsInLine(line);
+
+		std::string word;
+
+		//Loop over words in the line, using whitespace as delimiter.
+		while (wordsInLine >> word)
+		{
+			//Only consider words > 3 characters long.
+			if (word.length() < 4)
+			{
+				continue;
+			}
+			//Remove all punctuation characters:
+			//.,?'"!():
+			//Need to escape several of them because they are special pattern
+			//characters in regex expressions.
+			std::regex specialCharacters("\\.|,|\\?|'|\"|!|\\(|\\)|:");
+			std::string newWord = std::regex_replace(word, specialCharacters, "");
+
+			//Check again for the length of the word because it will be shorter if punctuation was removed.
+			if (newWord.length() < 4)
+			{
+				continue;
+			}
+
+			//Convert word to lower case.
+			std::transform(newWord.begin(), newWord.end(), newWord.begin(), ::tolower);
+
+			//Increment the count of the word.
+			wordCounts[newWord]++;
+		}
+	}
+
+	//Container to store the word counts sorted by count.
+	std::vector<std::pair<std::string, int>> sortedWordCounts;
+
+	std::copy(wordCounts.begin(), wordCounts.end(), std::back_inserter(sortedWordCounts));
+
+	std::sort(sortedWordCounts.begin(), sortedWordCounts.end(), compareWordCount);
+
+	for (auto wordCount : sortedWordCounts) {
+		std::cout << wordCount.first << ": " << wordCount.second << std::endl;
+	}
+
+	return 0;
+    
+}
--- a/exercises/tom_hampson/ex01_basics/src/main.cpp
+++ b/exercises/tom_hampson/ex01_basics/src/main.cpp
@@ -1,6 +1,6 @@
 /**
  * Read text file and count instances of words with >= 4 letters, then save the information to a file.
- * 
+ *
  */
 #include <iostream>
 #include <string>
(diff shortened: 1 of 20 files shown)


exercises/tom_hampson/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  //Something is wrong if there aren't any words.
+  if (wordCounts.size() == 0)
+  {
+    return false;
+  }
+
+  //Find length of longest word so that we can add padding to align everything nicely.
+  unsigned int longestWordLength = 0;
+  for (auto wordCount : wordCounts)
```
StephenSmith25: I think this could be a const reference, rather than taking a copy of the value
```suggestion
  for (const auto& wordCount : wordCounts)
```

exercises/tom_hampson/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+  if (wordCounts.size() == 0)
+  {
+    std::cout << "No valid words found in file " << inputFileName << std::endl;
+    return 1;
+  }
+
+  //Container to store the word counts sorted by count.
+  std::vector<std::pair<std::string, int>> sortedWordCounts;
+
+  std::copy(wordCounts.begin(), wordCounts.end(), std::back_inserter(sortedWordCounts));
```
StephenSmith25: Could this be: 
```suggestion
  std::move(wordCounts.begin(), wordCounts.end(), std::back_inserter(sortedWordCounts));
```
Doesn't look like we use the wordCounts after these lines, so the move will save us copying each pair.

exercises/tom_hampson/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  //If there are no words, don't do anything else.
+  if (wordCounts.size() == 0)
+  {
+    std::cout << "No valid words found in file " << inputFileName << std::endl;
+    return 1;
+  }
+
+  //Container to store the word counts sorted by count.
+  std::vector<std::pair<std::string, int>> sortedWordCounts;
```
StephenSmith25, line 170: I think you'd want to reserve the size of the vector, other with the back inserter will cause space to be allocated multiple time for the vector.
```suggestion
  std::vector<std::pair<std::string, int>> sortedWordCounts;
  sortedWordsCount.reserve(..)
```

exercises/tom_hampson/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * @brief Print all shapes with specified shape type (e.g. "circle").
+ * @param shapeType 
+*/
+void ShapeSorter::printShapesOfType(const std::string& shapeType) const
+{
+  std::cout << "Shapes of type \"" << shapeType << "\":" << std::endl;
+
+  for (auto shape : m_shapes)
```
StephenSmith25:
```suggestion
  for (const auto& shape : m_shapes)
```
so as not to increase the ref count of each of the shapes (shared pointers)

exercises/tom_hampson/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,38 @@
...
+  //circle.print();
+
+  //Rectangle rectangle(2, 4);
+  //rectangle.print();
+
+  ShapeSorter sorter;
+  sorter.addShape(std::make_shared<Circle>(3));
+  sorter.addShape(std::make_shared<Square>(5));
+  sorter.addShape(std::make_shared<Triangle>(5,4));
+  sorter.addShape(std::make_shared<Triangle>(6,4));
```
StephenSmith25, line 27: Would also have been worth considering a unique_ptr here, as we may not need the shared ownership of each shape, which comes at a cost. But not worth changing.

exercises/tom_hampson/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,30 @@
...
+
+  void printShapesOfType(const std::string& shapeType) const;
+  void printShapesWithNSides(const int n) const;
+  void printShapesAreaDescending() const;
+  void printShapesPerimeterDescending() const;
+  
+  void addShape(std::shared_ptr<Shape>);
+  void removeShape(std::shared_ptr<Shape>);
+
+  static bool compareArea(std::shared_ptr<const Shape> shapeA, std::shared_ptr<const Shape> shapeB);
```
StephenSmith25: Arguments to these functions could be passed by reference, unless we are trying to shared ownership of each shape with the function. Maybe passing each shape by a pointer/reference would be better, as then we are leaving the ownership of the objects to the caller.
```suggestion
  static bool compareArea(Shape * shapeA, Shape *shapeB);
```
Same for the comparePerimeter function

exercises/tom_hampson/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,126 @@
...
+ * @brief Print all shapes in descending order of perimeter size.
+*/
+void ShapeSorter::printShapesPerimeterDescending() const
+{
+  //Make a copy of the shapes vector so we preserve the order of the member shape vector.
+  std::vector<std::shared_ptr<Shape>> sortedShapes = m_shapes;
+  std::sort(sortedShapes.begin(), sortedShapes.end(), comparePerimeter);
+
+  std::cout << "Shapes ordered by perimeter:" << std::endl;
+  for (auto shape : sortedShapes)
```
StephenSmith25: Again could use const reference in these loops.

exercises/tom_hampson/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,126 @@
...
+ * @brief Print all shapes by area in descending order.
+*/
+void ShapeSorter::printShapesAreaDescending() const
+{
+  //Make a copy of the shapes vector so we preserve the order of the member shape vector.
+  std::vector<std::shared_ptr<Shape>> sortedShapes = m_shapes;
+  std::sort(sortedShapes.begin(), sortedShapes.end(), compareArea);
+
+  std::cout << "Shapes ordered by area:" << std::endl;
+  for (auto shape : sortedShapes)
```
StephenSmith25: Const reference here as well.

exercises/tom_hampson/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/**
+ * @brief Print all shapes with n sides.
+ * @param n 
+*/
+void ShapeSorter::printShapesWithNSides(const int n) const
+{
+  std::cout << "Shapes with " << n << " sides:" << std::endl;
+  for (auto shape : m_shapes)
```
StephenSmith25: Const reference here as well.

StephenSmith25 (commented):
Looks good, just some minor comments.

exercises/tom_hampson/ex01_basics/src/main.cpp
thomashampson: Agreed.

exercises/tom_hampson/ex01_basics/src/main.cpp
thomashampson: Good idea

exercises/tom_hampson/ex01_basics/src/main.cpp
thomashampson: I found another instance of this too.

exercises/tom_hampson/ex02_oo_basics/src/ShapeSorter.h
thomashampson: Changed to reference

exercises/tom_hampson/ex02_oo_basics/src/main.cpp
thomashampson: Noted

exercises/tom_hampson/ex01_basics/src/main.cpp
thomashampson: Done

exercises/tom_hampson/ex02_oo_basics/src/ShapeSorter.cpp
thomashampson: Done

exercises/tom_hampson/ex02_oo_basics/src/ShapeSorter.cpp
thomashampson: done

exercises/tom_hampson/ex02_oo_basics/src/ShapeSorter.cpp
thomashampson: done

exercises/tom_hampson/ex02_oo_basics/src/ShapeSorter.cpp
thomashampson: done

PR #81, opened 2021-12-03
Rachel Baust exercises

rbauststfc:
Python new starter exercises

### DIFF
--- a/exercises-python/rachel_baust/ex01_basics/main.py
+++ b/exercises-python/rachel_baust/ex01_basics/main.py
@@ -55,7 +55,7 @@ def print_list(heading, tuple_list):
     :param tuple_list: the list of tuples to be printed
     :return: output is printed to the console
     """
-    print('{0}\n'.format(heading))
+    print(heading)
     for key, value in tuple_list:
         print(key, '   ', value)
 
--- a/exercises-python/rachel_baust/ex01_basics/main.py
+++ b/exercises-python/rachel_baust/ex01_basics/main.py
@@ -1,9 +1,44 @@
 #!/usr/bin/env python3
+from re import split
+import argparse
 
 
 def main():
-	pass
+    parser = argparse.ArgumentParser(description='Count words from a file')
+    parser.add_argument('filename', help='The path to the file')
+    args = parser.parse_args()
+
+    file_content = read_content(args.filename)
+    word_count = create_word_dictionary(file_content)
+    print_list('Word   Usage\n', sorted(word_count.items(), key=lambda x: x[1], reverse=True))
+
+
+def read_content(filename):
+    try:
+        file = open(filename, mode='rt', encoding='utf-8')
+        return file.readlines()
+    finally:
+        file.close()
+
+
+def create_word_dictionary(content):
+    word_count = {}
+    for line in content:
+        words = split(r'\W+', line.replace("\'", ''))
+        for word in words:
+            if len(word) > 4:
+                if word.casefold() in word_count:
+                    word_count[word.casefold()] += 1
+                else:
+                    word_count[word.casefold()] = 1
+    return word_count
+
+
+def print_list(heading, dictionary):
+    print(heading)
+    for key, value in dictionary:
+        print(key, '   ', value)
 
 
 if __name__ == "__main__":
-	main()
+    main()
--- a/exercises-python/rachel_baust/ex01_basics/main.py
+++ b/exercises-python/rachel_baust/ex01_basics/main.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 from re import split
+from os.path import exists
 import argparse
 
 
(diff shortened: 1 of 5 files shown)


exercises-python/rachel_baust/ex01_basics/main.py
```diff
@@ -0,0 +1,64 @@
...
+        print_list('Word   Usage', sorted(word_count.items(), key=lambda x: x[1], reverse=True))
+
+
+def read_content(filename):
+    """
+    Reads in all the lines from the file at the given path
+    :param filename: the path to the file to be read
+    :return: list[str] or None if the file couldn't be found
+    """
+    if exists(filename):
```
samtygier-stfc: This would be a good place to use a try/except. You could give a different message for FileNotFoundError and other file issues OSError.

https://docs.python.org/3/tutorial/errors.html#exceptions

exercises-python/rachel_baust/ex01_basics/main.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    word_count = {}
+    for line in content:
+        words = split(r'\W+', line.replace("\'", ''))
+        for word in words:
+            if len(word) > 4:
+                if word.casefold() in word_count:
+                    word_count[word.casefold()] += 1
+                else:
+                    word_count[word.casefold()] = 1
```
samtygier-stfc: The `collections` has a couple of special data types that let you avoid having to check if the key already exists. `Counter` is a specialised dict, which automatically initialise new keys to 0. There is also `defaultdict` that works for other types.

https://docs.python.org/3.8/library/collections.html#collections.defaultdict

exercises-python/rachel_baust/ex02_oo_basics/main.py
```diff
@@ -0,0 +1,104 @@
...
+    @abstractmethod
+    def perimeter(self):
+        """
+        Calculates the perimeter of the shape
+        :return: value for the perimeter of the shape
+        """
+        pass
+
+    def print_shape(self):
+        print('Shape: {0}, Area: {1}, Perimeter: {2}'.format(self.type, self.area(), self.perimeter()))
```
samtygier-stfc: A newish feature in python is f-strings. You could write as

`print(f'Shape: {self.type}, Area: {self.area()}, Perimeter: {self.perimeter()}')`

exercises-python/rachel_baust/ex02_oo_basics/main.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class Circle(Shape):
+
+    def __init__(self, radius):
+        super().__init__('Circle', 0)
+        self._radius = radius
+
+    def area(self):
+        return pi * sq(self._radius, 2)
```
samtygier-stfc, line 83: Renaming things when importing is useful, but can be confusing outside a few very common ones (e..g numpy as np).

You could use https://docs.python.org/3/library/functools.html#functools.partial to make a `sq` from `pow`.

samtygier-stfc (changes requested):
Great, nice and clear. Good use of python features.

I suggested a couple extra things you could try.

exercises-python/rachel_baust/ex01_basics/main.py
rbauststfc: Ah, thanks - I'd started off playing around with exceptions but got into a bit of tangle because I was using a more long-winded `file.open()`, `file.close()` approach to begin with. The exceptions seem to be working nicely with the way the code is now, though.

exercises-python/rachel_baust/ex01_basics/main.py
rbauststfc: This change makes it much cleaner, thanks

exercises-python/rachel_baust/ex02_oo_basics/main.py
rbauststfc: Thanks for that - the `functools,partial` feature looks really interesting. I've had a play around with it but couldn't get it to work with `pow` because it seems that `math.pow` uses positional arguments? On this one, my thinking behind renaming (where I actually made a very poor naming choice) was to avoid confusion between the built-in `pow` and the `math.pow` variants. Is this something that's worth worrying about, or is it fine to expect developers to check the import statement if they need to identify which version is being used?

samtygier-stfc: Ah, maybe partial won't work here. Another way would be with a lambda `sq = lambda x: pow(x,2)`.

I forgot about the built in `pow`. You could use it as an argument for doing `import math` and then using `math.pow()`. Probably most of the time the difference between them does not matter, if you know that it did then you'd have to check. So it would usually be fine to do `from math import pow`

samtygier-stfc (approved):

samtygier-stfc (approved):

gemmaguest:
Closing seeing as this was all completed

PR #84, opened 2022-09-01
Donald chung exercises

DonaldChung-HK:


exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+			c = ::tolower(c);
+			});
+		if (current_word.length() > 4)
+		{
+			++word_counts[current_word];
+		}
+	}
+}
+
+bool compare(std::pair<std::string, int>& a, std::pair<std::string, int>& b)
```
robertapplin, line 29: These `a` and `b` parameters can both be `const` as they are not being changed within this function. Using `const` allows you to document your code more clearly and actually enforce that documentation.

See [this link](https://www.cprogramming.com/tutorial/const_correctness.html) for more info

Another thing to note about `const` is that you can have 'east const' and 'west const' as described on [this link](https://hackingcpp.com/cpp/design/east_vs_west_const.html). In general, it is probably better to prefer using 'east const' as it allows for greater consistency in your code, however if 'west const' is used in a particular area of code, then following the convention is usually the best thing to do.

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+	return a.second > b.second;
+}
+
+void sorting(word_counting_container& word_counts, std::vector<std::pair<std::string, int>>& arrange)
+{
+
+	for (auto& it : word_counts) {
+		arrange.emplace_back(it);
+	}
+	std::sort(arrange.begin(), arrange.end(), compare);
```
robertapplin, line 40: Nice use of std algorithms :+1:

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+bool compare(std::pair<std::string, int>& a, std::pair<std::string, int>& b)
+{
+	return a.second > b.second;
+}
+
+void sorting(word_counting_container& word_counts, std::vector<std::pair<std::string, int>>& arrange)
+{
+
+	for (auto& it : word_counts) {
```
robertapplin, line 37: The `it` variable here can be `const`, and perhaps a name like `wordCount` would be a more descriptive name

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+}
+
+bool compare(std::pair<std::string, int>& a, std::pair<std::string, int>& b)
+{
+	return a.second > b.second;
+}
+
+void sorting(word_counting_container& word_counts, std::vector<std::pair<std::string, int>>& arrange)
```
robertapplin, line 34: `word_counts` can also be `const`

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+ * Skeleton main routine
+ */
+#include <algorithm>
+#include <string>
+#include <iostream>
+#include <map>
+#include <fstream>
+#include <vector>
+
+typedef std::map<std::string, int> word_counting_container; //typealias
```
robertapplin, line 11: This is a nice idea to define complicated types to improve code readability :+1: 

In C++, the convention we use when creating variables, functions and other named things is [camel case](https://en.wikipedia.org/wiki/Camel_case). In python we tend to use snake case as you are doing here

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+void word_counting(std::istream& input, word_counting_container& word_counts)
+{
+	std::string current_word;
+	while (input >> current_word)
+	{
+		current_word.erase(std::remove_if(current_word.begin(), current_word.end(), ispunct), current_word.end());//remove punctuation
+		std::for_each(current_word.begin(), current_word.end(), [](char& c) {
```
robertapplin, line 19: Nice use of a lambda :+1:

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <algorithm>
+#include <string>
+#include <iostream>
+#include <map>
+#include <fstream>
+#include <vector>
+
+typedef std::map<std::string, int> word_counting_container; //typealias
+
+void word_counting(std::istream& input, word_counting_container& word_counts)
```
robertapplin, line 13: We use doxygen-style commenting for our code, [this file](https://github.com/mantidproject/mantid/blob/62fefcd6e145062fa25ac999046b9bd33c4c1be5/Framework/CurveFitting/src/SeqDomainSpectrumCreator.cpp#L39) has some good examples. This looks like a good candidate function to have a doxygen comment, and the same with some of the longer functions below

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
robertapplin: Just as a side note (no need to change anything), a similar effect can be achieved with
```
using WordCountingContainer = std::map<std::string, int>;
```

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+void output_result(std::vector<std::pair<std::string, int>>& arrange)
+{
+	std::ofstream outf{ "result.txt" };
+	if (!outf)
+	{
+		std::cerr << "result.txt" << " could not be opened for writing!\n";
+		std::exit(1);
+	}
+	outf << "Word, Usage,\n";
+	for (int i = 0; i < arrange.size(); i++)
```
robertapplin, line 52: The return type of `arrange.size()` is a `std::size_t` (an integer which is positive or zero). In this case, giving `i` the `int` type will cause a conversion warning as we can't do an implicit conversion from `std::size_t` to `int`. Its probably best to fix this by using `auto i = 0u; i < arrange.size()` to auto-deduce the type of `i` from the `.size()` operation

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+	for (auto& it : word_counts) {
+		arrange.emplace_back(it);
+	}
+	std::sort(arrange.begin(), arrange.end(), compare);
+}
+
+void output_result(std::vector<std::pair<std::string, int>>& arrange)
+{
+	std::ofstream outf{ "result.txt" };
```
robertapplin, line 45: At the end of this function I think we want to close the output file stream with `outf.close();`

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+	{
+		std::cerr << "result.txt" << " could not be opened for writing!\n";
+		std::exit(1);
+	}
+	outf << "Word, Usage,\n";
+	for (int i = 0; i < arrange.size(); i++)
+	{
+		outf << arrange[i].first << "," << arrange[i].second << ",\n";
+	}
+	std::cout << "result output to result.txt";
```
robertapplin, line 56: At the end of this line we want to 'flush' cout to make sure all the messages are displayed to the user. To do this you can do:
 ```
std::cout << "result output to result.txt" << std::endl;
```

Its explained a bit better [here](https://www.educative.io/answers/what-is-the-difference-between-endl-and-n-in-cpp)

File: exercises-cpp/donald_chung/ex01_basics/src/main.cpp [hunk 11]
```diff
@@ -0,0 +1,79 @@
...
+int main(int argc, char* argv[])
+{
+	if (argc > 1)
+	{
+		std::string arg1{ argv[1] };
+		std::ifstream input_file{ arg1 };
+		if (!input_file)
+		{
+			// Print an error and exit
+			std::cerr << arg1 << " could not be opened for reading!\n";
```
robertapplin, line 68: I think we want to flush `std::cerr` here too with `std::endl`

exercises-cpp/donald_chung/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+	}
+	std::sort(arrange.begin(), arrange.end(), compare);
+}
+
+void output_result(std::vector<std::pair<std::string, int>>& arrange)
+{
+	std::ofstream outf{ "result.txt" };
+	if (!outf)
+	{
+		std::cerr << "result.txt" << " could not be opened for writing!\n";
```
robertapplin, line 48: See the comment about 'flushing' std::cout

robertapplin (changes requested):
Nice work with exercise 1, it looks really good for a first attempt! I just added a few comments below

exercises-cpp/donald_chung/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <cmath>
+#include <iostream>
+#include "Circle.h"
+
+
+Circle::Circle(double radius_i) : Shape("Circle", 1), radius{ radius_i }
+{
+}
+
+const double pi = 3.14159265358979323846;
```
robertapplin, line 11: Instead of defining pi yourself, I think you can use the `M_PI` macro provided in `cmath`

exercises-cpp/donald_chung/ex02_oo_basics/src/Triangle.h
```diff
@@ -0,0 +1,17 @@
+#ifndef TRIANGLE_H
+#define TRIANGLE_H
+
+#include "Shape.h"
+
+class Triangle : public Shape
+{
+private:
+    double base, height;
```
robertapplin, line 9: The convention used in mantid is to use an `m_` in front of class member variable names. So these would be `m_base` and `m_height`. This is the only exception to the camel case rule.

exercises-cpp/donald_chung/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,20 @@
...
+#include <string>
+
+class Shape
+{
+protected:
+    std::string my_type;
+    int num_sides;
+public:
+    Shape(std::string my_type_i, int num_sides_i);
+    virtual double perimeter() = 0;
```
robertapplin, line 13: Good use of pure virtual functions :+1: We need to override these in the child classes

exercises-cpp/donald_chung/ex02_oo_basics/src/Rectangle.h
```diff
@@ -0,0 +1,17 @@
...
+
+class Rectangle : public Shape
+{
+private:
+    double len, wid;
+public:
+    Rectangle(double len_i, double wid_i);
+    double perimeter();
+    double area();
+    void print();
```
robertapplin, line 14: It would be good to mark these with the `override` keyword, and similarly in the other shape classes

exercises-cpp/donald_chung/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,20 @@
+#ifndef SHAPE_H
+#define SHAPE_H
+
+#include <string>
+
+class Shape
+{
+protected:
+    std::string my_type;
+    int num_sides;
```
robertapplin, line 10: Using the convention in Mantid, the member variable names for these would be `m_numSides` and `m_type`

exercises-cpp/donald_chung/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,33 @@
...
+{
+	double result{ 2 * pi * radius };
+	return result;
+}
+double Circle::area()
+{
+
+	double result{ pi * (std::pow(radius, 2)) };
+	return result;
+}
```
robertapplin, line 23: These two functions, `perimeter` and `area`, can just return the result without declaring a variable for it. e.g. `return M_PI * std::pow(radius, 2);`

exercises-cpp/donald_chung/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,33 @@
+
+#include <cmath>
+#include <iostream>
+#include "Circle.h"
+
+
+Circle::Circle(double radius_i) : Shape("Circle", 1), radius{ radius_i }
```
robertapplin, line 7: I like that you are using an initializer list :+1: We tend to use circular brackets for this as it makes it easier to read e.g. `m_radius(radius)`. The `i` on the end of the radius variable name can also probably be removed as for me it doesn't provide a better description for what the variable stores.

exercises-cpp/donald_chung/ex02_oo_basics/src/Circle.cpp
```diff
@@ -0,0 +1,33 @@
...
+}
+
+void Circle::print()
+{
+	std::cout << "Type: " << this->get_type() <<
+		"; Number of Sides: " << this->get_num_sides() <<
+		"; Radius: " << this->radius <<
+		"; Area: " << this->area() <<
+		"; Perimeter: " << this->perimeter() <<
+		"\n";
```
robertapplin, line 32: Same comment about flushing cout with `std::endl`. And in other similar files

exercises-cpp/donald_chung/ex02_oo_basics/src/Shape.h
robertapplin: There is also no state change of the classes within these functions, so they can be declared as `const`:
```
virtual double perimeter() const = 0;
```

exercises-cpp/donald_chung/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,20 @@
...
+public:
+    Shape(std::string my_type_i, int num_sides_i);
+    virtual double perimeter() = 0;
+    virtual double area() = 0;
+    int get_num_sides();
+    std::string get_type();
```
robertapplin, line 16: These non-pure virtual functions can also be `const`

exercises-cpp/donald_chung/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,38 @@
...
+{
+    std::vector<Shape*> shapes;
+    shapes.push_back(new Rectangle{ 10, 3 });
+    shapes.push_back(new Square{ 2.5 });
+    shapes.push_back(new Rectangle{ 50, 50 });
+    shapes.push_back(new Rectangle{ 20, 1 });
+    shapes.push_back(new Circle{ 3 });
+    shapes.push_back(new Circle{ 50 });
+    shapes.push_back(new Triangle{ 5, 2 });
+    shapes.push_back(new Triangle{ 7, 3 });
```
robertapplin, line 23: Using `emplace_back` instead of `push_back` will avoid a copy/move operation, and instead create the shapes inside the vector. This would be a more efficient and faster operation.

https://stackoverflow.com/questions/4303513/push-back-vs-emplace-back

exercises-cpp/donald_chung/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include "Square.h"
+#include "Rectangle.h"
+#include "ShapeSorter.h"
+#include "Circle.h"
+#include "Triangle.h"
+
+
+int main(int, char **)
+{
+    std::vector<Shape*> shapes;
```
robertapplin, line 15: In this case, you know the number of shapes you want to put into this vector, so it makes sense to reserve memory by doing `shapes.reserve(8);`. 

https://www.geeksforgeeks.org/using-stdvectorreserve-whenever-possible/

exercises-cpp/donald_chung/ex02_oo_basics/src/main.cpp
```diff
@@ -0,0 +1,38 @@
...
+    std::cout << "Only Print Rectangle \n";
+    ShapeSorter().print_type(shapes, "Rectangle");
+    std::cout << "Only Print Circle \n";
+    ShapeSorter().print_type(shapes, "Circle");
+    std::cout << "Only Print side 4 \n";
+    ShapeSorter().print_sides(shapes, 4);
+    std::cout << "Only Print side 1 \n";
+    ShapeSorter().print_sides(shapes, 1);
+    std::cout << "Only Print side 3 \n";
+    ShapeSorter().print_sides(shapes, 3);
```
robertapplin, line 37: Can we create a single instance of this `ShapeShorter` and reuse it?

exercises-cpp/donald_chung/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,54 @@
...
+		if (shapes[i]->get_type() == type)
+		{
+			shapes[i]->print();
+		}
+	}
+}
+
+void ShapeSorter::print_sides(std::vector<Shape*> shapes, int sides)
+{
+	for (int i = 0; i < shapes.size(); i++)
```
robertapplin, line 47: There are conversion problems in this file too between size_t and int

exercises-cpp/donald_chung/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,54 @@
...
+void ShapeSorter::print_byperimeter_dec(std::vector<Shape*> shapes)
+{
+	std::sort(shapes.begin(), shapes.end(), bool_byperimeter_dec);
+	for (int i = 0; i < shapes.size(); i++)
+	{
+		shapes[i]->print();
+	}
+}
+
+void ShapeSorter::print_type(std::vector<Shape*> shapes, std::string type)
```
robertapplin, line 34: The parameters passed to all these functions can probably be `const &`. Passing them in by value (as is done here) causes a copy of the vector of shapes and a copy of the type string. Passing them in by reference will avoid these copy operations

exercises-cpp/donald_chung/ex02_oo_basics/src/ShapeSorter.cpp
```diff
@@ -0,0 +1,54 @@
+#include "ShapeSorter.h"
+#include <vector>
+#include <algorithm>
+
+ShapeSorter::ShapeSorter() {};
+
+bool bool_byarea_dec(Shape *a, Shape *b) 
+{
+	return a->area() > b->area();
+}
```
robertapplin, line 10: The functions which are not part of a class are usually defined in something called an 'anonymous namespace' at the top of the file (underneath the includes). For example:
```
namespace { // anonymous namespace

bool isAreaLarger(Shape const *a, Shape const *b) 
{
	return a->area() > b->area();
}

}

robertapplin (changes requested):
Your second exercise looks good too. I've added some more comments. One thing to look out for is to use camel case wherever possible, as well as constness, and passing parameter values in by reference to avoid needless copies

DonaldChung-HK:
Hi Rob,

Thanks for reviewing my code.

Donald

PR #86, opened 2022-10-07
Jack Allen Exercise 1

JackEAllen:
Complete [Python exercise 1](https://developer.mantidproject.org/NewStarterPython.html#exercise)

Solution is a terminal based application which reads a text file (.txt) and performs text sanitisation on raw text and returns the frequency of words longer than 4 characters in the following format:

```
Word                 Usage
----                 -----
morning              10
hello                6
goodbye              3
```


jack_allen/template/ex01_basics/main.py
```diff
@@ -0,0 +1,140 @@
...
+The text file used as input must be in the same directory as the script to work
+
+Run the following command in your terminal to run the script: `python main.py text.txt`
+"""
+
+import sys
+import os
+
+# Method to read a text file and return as a string
+def read_text_file(filename: str):
```
samtygier-stfc, line 32: Can you annotate the return types as well. e.g.
`def read_text_file(filename: str) -> str:`

samtygier-stfc (changes requested):
Looks good. For the type annotations, adding return type annotations will allow quite a bit more checking.

For the counting algorithm, its worth switching to a more efficient single pass method.

JackEAllen:
> Looks good. For the type annotations, adding return type annotations will allow quite a bit more checking.
> 
> For the counting algorithm, its worth switching to a more efficient single pass method.

Refactored counting algorithm to use Counter from collections to achieve going through text in one pass.

samtygier-stfc (approved):

samtygier-stfc:
All good so closing (without merge, because its newstarter repo)

PR #87, opened 2022-10-07
Jack Allen exercises 2

JackEAllen:
Complete [Python exercise 2](https://developer.mantidproject.org/NewStarterPython.html#exercise)

Solution is a terminal based application written following OOP design methodology which reads returns filtered information about shapes based on criteria set in exercise.

Criteria:
1. Print out the Shapes that match a chosen type
2. Print out the Shapes that match a chosen number of sides
3. Print out the Shapes in order of area descending
4. Print out the Shapes in order of perimeter descending

### DIFF
--- a/jack_allen/template/ex02_oo_basics/main.py
+++ b/jack_allen/template/ex02_oo_basics/main.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-
+#pylint: disable=anomalous-backslash-in-string
 """
 Solution for Exercise 2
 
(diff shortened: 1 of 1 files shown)


samtygier-stfc (approved):
For a real PR there would be a few fixes I'd ask for, but there no need to implement them here, so I'll approve this.

(using math.pi instead of a literal number, using an [ABC](https://docs.python.org/3/library/abc.html) to define the Shape interface, and I'd suggest that the properties are just named `area` and `perimeter` as the `@property` is meant to make them feel like attributes)

JackEAllen:
> For a real PR there would be a few fixes I'd ask for, but there no need to implement them here, so I'll approve this.
> 
> (using math.pi instead of a literal number, using an [ABC](https://docs.python.org/3/library/abc.html) to define the Shape interface, and I'd suggest that the properties are just named `area` and `perimeter` as the `@property` is meant to make them feel like attributes)

I agree with you about both points. I chose to minimise library usage as was not sure what checks would be ran over code in remote branch.

PR #88, opened 2022-12-15
Mohammed almakki exercies

MohamedAlmaki:
This PR contains the solution of c++ exercise 1 and 2


exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+    else
+    {
+        throw std::invalid_argument("Invalid path. Please make sure the file path is correct and the file exists.");
+    }
+}
+
+std::vector<std::string> tokenizeFile(const std::string &file)
+{
+    std::regex re("[-\\s]");
```
Harrietbrown, line 28: is the inent of this that the experssion match a hyphen of a whitespace, but a double backslash exits the special character.

exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,123 @@
...
+std::unordered_map<std::string, uint32_t> countWords(const std::vector<std::string> words)
+{
+    std::unordered_map<std::string, uint32_t> wordsCounts;
+    for (auto word : words)
+    {
+        std::transform(
+            word.begin(), word.end(), word.begin(),
+            [](unsigned char c)
+            { return std::tolower(c); });
+        word.erase(std::remove_if(word.begin(), word.end(), ispunct), word.end());
```
Harrietbrown, line 43: The task specifies only needing to consider `.,?'"!():` you might have considered using regex to strip these away and replacing them with a whitespace at the same time as doing the hyphen.

exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,123 @@
...
+        if (word.length() > 4)
+        {
+            wordsCounts[word]++;
```
Harrietbrown, line 46: you should have a check to see if word is currently in the wordCounts map

exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
```diff
@@ -0,0 +1,123 @@
...
+}
+
+void sortAndWriteWords(const std::unordered_map<std::string, uint32_t> &wordsCounts, const std::string &outputDir)
+{
+    std::vector<std::pair<std::string, int>> wordsCountsPairs;
+    for (auto &pair : wordsCounts)
+    {
+        wordsCountsPairs.push_back(pair);
+    }
+    std::sort(
```
Harrietbrown, line 59: Good sorting process.

Harrietbrown (changes requested):
Exercise 1

exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
MohamedAlmaki: This regex is used to split the string by two delimiters, the space, and the hyphen. It will match a space or a hyphen. After that, the expression is used in a reverse way, to get all the words that don't match it i.e. the split words.

exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
MohamedAlmaki: As I understood, the comparison should be punctation insensitive. So for example, "what's" should be equal to "whats" but replacing it with whitespace will result in the word not being included (not longer than 4). So in this line, I just remove every punctuation mark.

exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
MohamedAlmaki: Hmmm, why is it needed?
If there is a get operation, then I need a check, but this is just used as a counter.

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,156 @@
+#pragma once
+
+#include <string>
+
+class Shape
+{
+public:
+    virtual std::string getType() = 0;
```
MialLewis: For const correctness, the getters/functions not intended to mutate member variables should be `const` member functions

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,156 @@
...
+    virtual int getSidesCount() = 0;
+    virtual double getArea() = 0;
+    virtual double getPerimeter() = 0;
+
+    virtual std::string toString() = 0;
+
+    Shape(int noOfSides) : noOfSides{noOfSides} {}
+
+private:
+    int noOfSides;
```
MialLewis: In this exercise, this could be `const` as there's no requirement to mutate the shapes after construction.

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,156 @@
...
+{
+public:
+    virtual std::string getType() = 0;
+    virtual int getSidesCount() = 0;
+    virtual double getArea() = 0;
+    virtual double getPerimeter() = 0;
+
+    virtual std::string toString() = 0;
+
+    Shape(int noOfSides) : noOfSides{noOfSides} {}
```
MialLewis: Argument could be `const`.

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,156 @@
...
+
+private:
+    int noOfSides;
+};
+
+class Square : public Shape
```
MialLewis: I'd have been tempted to have `square` inherit from `rectangle` to avoid some code duplication. Override functions where necessary.

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,156 @@
...
+        return 2 * side1 + 2 * side2;
+    }
+
+    std::string toString()
+    {
+        return "Rectangle(side1=" + std::to_string(side1) + ", side2=" + std::to_string(side2) + ")";
+    }
+
+private:
+    double side1, side2;
```
MialLewis: Picky in this instance, but we generally avoid declaring multiple variables in one line.

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    std::string toString()
+    {
+        return "Circle(radius=" + std::to_string(radius) + ")";
+    }
+
+private:
+    double radius;
+    const double pi = 3.14159265358979323846;
```
MialLewis: Can we get this from `math.h`?

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/Shape.h
```diff
@@ -0,0 +1,156 @@
...
+{
+public:
+    Square(double side1) : side1{side1}, Shape(2) {}
+
+    std::string getType()
+    {
+        return "Square";
+    }
+
+    int getSidesCount()
```
MialLewis: Could we have this defined in the `Shape` parent class, simply returning its member `noOfSides`? We then would not have to redefine it in every child.

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,75 @@
...
+#include "Shape.h"
+
+class ShapeSorter
+{
+public:
+    ShapeSorter(const std::vector<std::shared_ptr<Shape>> &shapes) : shapes{shapes} {}
+
+    void printShapesWithType(const std::string &type)
+    {
+        for (auto shape : shapes)
```
MialLewis: These functions are now at a point where they should no longer be defined inline, but in a separate `.cpp` file. I'm not sure if this is within the scope of the exercise.

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,75 @@
...
+            {
+                std::cout << shape->toString() << " ";
+            }
+        }
+        std::cout << "\n";
+    }
+
+    void printShapesWithSides(const int &sidesCount)
+    {
+        for (auto shape : shapes)
```
MialLewis: Likely needless complication in this instance, but could be interesting to see if you could reduce code duplication (use of function pointers?)

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/ShapeSorter.h
```diff
@@ -0,0 +1,75 @@
...
+        auto sortedShapes = shapes;
+
+        std::sort(
+            sortedShapes.begin(), sortedShapes.end(),
+            [](auto left, auto right)
+            {
+                return left->getArea() > right->getArea();
+            });
+
+        for (auto shape : sortedShapes)
```
MialLewis: As per the above - try to refactor to reduce the code duplication here.

MialLewis (commented):
Good job - a few comments on the second exercise. Some of the comments apply in more than one place, but I've only pointed out the first occurrence.

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/Shape.h
MohamedAlmaki: Okay

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/Shape.h
MohamedAlmaki: Yeah, that is really bad. Doesn't know why I did it like this.

exercises-cpp/mohammed_almakki/ex02_oo_basics/src/ShapeSorter.h
MohamedAlmaki: Yes, we can make a general function that prints shapes that meet a specific condition. The function will accept a lambda filter function as a parameter. The lambda function will accept a parameter of type shape and return a Boolean value. Also, we can get rid of the for loop and use for_each. But I think it is too complex for this small example.

MialLewis (approved):
Thanks for the improvements - your solution using the lambda filter functions is really neat.

MohamedAlmaki:
> Thanks for the improvements - your solution using the lambda filter functions is really neat.

Thanks for your great feedback!

exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
Harrietbrown: A Hyphen is also a special character and will need to be exited as well

exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
Harrietbrown: In that case I think you may be right

exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
Harrietbrown: Ultimatly it is not required for it to work, but it is better practice to have the check when performing an action on an object in a map where it is not certain that the object will exist.

exercises-cpp/mohammed_almakki/ex01_basics/src/main.cpp
MohamedAlmaki: Yes, it will be removed. For example, the string "lines are hard-coded" will be transformed into a vector of ["lines", "are", "hard", "coded"] by this expression.

PR #92, opened 2023-10-23
ex01 exercise completed

MikeSullivan7:
Written as part of the Mantid New Starter exercises.

- The code takes an ascii file as argument and counts the occurrences of unique words longer than 4 characters long. 
- The punctuation is removed, hyphenated words are split up, and the text is all made lowercase.
- A check is run to catch an error if the ascii file cannot be found and informs the user that this is the case, the code then exits.
- The code prints out the list of unique words and prints then in descending order of number of occurrences.
- It is also possible to specify using the --lines flag to only print out the most frequent e.g. 10 words so that the command line isn't taken up by a long list of every word.
- The code can be tested with the ascii file "Holmes.txt" in the exercise folder.

### DIFF
--- a/exercises-python/mike_sullivan/ex01_basics/main.py
+++ b/exercises-python/mike_sullivan/ex01_basics/main.py
@@ -1,0 +1,71 @@
+'''
+Write a command line program that will:
+
+Take a filename of an ascii file as an argument (you can use the example file here)
+
+Load that ascii file.
+
+Count the number of occurrences of unique words (longer than 4 characters and split hyphenated words, treating each part as different words). It should be case and punctuation insensitive. You only need to consider the following punctuation characters .,?'"!(): (hint: you will need a backslash escape character for the double-quote)
+
+Consider handling of common error cases, such as the wrong file name specified. Return error and status information to the user of the command line tool.
+
+Print the results to screen showing the unique words and the number of uses in descending order of usage, e.g.
+
+
+'''
+
+import argparse
+import re
+import numpy as np
+#import os.path
+def getfilename():
+	# take in filename from user in command line
+	parser = argparse.ArgumentParser(
+		prog='top',
+		description='read ascii file')
+	parser.add_argument('filename', nargs='+')
+	parser.add_argument('-l', '--lines', type=int) # give option to show top number of lines
+	args = parser.parse_args()
+	print(args)
+	print(f"Entered file is: {args.filename}")
+	#check_file = os.path.exists(args.filename)
+	print(args.filename[0], args.lines)
+	return args.filename[0], args.lines
+
+def readfile(filename):
+	f = open(filename, 'r')  # 'r' = read
+	linesRaw = f.read()
+
+	lines = re.sub(r'(\W+)', lambda x: ' '+x.group(0)+' ', linesRaw).split() # separate punctuation
+	lines = [i.lower() for i in lines] #force all strings to be lowercase
+	lines = [x for x in lines if len(x)>4] #words longer than 4 characters
+	f.close()
+	return lines
+
+def getuniquewords(lines):
+	values, counts = np.unique(lines, return_counts=True)
+	return values, counts
+
+def main():
+	# Ask for filename from user
+	filename, numlines = getfilename()
+	# Check if file exists
+	try:
+		lines = readfile(filename)
+	except FileNotFoundError:
+		print("File not found, please check file!")
+		exit()
+		# Cant find the file so exit the code
+
+	# Sort the words by their frequency in the file
+	values, counts = getuniquewords(lines)
+	sortedValuesCounts = [x for x in sorted(zip(counts,values),reverse=True)]
+
+	# print out words in descending order of frequency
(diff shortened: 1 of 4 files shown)


exercises-python/mike_sullivan/ex01_basics/main.py
```diff
@@ -0,0 +1,71 @@
...
+Consider handling of common error cases, such as the wrong file name specified. Return error and status information to the user of the command line tool.
+
+Print the results to screen showing the unique words and the number of uses in descending order of usage, e.g.
+
+
+'''
+
+import argparse
+import re
+import numpy as np
```
samtygier-stfc, line 19: In the checks I can see `ModuleNotFoundError: No module named 'numpy'` because the tests run in a fairly minimal python environment.

You could either try rewriting `getuniquewords()` to avoid using numpy. https://docs.python.org/3/library/collections.html?highlight=counter#collections.Counter might be useful. (Though in actual mantid dev work we use numpy a lot)

Or, you could adjust the test workflow to install numpy. There is a file `.github/workflows/python-tests.yml` that tells github actions what to do. A step to do `pip install numpy` would work.

samtygier-stfc (changes requested):

MikeSullivan7:
I have fixed the numpy issue by modifying the test workflow, however there is an Assertion Error:

`AssertionError: Lists differ: ["Namespace(filename=['/home/runner/files/Holmes.txt'],", 'lines=None)'] != ['Word', 'Usage']`

Looking into it now!

MikeSullivan7:
All tests for Ex_01 have passed! 

Will start working on Exercise 2 now.

MikeSullivan7 (commented):
Changes made to pass python_tests

MikeSullivan7:
ex_02 has been completed with all checks passing.

- The code has classes with a number of shapes with can be defined
- each class knows its type, its number of sides and its lengths of each side.
- the classes have methods to calculate their own perimeter and area, with a method to print out the information about each individual shape so they can be identified
- a shape sorting class is defined which takes a list of shape objects as input and is able to print out the info on shapes which have a certain type or number of sides
- methods also exist which can sort the list of shapes in descending area or perimeter and then print out the information accordingly
- The code could potentially be improved by randomising the sides of each shape so the lists are different each time but wasn't sure it was completely necessary.

exercises-python/mike_sullivan/ex01_basics/main.py
```diff
@@ -0,0 +1,72 @@
...
+	parser = argparse.ArgumentParser(
+		prog='top',
+		description='read ascii file')
+	parser.add_argument('filename', nargs='+')
+	parser.add_argument('-l', '--lines', type=int) # give option to show top number of lines
+	args = parser.parse_args()
+	#print(args)
+	#print(f"Entered file is: {args.filename}")
+	#check_file = os.path.exists(args.filename)
+	#print(args.filename[0], args.lines)
```
samtygier-stfc: These and other commented out code can be cleaned up.

exercises-python/mike_sullivan/ex01_basics/main.py
```diff
@@ -0,0 +1,72 @@
...
+Print the results to screen showing the unique words and the number of uses in descending order of usage, e.g.
+
+
+'''
+
+import argparse
+import re
+import numpy as np
+#import os.path
+def getfilename():
```
samtygier-stfc: Can you rename the function as it gets number of lines as well. Maybe `get_user_arguments()`

exercises-python/mike_sullivan/ex01_basics/main.py
```diff
@@ -0,0 +1,72 @@
...
+		exit()
+		# Cant find the file so exit the code
+
+	# Sort the words by their frequency in the file
+	values, counts = getuniquewords(lines)
+	sortedValuesCounts = [x for x in sorted(zip(counts,values),reverse=True)]
+
+	# print out words in descending order of frequency
+	print('Word    Usage')
+	for i in sortedValuesCounts[0:numlines]:
```
samtygier-stfc: This could be clearer by unpacking `i`, e.g.
`for word, usage in ...`

exercises-python/mike_sullivan/ex02_oo_basics/main.py
```diff
@@ -0,0 +1,137 @@
...
+
+# Imports here
+import numpy as np
+
+# Shape Classes here
+class square():
+	def __init__(self, side1):
+		self.side1 = side1
+		self.type = "Square"
+		self.number_sides = 4
```
samtygier-stfc: As these are the same for any Square, they could be class variables. e.g.
```
class square():
    type = "Square"
    number_sides = 4
    def __init__(self, side1):
        ...
```

samtygier-stfc (changes requested):
Looks good. Just need a few cleanups and tweaks.

MikeSullivan7:
All suggested changes have been made

samtygier-stfc (approved):
Thanks for changes.

